PR #36, opened 2019-06-04
Ui bug fixing

### LINKED ISSUE #24: Clickable "Failed Controls" button in Compliance View
Currently, the "Failed Controls" button in the Compliance View is not clickable.

### LINKED ISSUE #35: Link in Compliance view to rule incorrect
The link is still „analysis“ instead of „rules“

immqu:
Closes #24
Closes #35

### DIFF
--- a/clouditor-ui/src/app/compliance-detail/compliance-detail.component.html
+++ b/clouditor-ui/src/app/compliance-detail/compliance-detail.component.html
@@ -41,33 +41,28 @@
         <span class="input-group-text">
           <i class="fas fa-search"></i>
         </span>
-        <div class="input-group-text">
-          <label class="form-check-label"><input class="search-check" type="checkbox" name="filterWaiting"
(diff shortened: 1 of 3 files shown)


oxisto:
It would be nice that if you click on failed controls it would automatically change the filter in the compliance view to only failed controls. This should be doable with a query parameter similar to how it is solved in the rules component. Can you look into this?

immqu:
I have added the query params; however, the filterOptions persist when leaving the view, shall we reset it when visiting the certification normally?

oxisto:
Maybe we should get rid of the persistence all together and do it like this:
- Click on „failed controls“ -> failed only
- Click on „passed controls“ -> passed only
- Click on title -> all

oxisto:
Made a change to keep the old style buttons for now since we may want to globally replace them at some point. Otherwise ready to merge once tests are ok.

PR #51, opened 2019-07-04
 Fixed #48 by properly persisting scans now

### LINKED ISSUE #48: Configuration of scans and rules are not persisted in the database
**Describe the bug**

Currently, the configuration of rules, i.e. which ones are activated is not persisted in the database.

**Expected behavior**

After restarting the Clouditor, I expect to see all my enabled scans and rules again.

oxisto:
Instead of storing `Scan` objects in memory, all changes are now written to the persistance layer. When the `DiscoveryService` is initialized, enabled checks are automatically started. `Scanner` objects are now automatically created if the scan starts.


oxisto:
Test should be fixed now, thanks.

Issue #45, opened 2019-07-02
Allow the selection of assets within a rule


oxisto:
**Is your feature request related to a problem? Please describe.**

Currently, rules are evaluated against *all* assets of a particular type. It would be good, if we could select a subgroup of assets i.e. using a sql-like *where*-clause.

PR #61, opened 2019-08-13
Asset selection in rules

immqu:
- Extended CCL to allow applying rules only to specific assets using the "with" statement, e.g. _user with name="admin" has [condition]_
- Changed CCL rules to eliminate some recursive statements
- Extended and fixed TimeComparison with different operators for referring to a property's age and a property's expiry in the future:
1) Use the `before` and `after` operators to define rules concerning dates in the future, e.g. `user has expiry before 100 days in all accessKeys` (meaning the expiry date is before today's date plus 100 days)
2) Use the `younger` and `older` operators to define rules concerning the age of assets, e.g. `user has createDate younger 100 days in all accessKeys` (meaning the creation date is after today's date minus 100 days)
- Changed respective tests


clouditor-engine-core/src/main/java/io/clouditor/assurance/Rule.java
```diff
@@ -83,14 +95,16 @@ public void setCondition(Condition condition) {
...
     }
 
     // multiple conditions
-    if (this.conditions != null && !this.conditions.isEmpty()) {
+    if (this.conditions != null
+        && !this.conditions.isEmpty()
+        && this.conditions.get(0).getAssetType() != null) {
       // take the first one
-      return this.conditions.get(0).getAssetType();
+      return this.conditions.get(0).getAssetType().getField();
```
oxisto: This creates a problem since only the first condition will be evaluated regarding its applicability but in future we are reworking this anyway in #62

clouditor-engine-core/src/main/java/io/clouditor/assurance/RuleService.java
```diff
@@ -285,8 +285,13 @@ public void handle(DiscoveryResult result) {
       // evaluate all rules
       rulesForAsset.forEach(
           rule -> {
-            var eval = rule.evaluate(asset);
-
+            EvaluationResult eval;
+            if (!rule.evaluateApplicability(asset)) {
+              // simply add an empty EvaluationResult
+              eval = new EvaluationResult(rule, asset.getProperties());
```
oxisto: Fine for now, although I am not sure what the effect on the UI is with an empty evaluation result

clouditor-engine-core/src/main/java/io/clouditor/assurance/ccl/AssetType.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Map;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, property = "@class")
+public abstract class AssetType {
+
+  public abstract boolean evaluate(Map properties);
+
+  public abstract String getField();
+
+  Object getValueFromField(Map asset, String fieldName) {
```
oxisto: Obsolete function

clouditor-engine-core/src/main/java/io/clouditor/assurance/ccl/AssetType.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import java.util.Map;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS, property = "@class")
+public abstract class AssetType {
+
+  public abstract boolean evaluate(Map properties);
+
+  public abstract String getField();
```
oxisto: rename to getValue()

clouditor-engine-core/src/main/antlr/CCL.g4
```diff
@@ -2,18 +2,25 @@ grammar CCL;
...
 condition: assetType 'has' expression EOF;
 
-assetType : Identifier;
+assetType :
+    simpleAsset |
+    groupedAsset;
+
+simpleAsset: field;
+
+groupedAsset: field 'with' expression;
```
oxisto: I would prefer something like `filteredAssetType` and `simpleAssetType`

oxisto:
Fixes #45

Issue #82, opened 2020-10-26
unify condition members in rules


immqu:
Currently, rules can have either one condition or a list of conditions. Instead, we can simply treat all rules as having a list of conditions

PR #83, opened 2020-10-28
Unified condition members in rules

immqu:
For some reason our Rules had a single condition member, as well as a list of conditions (in case there are several conditions to fulfill, e.g. the password policy rule).

Now, Rules have only a list of conditions, see #82


oxisto (commented):
I would expected *some* kind of description what this pull request is about ;-) 

You probably need to also adjust 
https://github.com/clouditor/clouditor/blob/3dc0907c6d3b9ab6bcc61815949aadd4c0c7edf9/clouditor-ui/src/app/rule-detail/rule-detail.component.html#L44-L47

as well as

https://github.com/clouditor/clouditor/blob/3dc0907c6d3b9ab6bcc61815949aadd4c0c7edf9/clouditor-ui/src/app/rule.ts#L29

immqu:
For some reason our Rules had a single condition member, as well as a list of conditions (in case there are several conditions to fulfill, e.g. the password policy rule). 
Now, Rules have only a list of conditions, see #82

Issue #88, opened 2020-11-19
Fix quality gate


oxisto:
Seems, that we lost our quality gate, since we have less than 65 % test coverage. We should fix that.

PR #93, opened 2021-03-10
Fix Quality Gate

LeBogg:
#88 Fix Quality Gate:

- Testing REST classes (*Test): AccountsResource, AssetsResource, CertificationResource, DiscoveryResource, RulesResource, StatisticsResource, UsersResource

- build.gradle: Increasing JUnit version from 5.3.1 to 5.4.0 allowing test ordering

- CertificationService: Adding methods for removing one or all certifications
- RuleService: Adding methods for removing one rule, all rules, or all rules per asset type
- Creating test classes for CertificationService and RuleService (testing remove methods)

- AccountsResource: Adding annotations for producing or consuming JSON format
- CertificationResource: Removing unused code (Code Smells) and adding annotations for producing JSON format
- DiscoveryResource: Adding annotations for producing  JSON format
- RulesResource: Adding annotations for producing  JSON format


clouditor-engine-core/src/test/java/io/clouditor/rest/DiscoveryResourceTest.java
```diff
@@ -0,0 +1,195 @@
...
+            .get();
+
+    // Assertions
+    List<Scan> scans = response.readEntity(List.class);
+    Assertions.assertFalse(scans.isEmpty());
+    Assertions.assertEquals(Response.Status.OK.getStatusCode(), response.getStatus());
+  }
+
+  @Test
+  void testGetScan_whenRequestedScannerAvailable_thenStatusOkAndRespondIt() {
```
oxisto, line 83: Just a minor thing: Could you get rid of the `_` in the test names? This applies to almost all files. So in this case the test function would be `testGetScanWhenRequestedScannerNotAvailableThenStatusOkAndRespondIt`. The other would probably be more readable but we should stick to common naming conventions, which do not have underscores in method names

clouditor-engine-core/src/test/java/io/clouditor/rest/DiscoveryResourceTest.java
```diff
@@ -0,0 +1,195 @@
...
+    // Request
+    Response response =
```
oxisto, line 85: Just FYI, we are using a fairly recent version of Java, so you could also just use `var` instead of explicit type declarations, but its of course not forbidden.

oxisto (changes requested):
Looks good except the method names

clouditor-engine-core/src/test/java/io/clouditor/rest/DiscoveryResourceTest.java
LeBogg: Yes, done.

clouditor-engine-core/src/test/java/io/clouditor/rest/DiscoveryResourceTest.java
LeBogg: Also done. At least for the Response declarations.

oxisto (approved):

PR #199, opened 2021-05-18
Initial API definitions

oxisto:
This will list initial API definitions according to the MEDINA framework.

Some open discussion points:

- [x] Format of identifiers? strings? numbers? Combination of both
- [x] Is a tool allowed to implicitly define new metrics in the registration process? No, it can only refer to existing metrics and the orchestrator needs an API to expose metric information (+ configuration of target values)
- [x] An `Evidence` needs a `body` obviously. It now has a `raw` body, containing the raw evidence, e.g. retrieved from the Cloud provider as well as a `resource` property, which is a semantic description of the resource according to our semantic schema
- [x] Provide basic `cl` (CLI) commands for the most common API endpoints, even though they are not implemented yet
- [x] Use [`google.protobuf.Value`](https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#value) for dynamically typed fields, e.g. the range


subham-deepsource:
Hey, @oxisto! I'm Subham from DeepSource. We tried reaching you on chat and also on email, but it went undeliverable to your email address. So, messaging you here.

I'd request you to check your support chat and let us know if it's now clear why it's failing. Thank you for using DeepSource.

oxisto:
It at least contains all relevant endpoints for now, but we will probably need more refinement of those API endpoints

PR #221, opened 2021-05-27
Azure compute discovery

garuppel:
First Draft for Azure compute Discovery.


service/discovery/azure/compute_azure.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func (d *azureComputeDiscovery) Description() string {
+	return "Discovery Azure compute."
+}
+
+// Discover compute resources
+func (d *azureComputeDiscovery) List() (list []voc.IsResource, err error) {
+	// create an authorizer from env vars or Azure Managed Service Identity
+	authorizer, err := auth.NewAuthorizerFromCLI()
```
oxisto, line 58: Could we already prepare to move this into a dedicated authorise function? maybe even shared with the `azureStorageDiscovery`. Probably also combine those two structs into just a `azureDiscovery`?

PR #234, opened 2021-06-04
Kubernetes resource discovery

oxisto:


service/discovery/k8s/compute.go
```diff
@@ -0,0 +1,57 @@
+package k8s
```
garuppel, line 1: Missing copyright header

service/discovery/k8s/compute_test.go
```diff
@@ -0,0 +1,36 @@
+package k8s_test
```
garuppel, line 1: Missing copyright header

service/discovery/k8s/k8s.go
```diff
@@ -0,0 +1,46 @@
+package k8s
```
garuppel, line 1: Missing copyright header

voc/compute.go
```diff
@@ -54,6 +54,10 @@ func (v *VirtualMachineResource) GetLog() *Log {
 	return v.Log
 }
 
+type ContainerResource struct {
+	ComputeResource
```
garuppel, line 58: Should the required relationships already be entered here as comment, so that we only have to comment them out and fill in later?

service/discovery/k8s/network.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (d *k8sNetworkDiscovery) Name() string {
+	return "Kubernetes Network"
+}
+
+func (d *k8sNetworkDiscovery) Description() string {
+	return "Discover Kubernetes network resources."
+}
+
+func (k k8sNetworkDiscovery) List() ([]voc.IsResource, error) {
```
garuppel, line 29: If the ingress controller is a load balancer, than the AccessRestriction and the url relationship is missing from the Ontology.

garuppel (commented):
First review, with minor findings: 
- missing copyright header 
- missing of some relationsships of the ontology

service/discovery/k8s/network_test.go
```diff
@@ -0,0 +1,92 @@
+package k8s_test
```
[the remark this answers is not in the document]
oxisto: Interestingly enough the copyright check script did not complain, looks like it looks for the wrong path.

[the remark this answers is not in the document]
oxisto: I will fix this in a different commit

service/discovery/k8s/network.go
oxisto: Ingress is now a loadbalancer. Loadbalancer is now a network service

PR #243, opened 2021-06-08
Added simple endpoint to view assessment results

oxisto:
- Adds an endpoint to view assessment results
- Refines the query discovery endpoint to include an optional query parameter


cli/commands/assessment/assessment_test.go
```diff
@@ -0,0 +1,135 @@
+package assessment_test
```
garuppel, line 1: Missing copyright header ;)

cli/commands/assessment/assessment_test.go
```diff
@@ -0,0 +1,135 @@
...
+
```
garuppel, line 2: Missing copyright header :)

cli/commands/assessment/assessment_test.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	// TODO(oxisto): refactor the next lines into a common test login function
+	dir, err = ioutil.TempDir(os.TempDir(), ".clouditor")
+	if err != nil {
+		panic(err)
+	}
+
+	viper.Set("username", "clouditor")
+	viper.Set("password", "clouditor")
```
garuppel, line 130: Currently this is ok, but in the future the password should be read from a config file

oxisto: Why? these are only for the unit tests.

garuppel (commented):
Only minor things

voc/voc.go
```diff
@@ -62,6 +63,17 @@ func (r *Resource) GetType() []string {
 	return r.Type
 }
 
+func (r *Resource) HasType(t string) (b bool) {
```
garuppel, line 66: Better use names as variables

garuppel (commented):
Use names as variables

PR #249, opened 2021-06-10
Add error handling

garuppel:


garuppel:
> I think I need to revise the code again. Currently, the network interfaces and load balancers are not discovered if the virtual machine discovery fails.

DONE

service/discovery/azure/compute.go
```diff
@@ -70,10 +70,31 @@ func (d *azureComputeDiscovery) List() (list []voc.IsResource, err error) {
 	}
 
 	// Discover virtual machines
+	virtualMachines, _ := d.discoverVirtualMachines()
```
oxisto, line 73: Hm. This basically ignores the error in a higher level than before :)

Should be split this into individual discoverers?

service/discovery/azure/compute.go
```diff
@@ -99,12 +129,20 @@ func (d *azureComputeDiscovery) List() (list []voc.IsResource, err error) {
 		list = append(list, s)
 	}
 
-	// Discover Load Balancer
-	// TODO Client to get load balancer
+	return list, err
+}
+
+// Discover Load Balancer
+func (d *azureComputeDiscovery) discoverLoadBalancer() ([]voc.IsResource, error) {
```
oxisto, line 136: I have moved load balancers out of the `Compute` resource, so a separate discoverer would be more appropriate.

oxisto (changes requested):

service/discovery/azure/compute.go
garuppel: Why? The error was already handled in the discoverVirtualMachines() function.

oxisto (approved):
LGTM.

Would be nice if you could not modify the pb.go files. Apart from that looks good.

PR #252, opened 2021-06-16
AWS S3 Discovery

oxisto:


LeBogg:
The branch is ready for PR. Some ToDo's and commented-out blocks are still in there to review. Nevertheless, the code works as it is now and test coverage of the new files is >90%.

service/discovery/aws/storage.go
```diff
@@ -0,0 +1,382 @@
...
+	"github.com/aws/smithy-go"
+	"time"
+)
+
+// awsS3Discovery handles the AWS API requests regarding the S3 service
+// ToDo: Generalize from s3 to storage for including other types, like EFS -> Probably storage folder with 3 .go files
+type awsS3Discovery struct {
+	client        S3API
+	isDiscovering bool
+	buckets       []bucket
```
oxisto, line 49: is there a reason to store the results here? the discovery should not do that in my opinion

service/discovery/aws/storage.go
```diff
@@ -0,0 +1,382 @@
...
+}
+
+// bucket contains meta data about a S3 bucket
+type bucket struct {
```
oxisto, line 53: Is there no AWS structure for this?

service/discovery/aws/storage.go
```diff
@@ -0,0 +1,382 @@
...
+		optFns ...func(*s3.Options)) (*s3.ListBucketsOutput, error)
+}
+
+// getBuckets returns all buckets
+func (d *awsS3Discovery) getBuckets() (err error) {
+	log.Println("Getting buckets in s3...")
+	var resp *s3.ListBucketsOutput
+	resp, err = d.client.ListBuckets(context.TODO(), &s3.ListBucketsInput{})
+	if err != nil {
+		log.Error("Error occurred.")
```
oxisto, line 176: This is not a very good log line :) You can avoid logging the error alltogether here, this will be done by the calling function in the discovery service.

service/discovery/aws/storage.go
```diff
@@ -0,0 +1,382 @@
...
+	}
+	var region string
+	log.Printf("Retrieved %v buckets.", len(resp.Buckets))
+	for _, b := range resp.Buckets {
+		//d.buckets = resp
+		region, err = d.getRegion(aws.ToString(b.Name))
+		if err != nil {
+			return
+		}
+		d.buckets = append(d.buckets, bucket{
```
oxisto, line 191: why do you store it to the `buckets` field of the discoverer?

service/discovery/aws/storage.go
```diff
@@ -0,0 +1,382 @@
...
+
+// S3ListBucketsAPI is the interface for the List function (used for mock testing)
+type S3ListBucketsAPI interface {
+	ListBuckets(ctx context.Context,
+		params *s3.ListBucketsInput,
+		optFns ...func(*s3.Options)) (*s3.ListBucketsOutput, error)
+}
+
+// getBuckets returns all buckets
+func (d *awsS3Discovery) getBuckets() (err error) {
```
oxisto, line 171: Shouldn't you return a list of buckets here?

service/discovery/aws/storage.go
```diff
@@ -0,0 +1,382 @@
...
+			endpoint:     "https://" + aws.ToString(b.Name) + ".s3." + region + ".amazonaws.com",
+			// ToDo: Implement method for retrieving the number of objects per bucket (if needed)
+			numberOfObjects: -1,
+		})
+	}
+	return
+}
+
+// getEncryptionAtRest gets the bucket's encryption configuration
+func (d *awsS3Discovery) getEncryptionAtRest(bucket string) (e voc.AtRestEncryption, err error) {
```
oxisto, line 205: You should return a pointer to `voc.AtRestEncryption` here, to avoid copying the struct

service/discovery/aws/storage.go
```diff
@@ -0,0 +1,382 @@
...
+	resp, err = d.client.GetBucketEncryption(context.TODO(), &input)
+	if err != nil {
+		log.Error("Error occurred")
+		var oe *smithy.OperationError
+		if errors.As(err, &oe) {
+			log.Printf("failed to call service: %s, operation: %s, error: %v", oe.Service(), oe.Operation(), oe.Unwrap())
+		}
+		return
+	}
+	log.Println("Bucket is encrypted.")
```
oxisto, line 222: remove

service/discovery/aws/storage.go
```diff
@@ -0,0 +1,382 @@
...
+	log.Printf("Checking encryption for bucket %v.\n", bucket)
+	input := s3.GetBucketEncryptionInput{
+		Bucket:              aws.String(bucket),
+		ExpectedBucketOwner: nil,
+	}
+	var resp *s3.GetBucketEncryptionOutput
+
+	resp, err = d.client.GetBucketEncryption(context.TODO(), &input)
+	if err != nil {
+		log.Error("Error occurred")
```
oxisto, line 215: see above. weird log line

service/discovery/aws/storage.go
```diff
@@ -0,0 +1,382 @@
...
+		return
+	}
+	log.Println("Bucket is encrypted.")
+	e.Algorithm = string(resp.ServerSideEncryptionConfiguration.Rules[0].ApplyServerSideEncryptionByDefault.SSEAlgorithm)
+	if e.Algorithm == string(types.ServerSideEncryptionAes256) {
+		e.KeyManager = "SSE-S3"
+	} else {
+		e.KeyManager = "SSE-KMS"
+	}
+	e.Enabled = true
```
oxisto, line 229: it's always encrypted?

service/discovery/aws/storage.go
```diff
@@ -0,0 +1,382 @@
...
+// getBuckets returns all buckets
+func (d *awsS3Discovery) getBuckets() (err error) {
+	log.Println("Getting buckets in s3...")
+	var resp *s3.ListBucketsOutput
+	resp, err = d.client.ListBuckets(context.TODO(), &s3.ListBucketsInput{})
+	if err != nil {
+		log.Error("Error occurred.")
+		var oe *smithy.OperationError
+		if errors.As(err, &oe) {
+			log.Printf("failed to call service: %s, operation: %s, error: %v", oe.Service(), oe.Operation(), oe.Unwrap())
```
oxisto, line 179: rather than log printing it here, you could return a new error with `fmt.Errorf` that contains additional details where the error occurred.

service/discovery/aws/aws_test.go
```diff
@@ -0,0 +1,91 @@
...
+			ConfigSources:    nil,
+			APIOptions:       nil,
+			Logger:           nil,
+			ClientLogMode:    0,
+		}
+		return
+	}
+	client, err := NewClient()
+	if err != nil {
+		t.Error("EXPECTED no error but GOT one.")
```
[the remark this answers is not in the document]
LeBogg: Thanks, nice to know. I will look into that

service/discovery/aws/storage.go
LeBogg: There is one API call that returns all buckets, or better say, bucket names. And other API calls, like getRegion(bucket string), require the name of the bucket to retrieve the appropriate values for the response. The idea is (as you can see in the List() method): Get all bucket names with getBuckets, store them in the struct awsS3Discovery. And the other API methods having this struct as receiver can perform their operations by accessing the names in the struct receiver.

But as I think about: Since only List() is exported and relevant for the discoverer, we may could also store the names in a local variable in List().

service/discovery/aws/storage.go
LeBogg: No, at least not with exact all the information we need for our purpose (like the ARN which we have to build out of context)

service/discovery/aws/storage.go
LeBogg: See comment above: the API call ListBucekts gets only the names and createionTime. And is, currently, stored in the method receiver d (awsS3Discovery).

service/discovery/aws/storage.go
LeBogg: I wasn't sure whether the if-statement in line 178 is always true or what is if it is not. If there is a case it is not true, then there would be at least a message that an error occurred and one could further investigate it.
But I could put a return statement withing the if block and then add this error log directly afterwards so that there will be an error message in any case (and yes, then also with the message content of err itself).
I wanted to log it here, since I want to keep the AWS stuff (like smithy.OperationError) out of the general discovery and keep it coupled to this package here (so there is no import of aws stuff in the general discoverer).

But, yes, we can just return the error and handle it in the caller.

service/discovery/aws/storage.go
LeBogg: Right, good hint. that would also help to keep aws stuff out of the discoverer. I will look into that.

service/discovery/aws/storage.go
LeBogg: See above: To have the names and relevant data in that struct which can be then used in the List() function to iterate through all buckets.

service/discovery/aws/storage.go
LeBogg: The way the AWS API call works: Either encryption is in place and returned in the response field; or there is no encryption and the API call returns an error. In the case of error, the struct field will be kept as initialized with its zero values (==false). But I can add a comment to make that more clear.

But as I think about it, we also have to distinguish then between errors that are API Call failures (which should be propagated to the caller) and errors that are just no-encryption-is-set-errors. I have to look into that and maybe handle it via a switch statement.

go.mod
```diff
@@ -40,4 +45,5 @@ require (
 	k8s.io/client-go v0.21.2
 )
 
+// ToDo: In Goland: Missing dependency?
```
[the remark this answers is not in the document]
LeBogg: yes, referred to the `replace` line

service/discovery/aws/storage.go
lebogg: The go wiki states to not "pass pointers as function arguments just to save a few bytes" and only use pointers for "large structs, or even small structs that might grow". Is `voc.AtRestEncryption` really that big?

https://github.com/golang/go/wiki/CodeReviewComments#pass-values

oxisto: The point being "small structs that might grow". We do not know how big the voc objects will be in the future, potentially even having pointers to other objects, so I would argue that a pointer is the better choice here.

service/discovery/aws/aws_test.go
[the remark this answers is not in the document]
oxisto: This one is still open

PR #261, opened 2021-06-22
Extend compute load balancer

garuppel:
Fill out the compute load balancer properties


garuppel:
Which information needs the AccessRestriction. What exactly needs the property inbound and restrictedPorts in the case of a Azure load balancer?

oxisto:
> @oxisto @immqu What information should AccessRestriction have? What exactly needs the property inbound and restrictedPorts in the case of a Azure load balancer?

Not quite sure to be honest. I think we never properly filled this property. Not sure, if the load balancer really has this property, I think this is more related to a network policy of a compute resource.

garuppel:
> > @oxisto @immqu What information should AccessRestriction have? What exactly needs the property inbound and restrictedPorts in the case of a Azure load balancer?
> 
> Not quite sure to be honest. I think we never properly filled this property. Not sure, if the load balancer really has this property, I think this is more related to a network policy of a compute resource.

And what inbound information do we need for the network interface?

garuppel:
@oxisto @immqu What inbound information do we need for the network interface?

garuppel:
It is only a minor change. Before we can delete the last TODOs, we need to clarify some points: 
- Do we distinguish between osDisk and dataDisks in VMs?
- Has the Load Balancer restricted ports?
- Do we need the httpEndpoint for the load balancer?
- What information should AccessRestriction have?
- What inbound information do we need for the network interface?

oxisto:
> It is only a minor change. Before we can delete the last TODOs, we need to clarify some points:
> 
> * Do we distinguish between osDisk and dataDisks in VMs?
> * Has the Load Balancer restricted ports?
> * Do we need the httpEndpoint for the load balancer?
> * What information should AccessRestriction have?
> * What inbound information do we need for the network interface?

Good points. Let's discuss them in the Jour-Fixe tomorrow.

PR #264, opened 2021-06-26
Building a resource graph

oxisto:


garuppel:
I´ve put the blockStorage references to the VirtualMachineResources, based on the Ontology. I did not find any storages in the pod object.

garuppel:
Currently the k8s discovery fills the ContainerResource with pods. Shouldn't we change this and take the containers, since pods can contain multiple containers?

service/discovery/azure/network.go
```diff
@@ -78,7 +78,7 @@ func (d *azureNetworkDiscovery) List() (list []voc.IsResource, err error) {
 	// Discover Load Balancer
 	loadBalancer, err := d.discoverLoadBalancer()
 	if err != nil {
-		return nil, fmt.Errorf("could not discover load balancer: %w", err)
+		return list, fmt.Errorf("could not discover load balancer: %w", err)
```
garuppel, line 81: Here we have to return the list, otherwise the discovered networks are ignored if there is an error in the ingress discovery.

service/discovery/azure/compute.go
```diff
@@ -121,7 +121,7 @@ func (d *azureComputeDiscovery) handleVirtualMachines(vm *compute.VirtualMachine
 
 	vmExtended, err := d.getExtendedVirtualMachine(vm)
 	if err != nil {
-		return nil, fmt.Errorf("could not get virtual machine: %w", err)
+		return nil, err
```
garuppel, line 124: Don't we have to pass on the error here instead of creating a new one?  If yes, then we need to check the whole code again.

service/discovery/k8s/compute.go
```diff
@@ -70,14 +70,23 @@ func (k k8sComputeDiscovery) List() ([]voc.IsResource, error) {
 }
 
 func (k k8sComputeDiscovery) handlePod(pod *v1.Pod) voc.IsCompute {
-	return &voc.ContainerResource{
+	r := &voc.ContainerResource{
 		ComputeResource: voc.ComputeResource{
 			Resource: voc.Resource{
-				ID:           fmt.Sprintf("/namespaces/%s/containers/%s", pod.Namespace, pod.Name),
+				ID:           voc.ResourceID(getContainerResourceID(pod)),
```
garuppel, line 76: The ContainerResource is filled with pods, shouldn´t we fill it with containers instead?

garuppel (commented):
Sorry, I forgot to finish the review.

garuppel (commented):
Reviewed

PR #276, opened 2021-07-08
Add azure function discovery

garuppel:


service/discovery/azure/compute.go
```diff
@@ -74,9 +75,52 @@ func (d *azureComputeDiscovery) List() (list []voc.IsResource, err error) {
...
+	list = append(list, function...)
+
 	return
 }
 
+// Discover function
+func (d *azureComputeDiscovery) discoverFunction() ([]voc.IsResource, error) {
+	var list []voc.IsResource
+
+	client := web.NewAppsClient(to.String(d.sub.SubscriptionID))
```
oxisto, line 92: Are you sure, this is the correct call for Azure functions? Isn't this referring to https://azure.microsoft.com/en-us/services/app-service/web/?

garuppel: honestly I do not know :) The call above gets the Function App shown here: https://portal.azure.com/#blade/HubsExtension/BrowseResource/resourceType/Microsoft.Web%2Fsites/kind/functionapp

I don´t know if and what is the difference between Azure Functions and Function Apps.

garuppel: You´re right, functions are created in function apps. Ok, than I have to extend that.

oxisto: It looks like it is the correct call, but it seems they are grouped in "App Service Plans", they seem to be be a container for those apps, but I think we can ignore that for now, so it seems that everything looks good for now.

oxisto (approved):
LGTM

oxisto:
Ah... you branched it off from the `resource-graph` branch instead of `main`, @garuppel, mistake?

garuppel:
> Ah... you branched it off from the `resource-graph` branch instead of `main`, @garuppel, mistake? Can you re-open a PR/branch from main?
No, I thought that the functions also have references and wanted to take the branch with the implemented references stuff.

PR #286, opened 2021-07-29
Usage of generated voc files based on the ontology

garuppel:
In this PR the generated voc are used with two exceptions: 
1. I deleted the generated file CloudResource.go because the needed information are in the file voc.go 
2. I changed the generated file LoadBalancer.go and deleted the "NetworkService []ResourceID" because otherwise there is a conflict (redeclared). How do we name the variable?
3. Same as point 2 for generated file ProxiedEndpoint.go


service/discovery/aws/storage_test.go
```diff
@@ -334,14 +335,14 @@ func TestGetEncryptionAtRest(t *testing.T) {
 	assert.Nil(t, err)
 	assert.True(t, encryptionAtRest.Enabled)
 	assert.Equal(t, "AES256", encryptionAtRest.Algorithm)
-	assert.Equal(t, "SSE-S3", encryptionAtRest.KeyManager)
+	assert.Equal(t, "SSE-S3", encryptionAtRest.Keymanager)
```
oxisto, line 338: We probably should rename this to `KeyManager` in the ontology file.

oxisto (approved):
I will reset the pb.go files back to main. Otherwise, LGTM. Some comments wrt to the Ontology itself

PR #290, opened 2021-08-03
Discovery of AWS compute resources

lebogg:
Adding compute services of AWS (EC2 and Lambda) to discovery.


service/discovery/aws/aws.go
```diff
@@ -39,20 +42,52 @@ var log = logrus.WithField("component", "aws-discovery")
...
 // NewClient constructs a new AwsClient
 // TODO(lebogg): "Overload" (switch) with staticCredentialsProvider
 func NewClient() (*Client, error) {
 	c := &Client{}
+
+	// load configuration
 	cfg, err := loadDefaultConfig(context.TODO())
 	if err != nil {
-		log.Errorf("Could not load default config: %v", err)
+		return c, err
```
oxisto, line 69:
```suggestion
		return nil, fmt.Errorf("could not load config: %w", err)
```

service/discovery/aws/aws.go
```diff
@@ -39,20 +42,52 @@ var log = logrus.WithField("component", "aws-discovery")
...
 	}
 	c.Cfg = cfg
+
+	// load accountID
+	stsClient := newFromConfigSTS(cfg)
+	resp, err := stsClient.GetCallerIdentity(context.TODO(), &sts.GetCallerIdentityInput{})
+	if err != nil {
+		return c, err
```
oxisto, line 77:
```suggestion
		return nil, fmt.Errorf("could not get caller identity: %w", err)
```

service/discovery/aws/compute.go
```diff
@@ -0,0 +1,243 @@
...
+// Name is the method implementation defined in the discovery.Discoverer interface
+func (d computeDiscovery) Name() string {
+	return "AWS Compute"
+}
+
+// List is the method implementation defined in the discovery.Discoverer interface
+func (d computeDiscovery) List() (resources []voc.IsCloudResource, err error) {
+	listOfVMs, err := d.discoverVirtualMachines()
+	if err != nil {
+		return
```
oxisto, line 89:
```suggestion
		return nil, fmt.Errorf("could not discover virtual machines: %w", err)
```

I would wrap the error here, so one knows that it failed while listing virtual machines

service/discovery/aws/compute.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+	for _, machine := range listOfVMs {
+		resources = append(resources, &machine)
+	}
+
+	listOfFunctions, err := d.discoverFunctions()
+	if err != nil {
+		return
```
oxisto, line 97:
```suggestion
		return nil, fmt.Errorf("could not discover virtual functions: %w", err)
```

I would wrap the error here, so one knows that it failed while listing functions

service/discovery/aws/compute.go
```diff
@@ -0,0 +1,243 @@
...
+	var resources []voc.VirtualMachine
+	for _, reservation := range resp.Reservations {
+		for i := range reservation.Instances {
+			vm := &reservation.Instances[i]
+			computeResource := &voc.Compute{
+				CloudResource: &voc.CloudResource{
+					ID:   d.getARNOfVM(vm),
+					Name: d.getNameOfVM(vm),
+					// TODO(all): -1 or 0 as default value when no creation time is available?
+					CreationTime: -1,
```
oxisto, line 125: `0` is probably easier to catch. Even better would be, if creation time would be an actual `*time.Time` so we can just set it to `nil`. However, this would require a change in the ontology generation.

service/discovery/aws/compute.go
```diff
@@ -0,0 +1,243 @@
...
+				BlockStorage:     d.getBlockStorageIDsOfVM(vm),
+				Log:              d.getLogsOfVM(vm),
+			})
+		}
+	}
+	return resources, nil
+}
+
+// discoverFunctions discovers all lambda functions
+// TODO(all): FunctionVersion in input to ALL ok? (=all versions of a lambda functions are discovered)
```
oxisto, line 142: I would say that only the last version of the function would suffice.

service/discovery/aws/compute.go
```diff
@@ -0,0 +1,243 @@
...
+
+// computeDiscovery handles the AWS API requests regarding the computing services (EC2 and Lambda)
+type computeDiscovery struct {
+	virtualMachineAPI EC2API
+	functionAPI       LambdaAPI
+	isDiscovering     bool
+	awsConfig         *Client
+}
+
+// EC2API describes the EC2 api interface (for mock testing)
```
oxisto, line 51: I would remove the "for mock testing" comment from here and also the others, it is a bit confusing or maybe make the comment more detailed, i.e. that this conforms to the API used by the official AWS client but can be used to mock it or something like that.

service/discovery/aws/compute.go
```diff
@@ -0,0 +1,243 @@
...
+				Log:              d.getLogsOfVM(vm),
+			})
+		}
+	}
+	return resources, nil
+}
+
+// discoverFunctions discovers all lambda functions
+// TODO(all): FunctionVersion in input to ALL ok? (=all versions of a lambda functions are discovered)
+// TODO(oxisto): Is this a good approach with "for hasNextMarker" loop (I would use overloaded methods - not in Go)
```
oxisto, line 143: Not sure, it seems to be a little bit of copy/paste code. Can we combine the two ListFunctions calls some how?

service/discovery/aws/compute.go
```diff
@@ -0,0 +1,243 @@
...
+		}
+		resources = append(resources, d.getFunctionResources(resp.Functions)...)
+		hasNextMarker = resp.NextMarker != nil
+	}
+
+	return resources, nil
+}
+
+// getFunctionResources iterates functionConfigurations and returns a list of corresponding FunctionResources
+func (d *computeDiscovery) getFunctionResources(functions []types2.FunctionConfiguration) (resources []voc.Function) {
```
oxisto, line 184: `getFunctionResources` does not sound like good name for this. Maybe something more like `map` or `translate`?

service/discovery/aws/compute.go
```diff
@@ -0,0 +1,243 @@
...
+import (
+	"clouditor.io/clouditor/api/discovery"
+	"clouditor.io/clouditor/voc"
+	"context"
+	"errors"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"github.com/aws/aws-sdk-go-v2/service/lambda"
+	types2 "github.com/aws/aws-sdk-go-v2/service/lambda/types"
```
oxisto, line 39: The usual approach here is to use something like `lambda_types`

service/discovery/aws/aws.go
```diff
@@ -39,20 +42,52 @@ var log = logrus.WithField("component", "aws-discovery")
 // loadDefaultConfig holds config.LoadDefaultConfig() so the test function can mock it
 var loadDefaultConfig = config.LoadDefaultConfig
 
+// newFromConfigSTS holds sts.NewFromConfig() so the test function can mock it
+var newFromConfigSTS = loadSTSClient
+
 // Client holds configurations across all services within AWS
 // TODO(lebogg): deepsource.io wants the struct to exported since NewAwsStorageDiscovery is exported. Encapsulation?
 type Client struct {
 	Cfg aws.Config
```
oxisto, line 51: Any particular reason why the client Cfg is exposed?

oxisto (commented):
A few comments here and there

service/discovery/aws/aws.go
lebogg: Since it is an error within an AWS service call, I will wrap it with AWS own  smithy.APIError interface type.

service/discovery/aws/compute.go
lebogg: Yes, but without "virtual", right?

service/discovery/aws/compute.go
lebogg: Yes, I agree with `*time.Time` as better option. It is also the type returned by AWS (most of the times), e.g. for launchTime. But for now I will set it to `0` - I also don't think someone could create an AWS virtual machine in 1970.

service/discovery/aws/compute.go
lebogg: The only reason for this interface to exist is for mock testing. Otherwise, I think, building an interface like that has no real advantage. That's why I want to put a comment there to clarify why we do it.
Maybe: "...interface which is implemented by the official AWS client and mock clients in tests"?

service/discovery/aws/compute.go
lebogg: I found a possible solution: I loop from the beginning (`for true`) and in the first execution, the `Marker` in the input of `ListFunctions` is set to nil. At the end of the loop it is checked if the response has a `Marker` returned: If yes, execute `ListFunctions` again with the respective `Marker` and break, otherwise.

service/discovery/aws/compute.go
lebogg: Yes, `map` is better. Also will change similar methods.

service/discovery/aws/compute.go
lebogg: I will use `typesLambda` and `typesEC2` (no snake case)

service/discovery/aws/aws.go
lebogg: Not really. In the package `discovery` we cannot access it then. But it is not needed. I changed it and the respective lines of code.

service/discovery/aws/aws.go
```diff
@@ -29,30 +29,69 @@ package aws
...
 // NewClient constructs a new AwsClient
 // TODO(lebogg): "Overload" (switch) with staticCredentialsProvider
 func NewClient() (*Client, error) {
 	c := &Client{}
+
+	// load configuration
 	cfg, err := loadDefaultConfig(context.TODO())
 	if err != nil {
-		log.Errorf("Could not load default config: %v", err)
+		return nil, fmt.Errorf("could not load default config: %v", err)
```
oxisto, line 69:
```suggestion
		return nil, fmt.Errorf("could not load default config: %w", err)
```

oxisto (approved):

PR #294, opened 2021-08-04
Fix voc files

garuppel:
Changes in this PR: 
- voc file naming changed from camleCase to snake_case
- variable names of Slices changed to plural (in case of 'y' to 'ies': functionality - functionalities)


oxisto (approved):
Very nice! LGTM

PR #332, opened 2021-10-04
Add region to cloud resources

garuppel:
Added region to cloud resources and to voc.go. 

@oxisto do pods have a location?
@lebogg can you please fix the test TestComputeDiscovery_discoverFunctions.


oxisto:
> Added region to cloud resources and to voc.go.
> 
> @oxisto do pods have a location? @lebogg can you please fix the test TestComputeDiscovery_discoverFunctions.

Do we know in which cluster the pod is running? Then I would say we would derive the location from the cluster.

garuppel:
For now I´m done with the PR, but the k8s container region is still missing.

Issue #382, opened 2021-11-23
`PUT` in `/assessment_results` should be a `POST`


oxisto:
https://github.com/clouditor/clouditor/blob/5c599795d87e6dbeb538632946453784e1c57fea/proto/orchestrator.proto#L87

PR #406, opened 2021-12-06
Fix API

### LINKED ISSUE #383: Evidence Store API does not have REST options yet
See https://github.com/clouditor/clouditor/blob/5c599795d87e6dbeb538632946453784e1c57fea/proto/evidence_store.proto

lebogg:
Fix #383 and #382


proto/assessment.proto
```diff
@@ -39,33 +39,30 @@ option go_package = "api/assessment";
...
-      post : "/v1/assessment/results"
+      get : "/v1/assessment/results"
       response_body : "*"
     };
   }
 
-  // Assesses the evidence sent by discovery. Part of the public API,
-  // also exposed as REST
+  // Assesses the evidence sent by the discovery. Part of the public API, also exposed as REST.
+  // TODO(all): Evidence instead of AssessEvidenceRequest?
```
lebogg, line 56: If request only contains one item/message, use it directly instead of creating a wrapper #googleapiconventions? @oxisto  @garuppel

proto/evidence_store.proto
```diff
@@ -29,20 +29,33 @@ syntax = "proto3";
...
 // Manages the storage of evidences
 service EvidenceStore {
-  // Stores an evidence to the evidence storage
-  rpc StoreEvidence(Evidence) returns (StoreEvidenceResponse);
+  // Stores an evidence to the evidence storage. Part of the public API, also exposed at REST.
+  rpc StoreEvidence(Evidence) returns (StoreEvidenceResponse) {
+    option (google.api.http) = {
+      post : "/v1/evidence_store/evidence"
+      body : "evidence"
+      // TODO(all): What is the difference between this response_body ("*") and to don't specify it at all
```
lebogg, line 44: @oxisto

proto/assessment.proto
oxisto: Doesn't this only relate to responses?

garuppel: > 

The following is from the Google API design guide (also in the proto README.md).

"For custom methods, they should have their own XxxResponse messages even if they are empty, because it is very likely their functionality will grow over time and need to return additional data." ([Link](https://cloud.google.com/apis/design/design_patterns#empty_responses))

garuppel: > 

Regarding the request message I didn´t find any information, but most examples have a dedicated request message with one field.

lebogg: [Request and response messages (API designe guide)](https://cloud.google.com/apis/design/naming_convention#request_and_response_messages):
> The request and response messages for RPC methods should be named after the method names with the suffix Request and Response, respectively, unless the method request or response type is:
    - an empty message (use google.protobuf.Empty),
    - a resource type, or
    - a resource representing an operation 

But see also [Empty Responses (API designe guide)](https://cloud.google.com/apis/design/design_patterns#empty_responses):
> For custom methods, they should have their own XxxResponse messages even if they are empty, because it is very likely their functionality will grow over time and need to return additional data

Thus, I would say XxxResponse for response messages and resource type for request if possible. But we can do that in another PR as well; since such changes would affect the code directly. And we want to keep our PRs small :)

garuppel: I would use response and request messages as long as we don´t know if they are changing again. So, we do not have to rebuild everything if we are changing the messages.  :)

lebogg: We do not use XxxRequest messages in `rpc AssessEvidences(stream Evidence)` and in the `evidence.proto` file. Should I change them in this PR?

garuppel: I would change it in this PR. 

But I would wait for Christian's opinion before you change it back again :smile:

oxisto: Let's discuss this briefly in our JF today. I personally would follow the example here: https://cloud.google.com/apis/design/naming_convention#method_names. Which basically uses `Request` all the the time (probably for forwards compatibility) but omits `Response` for single resource, such as `Book` in the example for convenience.

proto/discovery.proto
```diff
@@ -36,7 +36,7 @@ option go_package = "api/discovery";
 // Discovers cloud resources and maps these properties to the format of the
 // ontology
 service Discovery {
-  // Starts discovering the cloud resources
+  // Starts discovering the cloud resources, exposed at REST.
```
oxisto, line 39:
```suggestion
  // Starts discovering the cloud resources, exposed as REST.
```

proto/discovery.proto
```diff
@@ -45,7 +45,7 @@ service Discovery {
     };
   };
 
-  // Lists all evidences collected in the last run
+  // Lists all evidences collected in the last run, exposed at REST.
```
oxisto, line 48:
```suggestion
  // Lists all evidences collected in the last run, exposed as REST.
```

proto/evidence_store.proto
```diff
@@ -29,20 +29,33 @@ syntax = "proto3";
...
 import "google/protobuf/empty.proto";
 
 option go_package = "api/evidence";
 
 // Manages the storage of evidences
 service EvidenceStore {
-  // Stores an evidence to the evidence storage
-  rpc StoreEvidence(Evidence) returns (StoreEvidenceResponse);
+  // Stores an evidence to the evidence storage. Part of the public API, also exposed at REST.
+  rpc StoreEvidence(Evidence) returns (StoreEvidenceResponse) {
```
oxisto, line 40: Shoudn't this now be a `StoreEvidenceRequest`?

proto/evidence_store.proto
```diff
@@ -29,20 +29,33 @@ syntax = "proto3";
...
+    option (google.api.http) = {
+      post : "/v1/evidence_store/evidence"
+      body : "*"
+      // TODO(all): What is the difference between this response_body ("*") and to don't specify it at all
```
oxisto, line 44: Remove todo

proto/evidence_store.proto
```diff
@@ -29,20 +29,33 @@ syntax = "proto3";
...
+      response_body : "*"
+    };
+  }
 
-  // Stores a stream of evidences to the evidence storage
-  rpc StoreEvidences(stream Evidence) returns (google.protobuf.Empty);
+  // Stores a stream of evidences to the evidence storage. Part of the public API, not exposed at REST.
```
oxisto, line 49:
```suggestion
  // Stores a stream of evidences to the evidence storage. Part of the public API, not exposed as REST.
```

proto/orchestrator.proto
```diff
@@ -41,7 +41,7 @@ service Orchestrator {
       returns (AssessmentTool) {
     option (google.api.http) = {
       post : "/v1/orchestrator/assessment_tools"
-      body : "tool"
+      body : "*"
```
oxisto, line 44: I this case I think we said to keep it to `tool` for convience, so that REST users do not need the additional request wrappers, which is not intuitive

proto/orchestrator.proto
```diff
@@ -67,7 +67,7 @@ service Orchestrator {
       returns (AssessmentTool) {
     option (google.api.http) = {
       put : "/v1/orchestrator/assessment_tools/{tool_id}"
-      body : "tool"
```
oxisto, line 70: see above

proto/orchestrator.proto
```diff
@@ -84,8 +84,8 @@ service Orchestrator {
   rpc StoreAssessmentResult(StoreAssessmentResultRequest)
       returns (StoreAssessmentResultResponse) {
     option (google.api.http) = {
-      put : "/v1/orchestrator/assessment_results"
-      body : "result"
```
oxisto, line 88: see above

proto/evidence_store.proto
lebogg: Yes, wait for the new commit :)

proto/orchestrator.proto
```diff
@@ -51,14 +52,16 @@ service Orchestrator {
       returns (ListAssessmentToolsResponse) {
     option (google.api.http) = {
       get : "/v1/orchestrator/assessment_tools"
-      additional_bindings {get : "/v1/orchestator/{metric_id}/assessment_tools"}
+      additional_bindings {get : "/v1/orchestrator/{metric_id}/assessment_tools"}
+      response_body : "tools"
```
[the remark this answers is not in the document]
garuppel: Yes, you cannot use repeated fields.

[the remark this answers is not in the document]
garuppel: Yes, you cannot use repeated fields. But in this case a * should be sufficient, right?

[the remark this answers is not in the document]
lebogg: My goal was to get sth. like
`{tools: {id: "tool1",...}, {id: "tool2", ...},...}` 
instead of
`{ListAssessmentToolsResponse : tools: [{id: "tool1",...}, {id: "tool2", ...},...]}`

Looking in the openapi folder of the main branch, though, it seems that it automatically does the former one when `response_body` is omitted (being equal to `*`). I may do some tests with Postman, but I guess it is done right.

lebogg:
ToDo: Add some tests

oxisto (commented):
Just minor things

proto/auth.proto
```diff
@@ -36,6 +36,7 @@ service Authentication {
   rpc Login(LoginRequest) returns (LoginResponse) {
     option (google.api.http) = {
       post : "/v1/auth/login"
+      // TODO: Here, "*" is good since username and password (fields of LoginRequest) will be the corresponding fields, right?
```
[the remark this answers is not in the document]
lebogg: 1. In the HTTP case, I was wondering if username and password are wrapped within `LoginRequest` or if they are in the root level of the body (`{LoginRequest : { username: "", ...}` vs `{username: "", password ""}`).

2. `auth.proto` compilation does not include Open API output although it has HTTP options? Should I change that?

proto/orchestrator.proto
```diff
@@ -67,13 +68,15 @@ service Orchestrator {
       returns (AssessmentTool) {
     option (google.api.http) = {
       put : "/v1/orchestrator/assessment_tools/{tool_id}"
+      // TODO: Why do we have tool_id separated when it is in tool already?
```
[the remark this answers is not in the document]
lebogg: Ok, interesting :+1:

proto/auth.proto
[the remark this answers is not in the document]
oxisto: Leave it for now, we probably need to re-eng auth at some point anyway.

oxisto (approved):

Issue #371, opened 2021-11-16
Can not discover Azure resources after a certain amount of time


oxisto:
**Describe the bug**

It seems that we need to re-authenticate the Azure credentials after a certain point in time. 

```
ERRO[2419] Could not retrieve resources from discoverer 'Azure': could not discover IaC template: could not discover resource groups: resources.GroupsClient#List: Failure responding to request: StatusCode=401 -- Original Error: autorest/azure: Service returned an error. Status=401 Code="ExpiredAuthenticationToken" Message="The access token expiry UTC time '11/16/2021 12:13:44 PM' is earlier than current UTC time '11/16/2021 12:16:58 PM'."  component=discovery
```

**To Reproduce**

Leave Clouditor running for more than 30 minutes

**Expected behavior**

Clouditor should automatically re-authenticate.

oxisto:
According to @garuppel there might be a special flag in the Azure clients to do this.

PR #408, opened 2021-12-08
Azure Timout

garuppel:
Fix to Issue #371


garuppel:
Now the error message turned from 
`ERRO[2419] Could not retrieve resources from discoverer 'Azure': could not discover IaC template: could not discover resource groups: resources.GroupsClient#List: Failure responding to request: StatusCode=401 -- Original Error: autorest/azure: Service returned an error. Status=401 Code="ExpiredAuthenticationToken" Message="The access token expiry UTC time '11/16/2021 12:13:44 PM' is earlier than current UTC time '11/16/2021 12:16:58 PM'."  component=discovery` to 
grafik

garuppel:
ERRO[4093] Could not retrieve resources from discoverer 'Kubernetes Network': could not list services: Get "https://eks-dns-0b4f013e.hcp.westeurope.azmk8s.io:443/api/v1/services": http2: client connection lost  component=discovery
ERRO[4912] Could not retrieve resources from discoverer 'Kubernetes Compute': could not list ingresses: Get "https://eks-dns-0b4f013e.hcp.westeurope.azmk8s.io:443/api/v1/pods": net/http: TLS handshake timeout  component=discovery
panic: runtime error: index out of range [0] with length 0

garuppel:
The azure discoverer storage, compute, network and armTemplate are ok. No authorization or refresh token error. 

I think the problem is in the K8S discoverer.

Issue #384, opened 2021-12-01
Provide hooks for incoming evidences and assessment results


oxisto:
In the Clouditor, we need some kind of hooks that an external user of the Clouditor framework can implement, which are executed when new evidences are stored in the evidence store and when new assessment results arrive.

Potentially, we want multiple hooks

PR #410, opened 2021-12-10
Add hooks

garuppel:
Add hooks for incoming (#384)
- evidences in evidence_store -> StoreEvidence() / StoreEvidences()
- assessment_results in orchestrator -> StoreAssessmentResult() / StoreAssessmentResults()


cmd/engine/engine.go
```diff
@@ -178,6 +178,30 @@ func doCmd(_ *cobra.Command, _ []string) (err error) {
 	assessmentService = service_assessment.NewService()
 	evidenceStoreService = service_evidenceStore.NewService()
 
+	assessmentService.RegisterAssessmentResultHook(func(result *assessment.AssessmentResult, err error) {
```
lebogg, line 181: Just for my understanding: Why do we need an `error` as parameter of the hook function? Maybe just state a short example. Thanks!

cmd/engine/engine.go
```diff
@@ -178,6 +178,30 @@ func doCmd(_ *cobra.Command, _ []string) (err error) {
...
+
+		if err != nil {
+			log.Errorf("error was passed: %v", err)
+		}
+		_, err = orchestratorService.StoreAssessmentResult(context.Background(), &orchestrator.StoreAssessmentResultRequest{
```
lebogg, line 186: So every time the hook function is called, the assessment result is sent to the orchestrator? I would say the assessment result should be always forwarded to the orchestrator as soon it is created in the code of `assessment.go`.

cmd/engine/engine.go
```diff
@@ -178,6 +178,30 @@ func doCmd(_ *cobra.Command, _ []string) (err error) {
...
+			Result: result})
+
+		if err != nil {
+			log.Errorf("error storing assessment result in orchestrator: %v", err)
+		}
+	})
+
+	// TODO(all): register assessment result hook
+	orchestratorService.RegisterAssessmentResultHook(func(result *assessment.AssessmentResult, err error) {
```
lebogg, line 195: Is this a real ToDo for here in Clouditor? I thought, we just define here the hook and the actual function is implemented and passed in the Clouditor Wrapper of MEDINA.

cmd/engine/engine.go
```diff
@@ -178,6 +178,30 @@ func doCmd(_ *cobra.Command, _ []string) (err error) {
...
+		// hash assessment result and send it to the DLT
+		// send assessment result to the Continuous Certification Evaluation
+	})
+
+	// TODO(all): register evidence hook
+	evidenceStoreService.RegisterEvidenceHook(func(result *evidence.Evidence, err error) {
```
lebogg, line 201: See above

service/evidence/evidence_store.go
```diff
@@ -40,28 +44,43 @@ func (s *Service) StoreEvidence(_ context.Context, req *evidence.StoreEvidenceRe
 
 	_, err = e.Validate()
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "invalid req: %v", err)
+		return resp, status.Errorf(codes.InvalidArgument, "invalid req: %v", err)
```
lebogg, line 47: I'm not sure but if we have an error (evidence is not accepted), we only should return the error (without `resp`), right? When I think about it, maybe we could have left out the `status` in the response message then.

service/evidence/evidence_store.go
```diff
@@ -40,28 +44,43 @@ func (s *Service) StoreEvidence(_ context.Context, req *evidence.StoreEvidenceRe
...
 	return resp, err
 }
 
 // StoreEvidences is a method implementation of the evidenceServer interface: It receives evidences and stores them
 func (s *Service) StoreEvidences(stream evidence.EvidenceStore_StoreEvidencesServer) (err error) {
 	var (
-		req *evidence.StoreEvidenceRequest
-		e   *evidence.Evidence
+		req              *evidence.StoreEvidenceRequest
+		receivedEvidence *evidence.Evidence
```
lebogg, line 67: I like the shorter representation `e` more because it has a short scope. I like the convention of the length of variable names being proportional to the scope.

service/orchestrator/orchestrator.go
```diff
@@ -58,6 +60,12 @@ type Service struct {
 	// metricConfigurations holds a double-map of metric configurations associated first by service ID and then metric ID
 	metricConfigurations map[string]map[string]*assessment.MetricConfiguration
 
+	// Currently only in-memory
+	Results map[string]*assessment.AssessmentResult
```
lebogg, line 64: A thought of mine: Now that we have Results stored in the orchestrator I'm thinking about removing them from the assessment in the future and also "redirect" the CLI command `ListAssessmentResults` to the orchestrator? Unless it should be possible to use the Assessment without orchestrator (service notion).
Either way, we should provide somehow `ListAssessmentResults` in the orchestrator as well, since it can have results from multiple assessment tools

service/orchestrator/orchestrator.go
```diff
@@ -185,6 +194,60 @@ func (s *Service) ListMetricConfigurations(ctx context.Context, req *orchestrato
...
+	return
+}
+
+func (s *Service) StoreAssessmentResults(stream orchestrator.Orchestrator_StoreAssessmentResultsServer) (err error) {
+	var receivedAssessmentResult *assessment.AssessmentResult
+
+	for {
+		receivedAssessmentResult, err = stream.Recv()
+		if err == io.EOF {
+			log.Infof("Stopped receiving streamed evidences")
```
lebogg, line 217: copy pasta error? :)

api/assessment/assessment.go
```diff
@@ -0,0 +1,77 @@
...
+	//value := result.MetricData.GetTargetValue()
+	//if value == nil {
+	//	return "", ErrResourceNotStruct
+	//}
+
+	if result.EvidenceId == "" {
+		return "", ErrEvidenceIdMissing
+	}
+
+	if result.NonComplianceComments == "" {
```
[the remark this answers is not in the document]
oxisto: Yeah keep it optional

service/evidence/evidence_store.go
oxisto: Yeah, I would return an error here and not have a `status` in the response, this is redundant otherwise. Keep the response empty for now.

api/assessment/assessment.go
```diff
@@ -0,0 +1,77 @@
...
+
+	if result.MetricData.Operator == "" {
+		return "", ErrMetricDataOperatorMissing
+	}
+
+	if result.MetricData.TargetValue == nil {
+		return "", ErrMetricDataTargetValueMissing
+	}
+
+	// TODO(all): Do we have to check the target value type?
```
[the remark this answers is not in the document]
garuppel: Ok, in the api/evidence.go it is also checked. Is it also not necessary?

service/evidence/evidence_store.go
garuppel: Ok, than I´ll leave it as it is.

cmd/engine/engine.go
garuppel: That was an info for Immanuel, but I delete it before merging.

service/orchestrator/orchestrator.go
garuppel: @oxisto @lebogg I´ve added a new discussion: #422

cmd/engine/engine.go
garuppel: Depending on the previous code, you may already have an error and want to handle it differently. See service/assessment.go line 117 (updated file).

cmd/engine/engine.go
garuppel: But exactly that is happening in service/assessment.go line 149/178 (updated file) 

How am I able to reference to a particular line of code?

service/assessment/assessment.go
```diff
@@ -107,15 +107,24 @@ func (s Service) AssessEvidences(stream assessment.Assessment_AssessEvidencesSer
...
-
 	}
 }
 
 // handleEvidence is the common evidence assessment of AssessEvidence and AssessEvidences
 func (s Service) handleEvidence(evidence *evidence.Evidence) error {
 	resourceId, err := evidence.Validate()
 	if err != nil {
-		return fmt.Errorf("invalid evidence: %w", err)
+		log.Errorf("Invalid evidence: %v", err)
```
garuppel, line 117: @oxisto @lebogg Can you have a quick look at this. Is it ok?

service/evidence/evidence_store.go
```diff
@@ -40,11 +45,29 @@ func (s *Service) StoreEvidence(_ context.Context, req *evidence.StoreEvidenceRe
 
 	_, err = e.Validate()
 	if err != nil {
-		return nil, status.Errorf(codes.InvalidArgument, "invalid req: %v", err)
+		log.Errorf("Invalid evidence: %v", err)
```
garuppel, line 48: @oxisto @lebogg Can you have a quick look at this. Is it ok?

service/orchestrator/orchestrator.go
```diff
@@ -185,6 +194,70 @@ func (s *Service) ListMetricConfigurations(ctx context.Context, req *orchestrato
...
+		if err != nil {
+			return fmt.Errorf("error handle assessment result: %w", err)
+		}
+	}
+
+}
+
+func (s Service) handleResult(result *assessment.AssessmentResult) error {
+	_, err := result.Validate()
+	if err != nil {
```
garuppel, line 231: @oxisto @lebogg Can you have a quick look at this. Is it ok?

cmd/engine/engine.go
lebogg: I see. Probably in our case, the hook should not be called as well when we do not accept the evidence. But I can imagine that there are proper use cases.

cmd/engine/engine.go
lebogg: Then why do we need the `StoreAssessmentResult` in the hook as well? Aren't we sending the assessment result twice in this way?

I also don't know how to reference to specific LOC :)

service/orchestrator/orchestrator.go
```diff
@@ -185,6 +194,70 @@ func (s *Service) ListMetricConfigurations(ctx context.Context, req *orchestrato
...
+		log.Errorf("Invalid assessment result: %v", err)
```
lebogg, line 232: Minor thing: You could switch the lines 232, 233 and do sth like `log.Error(newError)` to avoid duplicated code?

lebogg (commented):
Besides the question about the result hook in `engine.go` the PR is good to go

api/assessment/assessment.go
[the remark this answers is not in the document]
lebogg: Strange, I thought I already commented on this: In evidences we do not have target value types, so I think it isn't checked there.

cmd/engine/engine.go
garuppel: No, the StoreAssessmentResult in the hook is from the orchestrator which is called. 

Perhaps we should have a call about that :)

lebogg: I understand it now. IMO, the assessment should send the assessment results to the orchestrator within it's code via gRPC and not directly calling the orchestrator function within the result hook. I think, we can leave it like that for now. And in the future "workflow PR" we can discuss this further.

api/assessment/assessment.go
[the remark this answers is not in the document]
lebogg: Is it resolved?

api/assessment/assessment.go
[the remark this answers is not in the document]
garuppel: I've added a new issue for that.

service/orchestrator/orchestrator.go
garuppel: I thougt the log.Errorf() should start capitalized and the returned error lowercase.

api/assessment/assessment.go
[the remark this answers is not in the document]
garuppel: Yes, I deleted this the NonCompliantComments check.

Issue #425, opened 2021-12-17
Store operator/target_value in assessment result


garuppel:
In [assessment.go](https://github.com/clouditor/clouditor/blob/09fba7d44932098dfdc87f3bfc424e483cf55690/service/assessment/assessment.go#L142) the operator is not stored in the assessment result and thus the assessment result is checked as invalid (e.g., if storing the assessment result in the orchestrator.StoreAssessmentResult())

Futhermore, the target_value must be stored.

Issue #419, opened 2021-12-15
Change code base according to MEDINA workflow 


lebogg:
TBD, e.g.
- Evidences are sent to the assessment and then from the assessment to the Orchestrator instead of the Discovery directly sending it to the orchestrator.
- Also sending assessment result from assessment to orchestrator (without hook), #407

PR #429, opened 2021-12-20
Refactor according to MEDINA workflow

lebogg:
Address #419 (adapt according to Medina Workflow), #425 (fix operator and target value in assessment result) and #464 (add RPC `ListAssessmentResult` in Orchestrator)


service/assessment/assessment.go
```diff
@@ -81,6 +123,16 @@ func (s Service) AssessEvidence(_ context.Context, req *assessment.AssessEvidenc
 
 		return res, status.Errorf(codes.InvalidArgument, "invalid req: %v", newError)
 	}
+	// Set up stream to Evidence Store.
+	// TODO(lebogg): If assessment is used as standalone service (without fwd evidences) maybe adapt code, i.e. no error when ConfigureConnections sets no evidence store
+	if s.evidenceStoreStream == nil {
+		if err = s.setEvidenceStoreStream(); err != nil {
+			return nil, fmt.Errorf("could not assess evidence: %v", err)
```
garuppel, line 130:
```suggestion
			return nil, fmt.Errorf("could not set evidence store stream: %v", err)
```

lebogg: Your proposed message is included in the underlying error message and, thus, appended to this error message.

garuppel: Ok, but at this point it is not an "assess evidence" error but an 'could not set stream'-error,  right? 

Shouldn't that be something like `status.Errorf(codes.Internal, "could not set up stream for storing evidences: %v", err)`

service/assessment/assessment.go
```diff
@@ -52,22 +56,60 @@ Service is an implementation of the Clouditor Assessment service. It should not
...
-		results: make(map[string]*assessment.AssessmentResult),
+		results:       make(map[string]*assessment.AssessmentResult),
+		Configuration: Configuration{evidenceStoreTargetAddress: "localhost:9090"},
+	}
+}
+
+// Start starts assessment by setting up connections to other services. In the future, also with manual configurations in request
+// TODO: CURRENTLY NOT USED. Maybe rename (proto) to ConfigureConnections in next PR. For now, we do it in assess evidence
+// TODO: Add CLI Command
+func (s *Service) Start(_ context.Context, _ *assessment.StartAssessmentRequest) (resp *assessment.StartAssessmentResponse, err error) {
```
garuppel, line 93: Maybe we should rename that, because it doesn't start anything but configures the connections to other services.

garuppel: Another question, do we want to do that by an endpoint or by a configuration file?

service/assessment/assessment.go
```diff
@@ -171,24 +216,145 @@ func (s Service) handleEvidence(evidence *evidence.Evidence, resourceId string)
...
-			go hook(result, err)
+			// TODO(all): We could do hook concurrent again (assuming different hooks don't interfere with each other)
+			hook(result, err)
 		}
 	}
 }
 
+// sendToEvidenceStore sends the provided evidence to the evidence store
+func (s *Service) sendToEvidenceStore(e *evidence.Evidence) {
+	// Check if evidenceStoreStream is already established (via Start method)
```
garuppel, line 344:
```suggestion
	// Check if evidenceStoreStream is already established (via Configure method)
```

service/discovery/discovery.go
```diff
@@ -218,22 +206,11 @@ func (s Service) StartDiscovery(discoverer discovery.Discoverer) {
...
-			log.Errorf("Could not send evidence to Assessment: %v", err)
-		}
-
-		err = s.EvidenceStoreStream.Send(&evidence.StoreEvidenceRequest{Evidence: e})
-		if err != nil {
-			log.Errorf("Could not send evidence to EvidenceStore: %v", err)
+		if err = s.AssessmentStream.Send(&assessment.AssessEvidenceRequest{Evidence: e}); err != nil {
+			handleError(err, "Assessment")
 		}
+		// TODO(lebogg): Remove before Merge
```
garuppel, line 212: Don't forget :)

service/discovery/discovery.go
```diff
@@ -264,3 +241,44 @@ func (s Service) Query(_ context.Context, request *discovery.QueryRequest) (resp
...
+	if status.Code(err) == codes.Internal {
+		log.Errorf("%s. Internal error on the server side: %v", prefix, err)
+	} else if status.Code(err) == codes.InvalidArgument {
+		log.Errorf("Invalid evidence - provide evidence in the right format: %v", err)
+	} else {
+		log.WithError(err)
+	}
+}
+
+/*func saveResourcesToFilesystem(result ResultOntology, filename string) error {
```
garuppel, line 257: Delete commented code

service/discovery/discovery.go
```diff
@@ -264,3 +241,44 @@ func (s Service) Query(_ context.Context, request *discovery.QueryRequest) (resp
...
+
+	err = ioutil.WriteFile(filepath+filename, exported, 0666)
+	if err != nil {
+		return fmt.Errorf("write file failed %w", err)
+	} else {
+		fmt.Println("ontology resources written to: ", filepath+filename)
+	}
+
+	return nil
+}*/
```
garuppel, line 284:
```suggestion
```

service/discovery/discovery_test.go
```diff
@@ -181,6 +183,45 @@ func TestShutdown(t *testing.T) {
...
+			name: "someOtherErr",
+			args: args{
+				err:  errors.New("some other error"),
+				dest: "Some Destination",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			handleError(tt.args.err, tt.args.dest)
```
garuppel, line 220: Isn't the test function missing here?

policies/policies.go
```diff
@@ -130,6 +130,7 @@ func RunMap(baseDir string, metric string, m map[string]interface{}) (data map[s
 	} else if data["applicable"] == false {
 		return nil, nil
 	} else {
+		fmt.Println(data)
```
oxisto, line 133: remove

policies/operators.rego
```diff
@@ -1,5 +1,9 @@
 package clouditor
 
+
+operator = data.operator
```
oxisto, line 4: Are you sure, that this is necessary? Didn't it work before?

proto/assessment.proto
```diff
@@ -39,6 +39,13 @@ option go_package = "api/assessment";
 // Representing the link between orchestrator and discovery: Assessing evidences
 // from discovery and sending results to orchestrator
 service Assessment {
+  // Configure assessment, e.g. setting up connections to other services
+  rpc Configure(ConfigureAssessmentRequest) returns (ConfigureAssessmentResponse) {
+    option (google.api.http) = {
+      post : "/v1/assessment/start"
```
oxisto, line 45: URL configure instead of start? What is the idea of this RPC anyway?

service/assessment/assessment.go
```diff
@@ -52,22 +56,47 @@ Service is an implementation of the Clouditor Assessment service. It should not
...
 	return &Service{
-		results: make(map[string]*assessment.AssessmentResult),
+		results:       make(map[string]*assessment.AssessmentResult),
+		Configuration: Configuration{evidenceStoreTargetAddress: "localhost:9090"},
 	}
 }
 
+// Configure configures assessment by setting up connections to other services. In the future, also with manual configurations in request
+// TODO: Add CLI Command
+func (*Service) Configure(_ context.Context, _ *assessment.ConfigureAssessmentRequest) (resp *assessment.ConfigureAssessmentResponse, err error) {
```
oxisto, line 92: What is the point of this (empty) function?

service/assessment/assessment.go
```diff
@@ -171,24 +216,145 @@ func (s Service) handleEvidence(evidence *evidence.Evidence, resourceId string)
...
+		// TODO(all): Currently, we send result via hook. But I think it would be cleaner to do it straight here.
+		go s.informHooks(result, nil)
 	}
 
 	return nil
 }
 
+// convertTargetValue converts the given target_value in a format accepted by protobuf
+// Note: Therefore, we have to be specific about the format of target values in the REGO code
+func convertTargetValue(value interface{}) (convertedTargetValue *structpb.Value, err error) {
```
oxisto, line 228: This function replicates behaviour of `structpb.NewValue()` why? Not necessary in my opinion.

service/assessment/assessment.go
```diff
@@ -171,24 +216,145 @@ func (s Service) handleEvidence(evidence *evidence.Evidence, resourceId string)
...
+				listOfConvertedValues = append(listOfConvertedValues, number)
+			}
+		}
+		convertedTargetValue = &structpb.Value{Kind: &structpb.Value_ListValue{ListValue: &structpb.ListValue{Values: listOfConvertedValues}}}
+		return
+	}
+	return nil, fmt.Errorf("no assertion found for %v (%T). Consider adding it", value, value)
+}
+
+func assertNumber(rawTargetValue interface{}) (*structpb.Value, error) {
```
oxisto, line 308: see above. this should be handled by the `structpb` package.

service/assessment/assessment.go
```diff
@@ -202,3 +368,19 @@ func (s Service) ListAssessmentResults(_ context.Context, _ *assessment.ListAsse
 func (s *Service) RegisterAssessmentResultHook(assessmentResultsHook func(result *assessment.AssessmentResult, err error)) {
 	s.resultHooks = append(s.resultHooks, assessmentResultsHook)
 }
+
+func (s *Service) setEvidenceStoreStream() error {
```
oxisto, line 372: should not be called `setXXX` but rather something like init.

proto/assessment.proto
oxisto: I would rather tread the "Configuration" as a resource, meaning that we will have at least a `GetConfiguration" and `UpdateConfiguration` call that maps to /v1/assessment/configuration GET/PUT

policies/operators.rego
lebogg: I should add a comment there. It is not for the operator functions, actually. They are declared so that `operator` and `target_vlaue` are added to the output of each policy (I did not want to add these two lines for each rego policy). Otherweise, the assessment can not get these two properties from the evaluation since they are simply not there.

proto/assessment.proto
lebogg: First I called it start because I wanted to init the stream to evidence store separately from `AssessEvidence`. Then we decided to do it in `AssessEvidence` which is no problem. But we now have the problem that all Services are connecting to localhost:9000 but in MEDINA we have different services (instead of one engine.go) and, therefore, different ports. The idea of Configure is to decide that in real time. 

As of now, in the medina project, @garuppel changes the external clouditor code (library) to tackle this problem (set the ports right). We wanted to change in that in the next PR.

The URL I have to change, ofc.

service/assessment/assessment.go
lebogg: Right, we will do it only once - and not multiple times

service/assessment/assessment.go
lebogg: We do it with `json` methods `Marshal` and `Unmarshal` now

lebogg:
I have removed the go:generate command in `rest.go` and add the grpc gateway parameter to each single command into `compile_proto_files.go`. The reason is that `--grpc-gateway_opt=Massessment.proto=clouditor.io/clouditor/api/assessment` was not working correctly: It created a new folder `clouditor.io/clouditor` instead of just naming the imported package accodringly in the orchestrator generated file.

We need that option since now both, Orchestrator and Assessment, have the RPC `ListAssessmentResults` which share the same request and response type. Of course, we could also use different types. But I think for now they just deliver assessment results with no further properties attached to it. In the future, Orchestrator may also add the respective tool_id to it.

cli/commands/commands.go
```diff
@@ -46,6 +47,7 @@ func AddCommands(cmd *cobra.Command) {
 		metric.NewMetricCommand(),
 		tool.NewToolCommand(),
 		assessment.NewAssessmentCommand(),
+		orchestrator.NewOrchestratorCommand(),
```
oxisto, line 50: Hmm the only issue I have with this, is that all the other commands are structured according to the resource and not the service one retrieves it from. Metric and tool information are also accessed by the orchestrator. Would it make sense to just replace the assessment list commands with those in the orchestrator? Do we need to retrieve the list of assessment results from the assessment?

openapi/orchestrator/openapi.yaml
```diff
@@ -289,6 +289,19 @@ paths:
                         application/json:
                             schema:
                                 $ref: '#/components/schemas/Metric'
+    /v1/orchestrator/results:
```
oxisto, line 292: I would suggest `/v1/orchestrator/assessment_results`

cli/commands/commands.go
lebogg: It definitely makes sense if the orchestrator is deployed as well. I always think in terms when somebody want to use one service, e.g. assessment, exclusively. But is that even realistic?

lebogg: Ok, since we won't use the CLI in deployment, I guess it doesn't matter anyway? To verify that we are on the same page: you want to keep the `cl assessment list` command but it should point to the Orchestrator method?

oxisto: > Ok, since we won't use the CLI in deployment, I guess it doesn't matter anyway? To verify that we are on the same page: you want to keep the `cl assessment list` command but it should point to the Orchestrator method?

That would be "in line" with the previous commands, yes. If this really makes sense: not sure

PR #430, opened 2021-12-20
Harmonize gRPC calls

garuppel:
Create handle-functions for gRPC calls
- orchestrator.StoreAssessmentResult(s)
- evidenceStore.StoreEvidence(s)
- assessment.AssessEvidence(s)


service/evidence/evidence_store.go
```diff
@@ -36,63 +36,66 @@ func init() {
...
-			for _, hook := range s.EvidenceHook {
-				go hook(nil, newError)
-			}
+	if err != nil {
+		resp = &evidence.StoreEvidenceResponse{
+			Status: false,
 		}
 
-		return resp, status.Errorf(codes.InvalidArgument, "invalid req: %v", err)
+		return resp, status.Errorf(codes.Internal, "Error while handling evidence: %v", err)
```
oxisto, line 48: This is probably the correct PR for this: The problem that I see here is that it is now indistinguishable whether we have an internal error (such as the DB connection was lost or so), which the user cannot fix from an invalid argument error, which the user can fix, e.g. by supplying the correct argument.

service/assessment/assessment.go
```diff
@@ -52,9 +52,9 @@ Service is an implementation of the Clouditor Assessment service. It should not
 NewService constructor should be used. It implements the AssessmentServer interface
 */
 type Service struct {
-	// ResultHook is a hook function that can be used if one wants to be
+	// resultHook is a hook function that can be used if one wants to be
 	// informed about each assessment result
-	ResultHook []func(result *assessment.AssessmentResult, err error)
+	resultHook []func(result *assessment.AssessmentResult, err error)
```
oxisto, line 57: Let's call it resultHooks. I would also suggest declaring a `ResultHookFunc` type, to make it easier for users which function signature to implement.

service/evidence/evidence_store.go
```diff
@@ -20,7 +20,7 @@ type Service struct {
 	evidences map[string]*evidence.Evidence
 
 	// Hook
-	EvidenceHook []func(result *evidence.Evidence, err error)
+	evidenceHook []func(result *evidence.Evidence, err error)
```
oxisto, line 23: evidenceHooks?

service/evidence/evidence_store.go
oxisto: So I would suggest to at least call the validate function outside of the handleXX function. All errors from validate should be invalid arguments. all errors from the handleXXX can be internal errors.

service/assessment/assessment.go
```diff
@@ -52,9 +52,9 @@ Service is an implementation of the Clouditor Assessment service. It should not
 NewService constructor should be used. It implements the AssessmentServer interface
 */
 type Service struct {
-	// ResultHook is a hook function that can be used if one wants to be
+	// resultHook is a hook function that can be used if one wants to be
```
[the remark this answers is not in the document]
lebogg: and maybe plural: `resultHooks`?

service/assessment/assessment_test.go
```diff
@@ -191,7 +191,7 @@ func TestService_AssessEvidences(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			s := Service{
-				ResultHook:                    tt.fields.ResultHook,
+				resultHook:                    tt.fields.ResultHook,
```
lebogg, line 194: Would be also plural then (`fields.resultHooks`)

service/discovery/discovery_test.go
```diff
@@ -337,7 +337,7 @@ func (mockIsCloudResource) GetCreationTime() *time.Time {
 //	var ready = make(chan bool)
 //
 //	assessmentServer := standalone.NewAssessmentServer().(*service_assessment.Service)
-//	assessmentServer.ResultHook = func(result *assessment.Result, err error) {
+//	assessmentServer.resultHook = func(result *assessment.Result, err error) {
```
lebogg, line 340: Much commented out code here. Maybe remove or stage it?

service/evidence/evidence_store.go
lebogg, line 23: Also plural?

service/assessment/assessment.go
```diff
@@ -171,17 +163,21 @@ func (s Service) handleEvidence(evidence *evidence.Evidence) error {
...
-				go hook(result, nil)
-			}
-		}
+		s.informHook(result, nil)
 	}
 
 	return nil
 }
 
+func (s Service) informHook(result *assessment.AssessmentResult, err error) {
```
lebogg, line 172: informHooks?

service/orchestrator/orchestrator.go
```diff
@@ -213,51 +213,56 @@ func (s *Service) StoreAssessmentResults(stream orchestrator.Orchestrator_StoreA
...
 
-	return nil
+	return
 }
 
 func (s *Service) RegisterAssessmentResultHook(hook func(result *assessment.AssessmentResult, err error)) {
 	s.AssessmentResultsHook = append(s.AssessmentResultsHook, hook)
 }
 
+func (s Service) informHook(result *assessment.AssessmentResult, err error) {
```
lebogg, line 257: informHooks?

service/assessment/assessment.go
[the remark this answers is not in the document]
garuppel: :+1:

service/discovery/discovery_test.go
garuppel: I didn't comment it out, but I think we do not need it any more? @oxisto @lebogg

oxisto: Yeah we can remove it

service/evidence/evidence_store.go
garuppel: @oxisto I changed it. But I think it it sufficient if the StoreEvidences() calls the StoreEvidence(). We don't need the handle method. Please, can you take a quick look: https://github.com/clouditor/clouditor/blob/8e6f24de89f83fca5b573138a8f24edb41ea5e60/service/evidence/evidence_store.go#L39

garuppel: Should we than do the same for orchestrator.StoreAssessmentResult(s)()? There is the same error regarding the status message Internal vs InvalidArgument.

garuppel:
I've changed the assessment.AssessEvidence(s)() and the orchestratore.StoreAssessmentResult(s)() as well. Now, the stream method calls the unary method.

lebogg (approved):

PR #433, opened 2021-12-21
Api validation tests

garuppel:
Add tests for evidence and assessment_result validation methods.

### DIFF
--- a/service/assessment/assessment.go
+++ b/service/assessment/assessment.go
@@ -100,6 +100,7 @@ func (s Service) AssessEvidences(stream assessment.Assessment_AssessEvidencesSer
 				log.Infof("Stopped receiving streamed evidences")
 				return stream.SendAndClose(&emptypb.Empty{})
 			}
+
 			return err
 		}
 
--- a/service/assessment/assessment.go
+++ b/service/assessment/assessment.go
@@ -52,9 +52,9 @@ Service is an implementation of the Clouditor Assessment service. It should not 
 NewService constructor should be used. It implements the AssessmentServer interface
 */
 type Service struct {
-	// resultHook is a hook function that can be used if one wants to be
+	// resultHooks is a hook function that can be used if one wants to be
 	// informed about each assessment result
-	resultHook []func(result *assessment.AssessmentResult, err error)
+	resultHooks []func(result *assessment.AssessmentResult, err error)
 
 	results map[string]*assessment.AssessmentResult
 	assessment.UnimplementedAssessmentServer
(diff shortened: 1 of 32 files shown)


service/orchestrator/orchestrator.go
```diff
@@ -213,51 +213,56 @@ func (s *Service) StoreAssessmentResults(stream orchestrator.Orchestrator_StoreA
...
+				return stream.SendAndClose(&emptypb.Empty{})
+			}
+
+			return err
 		}
 
 		err = s.handleResult(result)
 		if err != nil {
-			return fmt.Errorf("error handle assessment result: %w", err)
+			return status.Errorf(codes.Internal, "Error while handling assessment result: %v", err)
```
oxisto, line 229: I think error messages should begin with lowercase. It is definitely the case for `fmt.Errorf`, not 100% sure about `status.Errorf`

service/assessment/assessment.go
```diff
@@ -52,9 +52,9 @@ Service is an implementation of the Clouditor Assessment service. It should not
 NewService constructor should be used. It implements the AssessmentServer interface
 */
 type Service struct {
-	// ResultHook is a hook function that can be used if one wants to be
+	// resultHook is a hook function that can be used if one wants to be
 	// informed about each assessment result
-	ResultHook []func(result *assessment.AssessmentResult, err error)
+	resultHook []func(result *assessment.AssessmentResult, err error)
```
oxisto, line 57: resultHooks? Since its more than one?

service/evidence/evidence_store.go
```diff
@@ -36,63 +36,66 @@ func init() {
...
-			for _, hook := range s.EvidenceHook {
-				go hook(nil, newError)
-			}
+	if err != nil {
+		resp = &evidence.StoreEvidenceResponse{
+			Status: false,
 		}
 
-		return resp, status.Errorf(codes.InvalidArgument, "invalid req: %v", err)
+		return resp, status.Errorf(codes.Internal, "Error while handling evidence: %v", err)
```
oxisto, line 48: The problem that I see here is that it is now indistinguishable whether we have an internal error (such as the DB connection was lost or so), which the user cannot fix from an invalid argument error, which the user can fix, e.g. by supplying the correct argument.

garuppel: That's a good point.

service/orchestrator/orchestrator.go
garuppel: For me it's ok, if every error starts with lowercase.

garuppel: @oxisto but the log.Errorf start with uppercase?!?

oxisto: Lowercase. See https://github.com/golang/go/wiki/CodeReviewComments#error-strings

oxisto: > @oxisto but the log.Errorf start with uppercase?!?

Yes. See https://github.com/golang/go/wiki/CodeReviewComments#error-strings. The idea is that the error message should be lowercase because it is often logged as "The error is: %v" in a log function (such as log.Errorf). Log messages can/should start with uppercase.

service/evidence/evidence_store.go
garuppel: Done in PR #430.

Issue #404, opened 2021-12-06
Add CLI commands for evidence store


lebogg:
**Problem**
We cannot fetch the information of which evidences are currently stored in the evidence store via the CLI. The problem arised when an external tool tried to send evidences to the orchestrator for assessing them. Nothing happend though, and via CLI we couldn't find the place where to process actually went wrong.

**Solution**
CLI commands for doing so.

**Alternative**
An alternative would be to get info logs directly in the terminal when an evidence is stored or, one step earlier, retrieved of a discovery tool. To avoid bloated terminal output, implement it as `log.debug`?

PR #465, opened 2022-01-25
Update CLI: Change structure and add commands 

lebogg:
- [x] CLI commands should be divided into resources (e.g. `cl evidence list`) and services (e.g. `cl service discovery start`)
- [x] #404:  (see first example above)
- [x] `cl assessment list` should call Orchestrator's method, not of the Assessment. The Assessment's corresponding method should be executable via `cl service assessment list`


lebogg:
Now, discovery moved to service CMDs. Should we also add `resource` command for Discovery Query (`cl resource list` == `discovery.query(...)`)?

cli/commands/assessment/assessment.go
```diff
@@ -26,50 +26,23 @@
...
+	// Change use for better readability (cl assessment_result list vs cl assessment_result list_assessment_results)
+	cmd.Use = "list"
 	return cmd
 }
 
 // NewAssessmentCommand returns a cobra command for `assessment` subcommands
 func NewAssessmentCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "assessment",
+		Use:   "assessment_result",
```
oxisto, line 45: I think in the commands we should stick to dashes instead of underscores for a little bit of better readability.

cli/commands/commands.go
```diff
@@ -30,10 +30,10 @@ import (
 	"clouditor.io/clouditor/cli/commands/assessment"
 	"clouditor.io/clouditor/cli/commands/cloud"
 	"clouditor.io/clouditor/cli/commands/completion"
-	"clouditor.io/clouditor/cli/commands/discovery"
+	evidences "clouditor.io/clouditor/cli/commands/evidence"
```
oxisto, line 33: Why is this imported as evidences?

cli/commands/service/discovery/discovery.go
```diff
@@ -97,11 +97,11 @@ func NewQueryDiscoveryCommand() *cobra.Command {
 	return cmd
 }
 
-// NewStartDiscoveryCommand returns a cobra command for `discovery` subcommands
+// NewDiscoveryCommand returns a cobra command for `discovery` subcommands
 func NewDiscoveryCommand() *cobra.Command {
```
oxisto, line 101: Should we have another set of commands for discovery results / resources?

cli/commands/service/orchestrator/orchestrator.go
```diff
@@ -35,11 +35,11 @@ import (
...
 )
 
-// NewListResultsCommand returns a cobra command for the `list` subcommand
-func NewListResultsCommand() *cobra.Command {
+// NewListAssessmentResultsCommand returns a cobra command for the `list` subcommand
+func NewListAssessmentResultsCommand() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "list",
-		Short: "Lists all assessment results",
+		Use:   "list_assessment_results",
```
oxisto, line 41:
```suggestion
		Use:   "list-assessment-results",
```
```suggestion
		Use:   "list_assessment_results",
```

oxisto (commented):
Some minor inconsistencies, but overall I think its a good approach.

oxisto:
> Now, discovery moved to service CMDs. Should we also add `resource` command for Discovery Query (`cl resource list` == `discovery.query(...)`)?

Yes!

cli/commands/login/login.go
```diff
@@ -37,6 +37,7 @@ import (
 
 const (
 	// URLFlag is the viper flag for the server url
+	// TODO(oxisto): Unused constant?
 	URLFlag = "url"
```
lebogg, line 41: Do we need that constant @oxisto ? It seems to be unused.

cli/commands/commands.go
```diff
@@ -42,14 +43,17 @@ import (
 // AddCommands adds all subcommands
 func AddCommands(cmd *cobra.Command) {
 	cmd.AddCommand(
+		// commands for resources
 		login.NewLoginCommand(),
```
lebogg, line 47: Seems that the package name always is included in the function names as well. In general, such repetition should be avoided according to this [go dev blog](https://go.dev/blog/package-names) (at least for types).
But in that case, I guess it is ok, since 1st) just `New` or `NewCommand` is not meaningful enough either and 2nd) replace with a generic name like `NewParentCommand` is also not that meaningful.

cli/commands/login/login.go
oxisto: Keep it for now. We need to make all of those URLs configurable at some point.

cli/commands/commands.go
oxisto: In this case, this seems to be somewhat idiomatic to the cobra/viper package and as far as I remember they structure the commands according to this schema in their examples, so I would stick with this.

Issue #463, opened 2022-01-21
Validate or set assessment ID in Orchestrator


lebogg:
**Describe the bug**
In the RPC `StoreAssessmentResult` in the orchestrator, we [validate the assessment result (Line 202)](https://github.com/clouditor/clouditor/blob/e481a8b99d1584531f5919554377a1e3ddea27ab/service/orchestrator/orchestrator.go#L202) and then [store it into the map with its ID(L213)](https://github.com/clouditor/clouditor/blob/e481a8b99d1584531f5919554377a1e3ddea27ab/service/orchestrator/orchestrator.go#L213).
But the function `validate` doesn't check the ID. I.e., an assessment result with empty ID is stored with empty key "" and will be overwritten by future assessment results with empy IDs.

**Expected behavior**
Either `validate` checks if ID is set (meaning that results with empty IDs are rejected) or the Orchestrator creates one itself (e.g. UUID) for assessment with no IDs.

oxisto:
As discussed, the assessment tool should set a UUID and the orchestrator should validate this.

lebogg:
Yes, will be addressed in the next PR.

PR #470, opened 2022-01-27
Add validation for assessment result ID

lebogg:
- #463

### DIFF
--- a/api/assessment/assessment.go
+++ b/api/assessment/assessment.go
@@ -30,17 +30,21 @@ import "errors"
 type ResultHookFunc func(result *AssessmentResult, err error)
 
 var (
+	ErrIdMissing                             = errors.New("id in assessment result is missing")
 	ErrTimestampMissing                      = errors.New("timestamp in assessment result is missing")
 	ErrMetricIdMissing                       = errors.New("metric id in assessment result is missing")
 	ErrMetricConfigurationMissing            = errors.New("metric configuration in assessment result is missing")
 	ErrEvidenceIdMissing                     = errors.New("evidence id in assessment result is missing")
-	ErrNonComplianceCommentsMissing          = errors.New("non-compliance comments in assessment result is missing")
 	ErrMetricConfigurationOperatorMissing    = errors.New("operator in metric data is missing")
 	ErrMetricConfigurationTargetValueMissing = errors.New("target value in metric data is missing")
 )
 
 // Validate validates the assessment result according to several required fields
 func (result *AssessmentResult) Validate() (resourceId string, err error) {
+	if result.Id == "" {
+		return "", ErrIdMissing
+	}
+
 	if result.Timestamp == nil {
 		return "", ErrTimestampMissing
 	}
--- a/api/assessment/assessment.go
+++ b/api/assessment/assessment.go
@@ -46,7 +46,7 @@ var (
 // Validate validates the assessment result according to several required fields
 func (result *AssessmentResult) Validate() (resourceId string, err error) {
 	if _, err = uuid.Parse(result.Id); err != nil {
-		return "", fmt.Errorf("%v: %v", ErrInvalidFormat, err)
+		return "", fmt.Errorf("%v: %w", ErrInvalidFormat, err)
 	}
 
 	if result.Timestamp == nil {
--- a/api/assessment/assessment.go
+++ b/api/assessment/assessment.go
@@ -25,12 +25,15 @@
 
 package assessment
 
-import "errors"
+import (
+	"errors"
+	"fmt"
+	"github.com/google/uuid"
+)
 
 type ResultHookFunc func(result *AssessmentResult, err error)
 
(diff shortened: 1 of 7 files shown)


api/assessment/assessment.go
```diff
@@ -30,17 +30,21 @@ import "errors"
...
 	ErrMetricConfigurationMissing            = errors.New("metric configuration in assessment result is missing")
 	ErrEvidenceIdMissing                     = errors.New("evidence id in assessment result is missing")
-	ErrNonComplianceCommentsMissing          = errors.New("non-compliance comments in assessment result is missing")
 	ErrMetricConfigurationOperatorMissing    = errors.New("operator in metric data is missing")
 	ErrMetricConfigurationTargetValueMissing = errors.New("target value in metric data is missing")
 )
 
 // Validate validates the assessment result according to several required fields
 func (result *AssessmentResult) Validate() (resourceId string, err error) {
+	if result.Id == "" {
```
oxisto, line 44: Should we also check if the id is a UUID? Or do you think this is too performance-heavy? Also potentially for the evidence ID.

garuppel: Do we already know in which format the ID should be? If yes, then I would check that.

lebogg: Good point. We could use the error of `uuid.parse(result.Id)` or do a sanity check by just compare the length (36) via `len(result.Id)`. I did a small benchmark test for the 3 options:

0.0000004 ns/op for simple length check
0.0000009 ns/op for empty string check
0.0000050 ns/op for uuid.parse

Since length check is even faster (or at least somehow equally fast) than empty string check (does this makes sense, anyway :D), it wouldn't add significant time. The UUID approach is also not that slow but needs a dependency.

garuppel (commented):
Shouldn't we check the resourceID as well?

api/assessment/assessment.go
oxisto: We already have GitHub.com/google/uuid as a dependency

lebogg: Then I would just substitute the empty string check with uuid.Parse() way since it also throws error for empty string ("invalid UUID length: 0")

api/assessment/assessment.go
```diff
@@ -25,22 +25,31 @@
...
 	ErrEvidenceIdMissing                     = errors.New("evidence id in assessment result is missing")
-	ErrNonComplianceCommentsMissing          = errors.New("non-compliance comments in assessment result is missing")
 	ErrMetricConfigurationOperatorMissing    = errors.New("operator in metric data is missing")
 	ErrMetricConfigurationTargetValueMissing = errors.New("target value in metric data is missing")
 )
 
 // Validate validates the assessment result according to several required fields
 func (result *AssessmentResult) Validate() (resourceId string, err error) {
+	if _, err = uuid.Parse(result.Id); err != nil {
+		fmt.Println(uuid.IsInvalidLengthError(err))
```
oxisto, line 49: Remove unnecessary print

api/assessment/assessment.go
```diff
@@ -25,22 +25,31 @@
...
+import (
+	"errors"
+	"fmt"
+	"github.com/google/uuid"
+)
 
 type ResultHookFunc func(result *AssessmentResult, err error)
 
 var (
+	ErrAssessmentIdConfiguredIncorrectly     = errors.New("assessment result id is configured incorrectly")
```
oxisto, line 37: Maybe `ErrIdInvalidFormat` ? Configured correctly sounds misleading. It should say that the Id is not in an expected format (UUID)

api/assessment/assessment.go
```diff
@@ -25,22 +25,30 @@
...
 	ErrEvidenceIdMissing                     = errors.New("evidence id in assessment result is missing")
-	ErrNonComplianceCommentsMissing          = errors.New("non-compliance comments in assessment result is missing")
 	ErrMetricConfigurationOperatorMissing    = errors.New("operator in metric data is missing")
 	ErrMetricConfigurationTargetValueMissing = errors.New("target value in metric data is missing")
 )
 
 // Validate validates the assessment result according to several required fields
 func (result *AssessmentResult) Validate() (resourceId string, err error) {
+	if _, err = uuid.Parse(result.Id); err != nil {
+		return "", fmt.Errorf("%v: %v", ErrInvalidFormat, err)
```
oxisto, line 49:
```suggestion
		return "", fmt.Errorf("%v: %w", ErrInvalidFormat, err)
```

oxisto (approved):

oxisto:
> Shouldn't we check the resourceID as well?

Yes, I included this as well. The resource ID does not follow any convention (yet). So I am just checking for pure existence. 

I also removed the wrapping of the UUID error codes, since my gut feeling says that this is not a good idea from a security point of view. I would keep the validation errors constant without any means to reflect values back to somehow who potentially inserts weird values.

Issue #467, opened 2022-01-27
Separate database access from service packages using an interface


oxisto:
**Is your feature request related to a problem? Please describe.**

Currently, we are directly using the `gorm` structs in our service packages. This creates problems for external users that might depend on another solution for persistence or just would like to have the choice.

**Describe the solution you'd like**

We already have a `persistence` package, we should introduce some more interfaces to interact with the services packages and then have sub packages, e.g. for gore.

lebogg:
We could also start with using the in-memory DB for our services consequently instead of using maps for storing, e.g., assessment results and evidences.

PR #472, opened 2022-01-28
Separate DB access with interface

lebogg:
- [x] #467


lebogg:
Some comments (see also TODOs in code):
- For RPC, should we check if reuest is nil? Can it even be nil?
- DeepSource Bug about coupling: Should we do 2 separate methods for init DBs or even 2 interface implementations for gorm?
- Add DB initiatlization for each service in the respective `NewService` as part of `opts`?
- Adding tests to auth?
- In `UpdateXXX` RPCs, should we add ID as separate field as in `UpdateCloudService`. Is this some convention? For the REST case we can add it via option anyway - I would omit it.
- I added validation methods for requests if needed in the respective API file: Maybe we can introduce this as a convention within our project for RPCs?

api/orchestrator/orchestrator.go
```diff
@@ -0,0 +1,98 @@
+// Copyright 2022 Fraunhofer AISEC
```
oxisto, line 1: As discussed, I would move this to a) a separate PR and b) wait for Go 1.18 generics

cli/commands/cloud/cloud_test.go
```diff
@@ -65,23 +69,19 @@ func TestMain(m *testing.M) {
...
-	orchestrator.RegisterOrchestratorServer(server, s)
+	orchestratorService = service_orchestrator.NewService(gormX)
+	authService = service_auth.NewService(gormX)
 
-	target, err = s.CreateDefaultTargetCloudService()
+	sock, server, _, err = authService.StartDedicatedAuthServer(":0", service_auth.WithApiKeySaveOnCreate(false))
 	if err != nil {
 		panic(err)
 	}
+	orchestrator.RegisterOrchestratorServer(server, orchestratorService)
```
oxisto, line 84: I am not sure if this works. I am pretty sure that you cannot register additional services to a gRPC server once it has been started (which happens in authService.StartDedicatedAuthServer)

cli/commands/evidence/evidence_test.go
```diff
@@ -52,34 +52,40 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-var sock net.Listener
-var server *grpc.Server
+var (
+	sock                 net.Listener
+	server               *grpc.Server
+	evidenceStoreService *service_evidenceStore.Service
```
oxisto, line 58: is it needed to store those services?

cli/commands/login/login_test.go
```diff
@@ -41,18 +41,27 @@ import (
 	"google.golang.org/grpc"
 )
 
-var sock net.Listener
-var server *grpc.Server
+var (
+	sock        net.Listener
+	server      *grpc.Server
+	authService *service_auth.Service
```
oxisto, line 47: Is it needed to store it?

cli/commands/login/login_test.go
```diff
@@ -41,18 +41,27 @@ import (
...
+)
 
 func TestMain(m *testing.M) {
-	var err error
+	var (
+		err error
 
-	err = persistence.InitDB(true, "", 0)
+		gormX = new(persistence.GormX)
```
oxisto, line 54: In general, seems a very weird way to initialise it.

cli/commands/service/discovery/discovery_test.go
```diff
@@ -51,32 +51,34 @@ import (
 
 var sock net.Listener
 var server *grpc.Server
+var authServices *service_auth.Service
```
oxisto, line 54: all other tests are using var multiline

cmd/engine/engine.go
```diff
@@ -219,7 +221,10 @@ func doCmd(_ *cobra.Command, _ []string) (err error) {
 	// evidenceStoreService.RegisterEvidenceHook(func(result *evidence.Evidence, err error) {})
 	// assessmentService.RegisterAssessmentResultHook(func(result *assessment.AssessmentResult, err error) {}
 
-	authService.CreateDefaultUser(viper.GetString(APIDefaultUserFlag), viper.GetString(APIDefaultPasswordFlag))
+	err = authService.CreateDefaultUser(viper.GetString(APIDefaultUserFlag), viper.GetString(APIDefaultPasswordFlag))
+	if err != nil {
+		log.Errorf("could not create default user: %v", err)
```
oxisto, line 226:
```suggestion
		log.Errorf("Could not create default user: %v", err)
```

persistence/db.go
```diff
@@ -26,50 +26,108 @@
...
 	"gorm.io/driver/postgres"
 	"gorm.io/driver/sqlite"
 	"gorm.io/gorm"
 )
 
 var log *logrus.Entry
-var db *gorm.DB
+
+// TODO(lebogg): Lack for better term
+type GormX struct {
```
oxisto, line 41: This should probably be in a separate folder? so that we have a package `persistance/gorm` and `persistance/somethingelse`. At *least* it should be in a separate file.

oxisto: And the Name is weird :)

persistence/db.go
```diff
@@ -26,50 +26,108 @@
...
 
 	return nil
 }
 
+func (g *GormX) Create(r interface{}) error {
+	return g.db.Create(r).Error
+}
+
+func (g *GormX) Read(r interface{}, conds ...interface{}) (err error) {
+	switch r.(type) {
```
oxisto, line 79: Why is that type switch needed? This is not extensible at all.

persistence/db.go
```diff
@@ -26,50 +26,108 @@
...
+		return g.db.Session(&gorm.Session{AllowGlobalUpdate: true}).Delete(r).Error
+	}
+	// Remove record r with given ID
+	tx := g.db.Delete(r, "Id = ?", id)
+	if err := tx.Error; err != nil {
+		return err
+	}
+	// No record with given ID found
+	if tx.RowsAffected == 0 {
+		return gorm.ErrRecordNotFound
```
oxisto, line 113: We should return an error from the perstiance package, not from gorm

persistence/db.go
```diff
@@ -26,50 +26,108 @@
...
+
 // GetDatabase returns the database
-func GetDatabase() *gorm.DB {
-	return db
+func (g *GormX) GetDatabase() *gorm.DB {
+	return g.db
+}
+
+// Reset resets entire the database
+func (g *GormX) Reset() (err error) {
```
oxisto, line 125: Why?

persistence/db.go
```diff
@@ -26,50 +26,108 @@
...
+type GormX struct {
+	db *gorm.DB
+}
 
 func init() {
 	log = logrus.WithField("component", "db")
 }
 
-func InitDB(inMemory bool, host string, port int16) (err error) {
+func (g *GormX) Init(inMemory bool, host string, port int16) (err error) {
```
oxisto, line 49: Can we change this to a `NewXXX` style builder? Similar to what we have in the services? 

`var db = NewGORM(WithInMemory())`

`var db = NewGORM(WithPostgres("localhost", "user", "password))`

In both cases ´db` should then be of the interface type `Database`

persistence/persistence.go
```diff
@@ -0,0 +1,43 @@
...
+// $$ |      $$ |$$ |  $$ |$$ |  $$ |$$ |  $$ |$$ |  $$ |$$\ $$ |  $$ |$$ |
+// \$$$$$$\  $$ |\$$$$$   |\$$$$$   |\$$$$$$  |$$ |  \$$$   |\$$$$$   |$$ |
+//  \_______|\__| \______/  \______/  \_______|\__|   \____/  \______/ \__|
+//
+// This file is part of Clouditor Community Edition.
+
+package persistence
+
+// IsDatabase comprises a database interface
+// TODO(all): CRUD Pattern good enough?
```
oxisto, line 29: Just call it `Database` or `Storage`

persistence/persistence.go
```diff
@@ -0,0 +1,43 @@
...
+type IsDatabase interface {
+	// Init initializes the database
+	Init(inMemory bool, host string, port int16) error
```
oxisto, line 32: Probably not needed with the `New` builder pattern.

proto/orchestrator.proto
```diff
@@ -228,6 +228,7 @@ message GetCloudServiceRequest { string service_id = 1; }
 
 message RegisterCloudServiceRequest { CloudService service = 1; }
 
+// TODO(oxisto): Why separate service_id when it is in service already?
```
oxisto, line 231: Because Cloud service is only embedded in the request. the actual id belonging to the request is in service_id. This is also used in the HTTP path. Ref. to style guide ;)

service/auth/auth.go
```diff
@@ -121,7 +128,7 @@ func WithApiKeySaveOnCreate(saveOnCreate bool) ServiceOption {
 }
 
 // NewService creates a new Service representing an authentication service.
-func NewService(opts ...ServiceOption) *Service {
+func NewService(db persistence.IsDatabase, opts ...ServiceOption) *Service {
```
oxisto, line 131: Does it make sense to rather have a service option WithPersistance() and maybe default to an in-memory implementation of it?

service/auth/auth.go
```diff
@@ -341,25 +347,33 @@ func hashPassword(password string) (string, error) {
...
-	db.Model(&auth.User{}).Count(&count)
+// TODO(all): return created user?
+func (s *Service) CreateDefaultUser(username string, password string) error {
+	var (
+		storedUsers []*auth.User
+		err         error
+	)
 
-	if count == 0 {
+	err = s.db.Read(&storedUsers)
```
oxisto, line 357: Can we have a `Count()` function in the database interface please :) This seems very complicated and it is not necessary to fetch all users.

service/auth/auth.go
```diff
@@ -380,36 +394,66 @@ func (s Service) issueToken(subject string, fullName string, email string, expir
 	return
 }
 
+// StartDedicatedAuthServer starts a gRPC server containing just the auth service
+func (s *Service) StartDedicatedAuthServer(address string) (sock net.Listener, server *grpc.Server, err error) {
```
oxisto, line 398:
```suggestion
func (s *Service) StartDedicatedServer(address string) (sock net.Listener, server *grpc.Server, err error) {
```

service/orchestrator/cloud_service.go
```diff
@@ -91,40 +92,37 @@ func (s *Service) GetCloudService(_ context.Context, req *orchestrator.GetCloudS
...
-	if req.ServiceId == "" {
-		return nil, status.Errorf(codes.InvalidArgument, "service id is empty")
-	}
+	// Check if cloud service is in the DB
+	_, err = s.GetCloudService(context.Background(), &orchestrator.GetCloudServiceRequest{ServiceId: req.ServiceId})
 
-	var count int64
-	err = s.db.Model(&orchestrator.CloudService{}).Count(&count).Error
+	// Update cloud service
+	// TODO(lebogg): See questions above: Copying serviceId to the struct makes it unnecessary in the first place IMO
```
oxisto, line 106: see explanations above

service/orchestrator/cloud_service.go
```diff
@@ -91,40 +92,37 @@ func (s *Service) GetCloudService(_ context.Context, req *orchestrator.GetCloudS
...
-		return nil, status.Errorf(codes.InvalidArgument, "service is empty")
+	if err = req.Validate(); err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, err.Error())
 	}
 
-	if req.ServiceId == "" {
-		return nil, status.Errorf(codes.InvalidArgument, "service id is empty")
-	}
+	// Check if cloud service is in the DB
+	_, err = s.GetCloudService(context.Background(), &orchestrator.GetCloudServiceRequest{ServiceId: req.ServiceId})
```
oxisto, line 103: Not really necessary to call the complete getCloudservice function here. A simple Count() would be enough.

service/orchestrator/cloud_service.go
```diff
@@ -134,32 +132,32 @@ func (s *Service) RemoveCloudService(_ context.Context, req *orchestrator.Remove
...
-
-	if count == 0 {
-		// Create a default target cloud service
-		service =
+// CreateDefaultTargetCloudService implements method for OrchestratorServer interface for creating a new "default"
+// target cloud services, if no target service exists in the database.
+// Returns new created target cloud service. Otherwise, returns nil (if a record exists) or error if sth went wrong
+func (s *Service) CreateDefaultTargetCloudService() (*orchestrator.CloudService, error) {
+	var storedServices []*orchestrator.CloudService
+	err := s.db.Read(&storedServices)
```
oxisto, line 140: Same as above, please provide a Count interface in the persistence layer. this is unnecessarily resource exhaustive.

service/orchestrator/cloud_service.go
```diff
@@ -134,32 +132,32 @@ func (s *Service) RemoveCloudService(_ context.Context, req *orchestrator.Remove
...
 				Description: DefaultTargetCloudServiceDescription,
 			}
-
 		// Save it directly into the database, so that we can set the ID
-		err = s.db.Create(&service).Error
-
+		err = s.db.Create(&service)
 		if err != nil {
-			log.Infof("Created new default target cloud service %s", service.Id)
+			return nil, status.Errorf(codes.Internal, "error while writing new target cloud service to db: %v", err)
```
oxisto, line 155: This is not a gRPC call!

service/orchestrator/cloud_service.go
```diff
@@ -134,32 +132,32 @@ func (s *Service) RemoveCloudService(_ context.Context, req *orchestrator.Remove
...
 				Name:        DefaultTargetCloudServiceName,
 				Description: DefaultTargetCloudServiceDescription,
 			}
-
 		// Save it directly into the database, so that we can set the ID
-		err = s.db.Create(&service).Error
-
+		err = s.db.Create(&service)
 		if err != nil {
-			log.Infof("Created new default target cloud service %s", service.Id)
```
oxisto, line 158: Please keep some logging :)

service/orchestrator/orchestrator.go
```diff
@@ -89,11 +88,14 @@ func WithMetricsFile(file string) ServiceOption {
 	}
 }
 
-func NewService(opts ...ServiceOption) *Service {
+// NewService creates a new Orchestrator service
+// TODO(lebogg to oxisto): We could also provide dv via opts?
```
oxisto, line 92: yes, would be a better option in my opinion

cli/commands/cloud/cloud.go
```diff
@@ -197,6 +197,8 @@ func NewUpdateCloudServiceCommand() *cobra.Command {
 			res, err = client.UpdateCloudService(context.Background(), &orchestrator.UpdateCloudServiceRequest{
 				ServiceId: viper.GetString("id"),
 				Service: &orchestrator.CloudService{
+					// TODO(lebogg): Remove again if we decide to keep service ID in separate field above
+					Id:          viper.GetString("id"),
```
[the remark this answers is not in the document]
lebogg: Yea, maybe it should not part of this PR but it didn't and still doesn't feel right to use the ID in the way you say. Thus, you have to put the same ID twice (in REQ and in server object). Otherwise, maybe the DB overrides even the ID in the DB which should not be possible. Of course, we could mitigate this by adding a validation check that IDs should be the same or by implementing the DB Update operation accordingly (but see also comment in code with the differences of `gorm.Update()` vs `gorm.Save()`)

See also my comment above: "In UpdateXXX RPCs, should we add ID as separate field as in UpdateCloudService. Is this some convention? For the REST case we can add it via option anyway - I would omit it."

cli/commands/cloud/cloud_test.go
```diff
@@ -49,10 +49,14 @@ import (
...
-var server *grpc.Server
-var s *service_orchestrator.Service
-var target *orchestrator.CloudService
+var (
+	sock                net.Listener
+	server              *grpc.Server
+	authService         *service_auth.Service
+	orchestratorService *service_orchestrator.Service
+	target              *orchestrator.CloudService
+	gormX               = new(persistence.GormX)
```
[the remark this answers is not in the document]
lebogg: Yes, but for testing we use our implementation anyway. But yes, I could, of course, making the gormX of type `persistence.Database` and then later instantiate it as a gorm DB

cli/commands/cloud/cloud.go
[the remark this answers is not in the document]
oxisto: You only need to put the ID in the request, not in the embedded service object. The reasoning behind this is that basically the update requests contains a list of key/values what to change in the original server object. Now instead of supply a extra message that contains all the fields as the original service object (with exception) of the id, it is convention to just use the service object itself and keep the ID empty.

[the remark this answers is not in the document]
oxisto: See our referenced style guide for examples https://cloud.google.com/apis/design/standard_methods#create.

Annoyingly, they are using the extra ID field for create, but omit it for update. However, they are using patch instead of put for updates.

cli/commands/cloud/cloud_test.go
lebogg: But this was originally introduced by you :D And the tests don't fail.
I will look at it more closely.

cli/commands/evidence/evidence_test.go
lebogg: Why do you think they are stored?

cli/commands/cloud/cloud_test.go
oxisto: I know. The problem is that this seems to creep up on some tests now. Seems this introduces a weird race condition. I have proposed a solution in the #486 branch. Basically adding a function argument to `StartDedicatedAuthServer` that allows the start of additional services.

persistence/db.go
lebogg: For testing purposes to reset the DB after each s.t. the tests don't influence each other. I removed it again and added single reset function in test files.

cli/commands/evidence/evidence_test.go
oxisto: I meant stored in the file as a global variable. Before that they were only local or discarded (as `_`)

lebogg: sock and server were also stored as global variables - and don't have to. And sometimes also single services. I wanted to make it look consistent. But I can revert it back (and make other variables, e.g. sock and server, local as well if they can be local variables?)

lebogg:
TODOs:
- Improve test coverage
- Look at auth testing

cli/commands/cloud/cloud_test.go
```diff
@@ -229,3 +242,44 @@ func TestGetMetricConfiguration(t *testing.T) {
 
 	assert.Nil(t, err)
 }
+
+// startServer starts server with orchestrator and dedicated auth server. We don't do it in TestMain since you
```
oxisto, line 246:
```suggestion
// startServer starts a gRPC server with an orchestrator and auth service. We don't do it in TestMain since you
```

cli/commands/cloud/cloud_test.go
```diff
@@ -229,3 +242,44 @@ func TestGetMetricConfiguration(t *testing.T) {
...
+	orchestratorService = service_orchestrator.NewService()
+	authService = service_auth.NewService(service_auth.WithApiKeySaveOnCreate(false))
+
+	sock, server, err = authService.StartDedicatedServer(":0")
+	if err != nil {
+		panic(err)
+	}
+	orchestrator.RegisterOrchestratorServer(server, orchestratorService)
+
+	//defer sock.Close()
```
oxisto, line 266: Remove

cli/session_test.go
```diff
@@ -122,6 +126,8 @@ func TestSession(t *testing.T) {
 
 	s, ok := status.FromError(err)
 
+	// TODO(lebogg): Continue here
+	fmt.Println(err)
```
oxisto, line 130: ?

service/auth/auth.go
```diff
@@ -87,6 +92,8 @@ type Service struct {
 	}
 
 	apiKey *ecdsa.PrivateKey
+
+	storage persistence.Storage
```
oxisto, line 96: In other services, we called it `db`

service/auth/auth.go
```diff
@@ -341,25 +359,33 @@ func hashPassword(password string) (string, error) {
 }
 
 // CreateDefaultUser creates a default user in the database
-func (Service) CreateDefaultUser(username string, password string) {
-	db := persistence.GetDatabase()
-
-	var count int64
-	db.Model(&auth.User{}).Count(&count)
+// TODO(all): return created user?
```
oxisto, line 362: Yes, we could. There is not really a benefit to it though.

service/auth_test.go
```diff
@@ -25,201 +25,186 @@
 
 package service
 
-import (
```
oxisto, line 28: Please don't move it as part of this PR. I am any way working on resolving left over issues from the JWT/JWK. Please focus on the database path :)

lebogg: It is/was database related because the idea was to attach DB(s) to services. But I try to bring it back, yes. I let you know

lebogg: It's back. Wasn't that hard

lebogg:
Still some work (mainly testing) to do. In particular, testing the functionality of Update/Save of GORM.

service/orchestrator/cloud_service.go
```diff
@@ -58,31 +60,39 @@ func (s *Service) RegisterCloudService(_ context.Context, req *orchestrator.Regi
 	service.Description = req.Service.Description
 
 	// Persist the service in our database
-	s.db.Create(&service)
+	err = s.storage.Create(&service)
```
oxisto, line 63: This creates an unnecessary pointer pointer, since service is already `*CloudService`

service/orchestrator/cloud_service.go
```diff
@@ -58,31 +60,39 @@ func (s *Service) RegisterCloudService(_ context.Context, req *orchestrator.Regi
...
+		return nil, status.Errorf(codes.InvalidArgument, orchestrator.ErrRequestIsNil.Error())
+	}
+	if req.ServiceId == "" {
+		return nil, status.Errorf(codes.InvalidArgument, orchestrator.ErrIDIsMissing.Error())
 	}
 
 	response = new(orchestrator.CloudService)
-	err = s.db.First(&response, "Id = ?", req.ServiceId).Error
-	if errors.Is(err, gorm.ErrRecordNotFound) {
+	err = s.storage.Get(&response, "Id = ?", req.ServiceId)
```
oxisto, line 94: This creates an unnecessary pointer pointer, since service is already `*CloudService`

oxisto (approved):

service/auth/auth_test.go
```diff
@@ -271,3 +273,222 @@ func TestService_loadApiKey(t *testing.T) {
...
+				service, ok := i1.(*Service)
+				if !assert.True(tt, ok) {
+					return false
+				}
+
+				return assert.Equal(tt, "SomePath", service.config.keyPath)
+			},
+		},
+		{
+			name: "New service with Api Key Sve on Create",
```
oxisto, line 318: minor nit: Save

service/auth/auth_test.go
```diff
@@ -271,3 +273,222 @@ func TestService_loadApiKey(t *testing.T) {
...
+				service, ok := i1.(*Service)
+				if !assert.True(tt, ok) {
+					return false
+				}
+
+				return assert.Equal(tt, true, service.config.keySaveOnCreate)
+			},
+		},
+		{
+			name: "New service with Api Key Sve on Create",
```
oxisto, line 332: minor nit: Save

oxisto (approved):

PR #477, opened 2022-02-02
Add support for CORS in the REST API

oxisto:
This PR adds support for CORS in the REST API

### DIFF
--- a/cmd/engine/engine.go
+++ b/cmd/engine/engine.go
@@ -61,18 +61,21 @@ import (
 )
 
 const (
-	APIDefaultUserFlag     = "api-default-user"
-	APIDefaultPasswordFlag = "api-default-password"
-	APISecretFlag          = "api-secret"
-	APIgRPCPortFlag        = "api-grpc-port"
-	APIHTTPPortFlag        = "api-http-port"
-	DBUserNameFlag         = "db-user-name"
-	DBPasswordFlag         = "db-password"
-	DBHostFlag             = "db-host"
-	DBNameFlag             = "db-name"
-	DBPortFlag             = "db-port"
-	DBInMemoryFlag         = "db-in-memory"
-	CreateDefaultTarget    = "target-default-create"
+	APIDefaultUserFlag         = "api-default-user"
+	APIDefaultPasswordFlag     = "api-default-password"
+	APISecretFlag              = "api-secret"
+	APIgRPCPortFlag            = "api-grpc-port"
+	APIHTTPPortFlag            = "api-http-port"
+	APICORSAllowedOriginsFlags = "api-cors-allowed-origins"
+	APICORSAllowedHeadersFlags = "api-cors-allowed-headers"
+	APICORSAllowedMethodsFlags = "api-cors-allowed-methods"
+	DBUserNameFlag             = "db-user-name"
+	DBPasswordFlag             = "db-password"
+	DBHostFlag                 = "db-host"
+	DBNameFlag                 = "db-name"
+	DBPortFlag                 = "db-port"
+	DBInMemoryFlag             = "db-in-memory"
+	CreateDefaultTarget        = "target-default-create"
 
 	DefaultAPIDefaultUser      = "clouditor"
 	DefaultAPIDefaultPassword  = "clouditor"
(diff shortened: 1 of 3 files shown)


rest/rest.go
```diff
@@ -45,16 +48,91 @@
...
+	allowedOrigins []string
+
+	// allowedHeaders contains a list of the allowed headers
+	allowedHeaders []string
+
+	// allowedMethods contains a list of the allowed methods
+	allowedMethods []string
+}
+
+// CORSConfigOption represents functional-style options to modify the CORS configuration in RunServer.
```
lebogg, line 78: Is there a convention for naming function types with suffix `Func` or similar?

rest/rest.go
```diff
@@ -45,16 +48,91 @@
...
+var (
+	// DefaultAllowedOrigins contains an nil slice, as per default, no origins are allowed.
+	DefaultAllowedOrigins []string = nil
+
+	// DefaultAllowedHeaders contains sensible defaults for the Access-Control-Allow-Headers header.
+	// Please adjust accordingly in production using WithAllowedHeaders.
+	DefaultAllowedHeaders = []string{"Content-Type", "Accept", "Authorization"}
+
+	// DefaultAllowedMethods contains sensible defaults for the Access-Control-Allow-Methods header.
+	// Please adjust accordingly in production using WithAllowedMethods.
```
lebogg, line 90: Is it ok to allow POST, PUT and DELETE methods as default as they are changing stuff? Or is it enough since we block allowed origins already per default?

lebogg, line 90: typo: "a nil slice"

rest/rest.go
```diff
@@ -103,5 +183,61 @@
 
 	log.Printf("Starting REST gateway on :%d", httpPort)
 
-	return srv.ListenAndServe()
+	sock, err = net.Listen("tcp", srv.Addr)
+	if err != nil {
```
lebogg, line 187: Add some error message?

rest/rest.go
```diff
@@ -103,5 +183,61 @@
...
+				w.Header().Set("Access-Control-Allow-Headers", strings.Join(cors.allowedHeaders, ","))
+				w.Header().Set("Access-Control-Allow-Methods", strings.Join(cors.allowedMethods, ","))
+				return
+			}
+		}
+
+		h.ServeHTTP(w, r)
+	})
+}
+
```
lebogg, line 233: typo: supplied

cmd/engine/engine.go
```diff
@@ -233,7 +242,13 @@
...
 	go func() {
-		err = rest.RunServer(context.Background(), grpcPort, httpPort)
+		err = rest.RunServer(context.Background(),
+			grpcPort,
+			httpPort,
+			rest.WithAllowedOrigins(viper.GetStringSlice(APICORSAllowedOriginsFlags)),
+			rest.WithAllowedHeaders(viper.GetStringSlice(APICORSAllowedHeadersFlags)),
+			rest.WithAllowedMethods(viper.GetStringSlice(APICORSAllowedMethodsFlags)),
+		)
 		if errors.Is(err, http.ErrServerClosed) {
```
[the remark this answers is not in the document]
oxisto: `ErrServerClosed` is not a real error, it indicates a graceful shutdown of the server, so we can just exit here. All other errors are printed out a few lines below

rest/rest.go
```diff
@@ -2,3 +1,4 @@
+// Copyright 2016-2022 Fraunhofer AISEC
```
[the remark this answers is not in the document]
oxisto: I would change it if I see it, otherwise feel free to create a PR that creates it everywhere, but I would do it separately.

rest/rest.go
oxisto: Probably, but in this case we want to have a functional-style option and the convention there is to append `Option` to it and "hide" a little bit that this is a function type for user convenience.

rest/rest.go
oxisto: I think it should be ok, as we are blocking origins per default.

rest/rest.go
lebogg, line 233:
```suggestion
// originAllowed checks if the supplied origin is allowed according to our global CORS configuration.
```

lebogg, line 233:
```suggestion
// originAllowed checks if the supplied origin is allowed according to our global CORS configuration.
```

cmd/engine/engine.go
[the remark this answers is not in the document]
lebogg: Maybe I understand it wrong, but the entire program is stopped (also with "0" indicating a success)? So the user does not know that the server is being currently stopping now and he has to retry starting the engine later?

rest/rest.go
oxisto: I wanted to emulate the behaviour of the previous `ListenAndServe` which also returns the error directly. The only reason for the change is that I need to access the `sock`, to get the actual port used for unit tests. The unit tests are started with port `0`, which is the "next available one".

cmd/engine/engine.go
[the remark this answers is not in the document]
oxisto: The "error" occurs when the server is gracefully closed inside the `RunServer` function. This happens for example if the user signals ^C to the program or somebody like Docker or Kubernetes sends a SIGSTOP/SIGKILL to the process. Then we can gracefully shutdown the REST server and exit.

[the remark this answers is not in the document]
lebogg: Ah now I get it. Thx!

lebogg (approved):

PR #479, opened 2022-02-02
Implementing write access to metric database via REST API

oxisto:
This adds two new RPC calls to create and update metrics via gRPC. They are also exposed via the REST API


api/assessment/metric.go
```diff
@@ -2,10 +2,17 @@ package assessment
 
 import (
 	"encoding/json"
+	"errors"
+)
+
+var (
+	//ErrMetricIdMissing   = errors.New("metric id is missing")
```
garuppel, line 9:
```suggestion
```

service/orchestrator/metrics.go
```diff
@@ -0,0 +1,124 @@
...
+	err = json.Unmarshal(b, &metrics)
+	if err != nil {
+		return fmt.Errorf("error in JSON marshal: %w", err)
+	}
+
+	return nil
+}
+
+// CreateMetric creates a new metric in the database.
+func (*Service) CreateMetric(_ context.Context, req *orchestrator.CreateMetricRequest) (metric *assessment.Metric, err error) {
```
lebogg, line 59: StoreMetric would be more consistent with other RPCs like StoreAssessmentResult or StoreEvidence

oxisto: I would argue, that this creates a new metric and store assessment result stores an already existing evidence. It is also more consistent with the style guide for CRUD operations.

lebogg: Ok, sold!

lebogg (approved):

PR #483, opened 2022-02-04
Implementing JWT/JWK-based API security

oxisto:
This PR (finally) introduces a major feature: Authentication in the gRPC and REST API. 

# Authentication Middleware for gRPC Services

On the server side, it uses an interception feature of [go-grpc-middleware](https://github.com/grpc-ecosystem/go-grpc-middleware) to insert a `grpc.UnaryServerInterceptor` into the gRPC server that requires authentication credentials in each gRPC call. In the `service` package, the function `ConfigureAuth` can be used to create an appropriate [`AuthFunc`](https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/auth#AuthFunc) that is needed by the gRPC middleware. The `ConfigureAuth` has two functional-style options to configure the validation of the tokens transmitted by the client. Either a ECDSA public key can directly be specified (using `WithPublicKey`) or JWKS can be used to off-load the validation to an external server (using `WithJWKSURL`).

# Providing Authentications for Client

On the client side, a so called `Authorizer` can be used to insert Bearer tokens in each RPC-call using `grpc.WithPerRPCCredentials`. For now, only one `InternalAuthorizer` exists, which authenticates a client against the Clouditor Authentication Services (using the `Login` call using username and password) and uses the returned access token to provide a bearer token. In the future, we probably want an `OAuthAuthorizer` which performs a complete OAuth flow to retrieve an access token and use that in a future RPC call.

# Usage of Asymmetric Keys in Authentication Service

This PR makes the internal Clouditor Authentication Service a little bit more standard compliant. However, we will probably not ever implement a full OAuth 2.0 server:
- Switched from symmetric keys to sign tokens to asymmetric ECDSA keys
- Provide public ECDS using JWKS using the `ListPublicKeys` RPC call, which is also exposed under the `/.well-known/jwks.json` REST path
- Provide the possibility to persist API key, protected by a password using PKCS#8 and PBES2. Options in the server can be used to configure this (`WithApiKeyPassword`, `WithApiKeyPath`, `WithApiKeySaveOnCreate`)


proto/auth.proto
```diff
@@ -36,11 +37,21 @@ service Authentication {
...
       response_body : "*"
     };
   };
+
+  // ListPublicKeys lists all available public keys of the authentication server
+  // in the JSON Web Key Set (JWKS) format.
+  rpc ListPublicKeys(ListPublicKeysRequest) returns (ListPublicResponse) {
+    option (google.api.http) = {
+      get : "/.well-known/jwks.json"
+      response_body : "*"
```
lebogg, line 52: Test comment

cli/commands/tool/tool.go
```diff
@@ -92,11 +89,7 @@ func NewShowToolCommand() *cobra.Command {
 				return nil
 			}
 
-			if conn, err = grpc.Dial(session.URL, grpc.WithTransportCredentials(insecure.NewCredentials())); err != nil {
-				return fmt.Errorf("could not connect: %w", err)
-			}
-
-			client = orchestrator.NewOrchestratorClient(conn)
+			client = orchestrator.NewOrchestratorClient(session)
```
garuppel, line 92: Update description in l73 to NewShowToolCommand()

garuppel: And update l39 to NewListToolsCommand()

cmd/engine/engine.go
```diff
@@ -81,7 +85,6 @@ const (
 	DefaultAPIDefaultPassword  = "clouditor"
 	DefaultAPISecret           = "changeme"
 	DefaultAPIgRPCPort         = 9090
-	DefaultAPIHTTPPort         = 8080
```
garuppel, line 84: Why did you delete the DefaultAPIHTTPPort?

rest/rest_test.go
```diff
@@ -207,7 +243,69 @@ func TestCORS(t *testing.T) {
...
 			}
+
+			if tt.wantResponse != nil {
+				tt.wantResponse(t, resp)
+			}
 		})
 	}
+}
 
+func Test_corsConfig_OriginAllowed(t *testing.T) {
```
lebogg, line 254: I think I use the form "TestCorsConfig_OriginAllowed". But your way is probably the cleanest one since it is more readable and allows differentiating between exported and unexported objects.

service/assessment/bufconn_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"clouditor.io/clouditor/api/evidence"
+	"clouditor.io/clouditor/api/orchestrator"
+	service_auth "clouditor.io/clouditor/service/auth"
+	service_evidence "clouditor.io/clouditor/service/evidence"
+	service_orchestrator "clouditor.io/clouditor/service/orchestrator"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/test/bufconn"
+)
+
+const DefaultBufferSize = 1024 * 1024 * 2
```
lebogg, line 42: I think I only added the 2 for debugging purposes, but I accidentally left it. 1024*1024 would also do the trick but the 2 shouldn't hurt as well.

service/discovery/discovery.go
```diff
@@ -107,20 +126,35 @@
 	return s
 }
 
-func (s *Service) initAssessmentStream() error {
-	// Establish connection to assessment component
-	target := s.assessmentAddress
-	log.Infof("Establishing connection to assessment (%v)", target)
+// UsesAuthorizer implements UsesAuthorizer
```
lebogg, line 129:
```suggestion
// SetAuthorizer implements UsesAuthorizer
```

service/discovery/discovery.go
```diff
@@ -107,20 +126,35 @@
...
+func (s *Service) SetAuthorizer(auth service.Authorizer) {
+	s.authorizer = auth
+}
+
+func (s *Service) Authorizer() service.Authorizer {
```
lebogg, line 134:
```suggestion
\\ Authorizer implements UsesAuthorizer
func (s *Service) Authorizer() service.Authorizer {
```

service/auth.go
```diff
@@ -0,0 +1,178 @@
+// Copyright 2022 Fraunhofer AISEC
```
lebogg, line 1: why is this package not in the corresponding auth subfolder?

service/auth.go
```diff
@@ -0,0 +1,178 @@
...
+	// stored public key will be used
+	Jwks *keyfunc.JWKS
+
+	// publicKey will be used to validate API tokens, if JWKS is not enabled
+	publicKey *ecdsa.PublicKey
+
+	AuthFunc grpc_auth.AuthFunc
+}
+
+var DefaultJwksUrl = "http://localhost:8080/.well-known/jwks.json"
```
lebogg, line 63:
```suggestion
const DefaultJwksUrl = "http://localhost:8080/.well-known/jwks.json"
```

service/auth.go
```diff
@@ -0,0 +1,178 @@
...
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+var log *logrus.Entry
+
+type AuthConfig struct {
+	jwksUrl string
```
lebogg, line 50: jwksURL or JWKSURL, see https://github.com/golang/go/wiki/CodeReviewComments#initialisms

service/auth.go
```diff
@@ -0,0 +1,178 @@
...
+	useJwks bool
```
lebogg, line 51: useJWKS

service/auth.go
```diff
@@ -0,0 +1,178 @@
...
+
+	// Jwks contains a JSON Web Key Set, that is used if JWKS support is enabled. Otherwise a
+	// stored public key will be used
+	Jwks *keyfunc.JWKS
```
lebogg, line 55: JWKS

service/auth.go
```diff
@@ -0,0 +1,178 @@
...
+	AuthFunc grpc_auth.AuthFunc
+}
+
+var DefaultJwksUrl = "http://localhost:8080/.well-known/jwks.json"
+
+func init() {
+	log = logrus.WithField("component", "service-auth")
+}
+
+type AuthOption func(*AuthConfig)
```
lebogg, line 69:
```suggestion
// AuthOption is a function-style option to fine-tune authentication
type AuthOption func(*AuthConfig)
```

service/auth.go
```diff
@@ -0,0 +1,178 @@
...
+
+func WithJwksUrl(url string) AuthOption {
```
lebogg, line 71:
```suggestion
// WithJwksUrl is an authentication option specifying the JWKS URL and setting useJwks to true
func WithJwksUrl(url string) AuthOption {
```

service/auth.go
```diff
@@ -0,0 +1,178 @@
...
+	return func(ac *AuthConfig) {
+		ac.jwksUrl = url
+		ac.useJwks = true
+	}
+}
+
+func WithPublicKey(publicKey *ecdsa.PublicKey) AuthOption {
```
lebogg, line 78:
```suggestion
// WithPublicKey is an authentication option specifying the public key
func WithPublicKey(publicKey *ecdsa.PublicKey) AuthOption {
```

service/auth.go
```diff
@@ -0,0 +1,178 @@
...
+	return func(ac *AuthConfig) {
+		ac.publicKey = publicKey
+	}
+}
+
+type AuthContextKey string
+
+func ConfigureAuth(opts ...AuthOption) *AuthConfig {
```
lebogg, line 86: Add comment

service/authorizer.go
```diff
@@ -0,0 +1,203 @@
...
+	"time"
+
+	"clouditor.io/clouditor/api/auth"
+	"golang.org/x/oauth2"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+type Authorizer interface {
```
lebogg, line 41: comment

clouditor.code-workspace
```diff
@@ -5,26 +5,6 @@
...
-			"lebogg",
-			"oxisto",
-			"protobuf",
-			"Rego",
-			"stretchr",
-			"structpb",
-			"timestamppb",
-			"Warnf"
-		],
 		"workbench.colorCustomizations": {
```
lebogg, line 8: Are you sure you want to remove your words for visual code :)

lebogg (commented):
First initial review. I will inspect pem and auth stuff more closely tomorrow

proto/auth.proto
```diff
@@ -36,11 +37,21 @@ service Authentication {
...
       body : "*"
       response_body : "*"
     };
   };
+
+  // ListPublicKeys lists all available public keys of the authentication server
+  // in the JSON Web Key Set (JWKS) format.
+  rpc ListPublicKeys(ListPublicKeysRequest) returns (ListPublicResponse) {
+    option (google.api.http) = {
+      get : "/.well-known/jwks.json"
```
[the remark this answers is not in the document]
oxisto: Yes, this is intentional. The `(domain)/.well-known` path is a convention to include several public information about a domain or an endpoint. One of them is the list of public keys, if this domain supports an authorisation server. It is widely used. See for example https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-key-sets

rest/rest.go
```diff
@@ -75,8 +74,8 @@ type corsConfig struct {
 	allowedMethods []string
 }
 
-// CORSConfigOption represents functional-style options to modify the CORS configuration in RunServer.
-type CORSConfigOption func(*corsConfig)
+// ServerConfigOption represents functional-style options to modify the server configuration in RunServer.
+type ServerConfigOption func(*corsConfig, *runtime.ServeMux)
 
 var (
```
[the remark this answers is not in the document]
oxisto: If you refer to the allowed headers and methods: a `[]string` cannot be a const (for some reason).

service/auth/auth.go
```diff
@@ -67,6 +75,23 @@ func init() {
 	log = logrus.WithField("component", "auth")
 }
 
+func NewService() *Service {
+	key, _ := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
```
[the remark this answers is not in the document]
oxisto: Without being a complete export in cryptography, but p256 seems to be a good sensible default for many years to come (ignoring quantum computers).

service/auth/auth.go
```diff
@@ -84,11 +109,33 @@ func (s Service) Login(_ context.Context, request *auth.LoginRequest) (response
...
+	response = &auth.LoginResponse{
+		AccessToken: token,
+		TokenType:   "Bearer",
+		Expiry:      timestamppb.New(expiry),
+	}
+
+	return response, nil
+}
+
+func (s Service) ListPublicKeys(_ context.Context, _ *auth.ListPublicKeysRequest) (response *auth.ListPublicResponse, err error) {
```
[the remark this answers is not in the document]
oxisto: Not quite sure what you mean?

proto/auth.proto
```diff
@@ -57,4 +68,28 @@ message LoginRequest {
...
+
+  // expiry is an optional expiration time of the access token.
+  google.protobuf.Timestamp expiry = 4;
+}
+
+message ListPublicKeysRequest {}
+message ListPublicResponse { repeated JsonWebKey keys = 1; }
+
+message JsonWebKey {
+  string kid = 1;
```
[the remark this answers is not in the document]
oxisto: JWK does not specify a specific format for keys. Examples includes simple numbering or default or a key thumbprint.

service/auth/auth.go
```diff
@@ -163,9 +210,7 @@ func (Service) CreateDefaultUser(username string, password string) {
 
 // issueToken issues a JWT-based token
 func (s Service) issueToken(subject string, fullName string, email string, expiry time.Time) (token string, err error) {
-	key := []byte(s.TokenSecret)
-
-	claims := jwt.NewWithClaims(jwt.SigningMethodHS256,
+	claims := jwt.NewWithClaims(jwt.SigningMethodES256,
```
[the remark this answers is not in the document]
oxisto: Yes, it corresponds to the used ECDSA private key and we are only using P256 keys.

rest/rest_test.go
oxisto: This was the way it was autogenerated 🤷

service/auth.go
oxisto: Because this is not part of the auth *service*, but rather functions related to the authentication between services and/or clients. I was also thinking moving it into a `service/common` or `service/middleware` package. But it is definitely NOT a part of `service/auth`, which contains our authentication service (or we need to rename that package).

service/auth.go
oxisto: Should it be `DefaultJWKSURL`?

service/auth.go
oxisto: `WithJWKSURL`?

service/auth/pem.go
```diff
@@ -0,0 +1,223 @@
...
+	oidAESCBC         = asn1.ObjectIdentifier{1, 2, 840, 113549, 1, 5, 16, 12}
+	oidHMACWithSHA256 = asn1.ObjectIdentifier{1, 2, 840, 113549, 1, 5, 9}
+	oidPBKDF2         = asn1.ObjectIdentifier{1, 2, 840, 113549, 1, 5, 12}
+	oidPBES2          = asn1.ObjectIdentifier{1, 2, 840, 113549, 1, 5, 13}
+)
+
+var ErrNotECPrivateKey = errors.New("key is not a valid EC private key")
+
+// See https://datatracker.ietf.org/doc/html/rfc8018#appendix-A.2
+type PBKDF2Params struct {
```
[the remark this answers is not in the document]
oxisto: Coin flip? :D

rest/rest.go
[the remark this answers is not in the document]
lebogg: Yes, lines are off here... Yes, only the port (integer) could be constant. The others can't. One workarround would be to introduce methods like `getDefaultAllowedMethods(){return X}` but the question is if this makes the code more or even less readable.

Since they are all defined as non-constant variables they could also being put into the other `var` block but I guess it is intentional to separate them. Maybe then put a comment before and/or for each Default value to indicate that they are constants.

service/auth/auth.go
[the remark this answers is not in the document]
lebogg: Ok, I'm no expert either. I was just wondering if we can be more future safe

service/auth/auth.go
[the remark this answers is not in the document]
lebogg: The method name gives the impression that a list of public keys is returned. But the list is always of size one with the same key ID. So for the case that there are multiple keys only one will be returned. Maybe it is some convention to have a list of keys and we only use one?

service/auth/auth.go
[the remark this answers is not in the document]
lebogg: Ok, one could verify the key length via `s.apiKey.Params().Bitsize` but we can leave it like that if you say the P256 won't change anyway.

service/auth/pem.go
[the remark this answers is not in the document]
lebogg: Yes, we let the coin flip in SB LVI decide. You choose head or tails... Ok we can leave it like that

rest/rest_test.go
lebogg: Right, me too. But when I autogenerate, e.g., issueToken I get `TestService_issueToken`

service/auth.go
lebogg: Thx for explanation. Then, I would advocate the `service/common` approach. Or maybe even `pkg` like stated in [this stackoverflow answer](https://stackoverflow.com/a/48872252/16445647).

OT: Looking at the 2nd alternative of this answer, I start thinking about putting the protobuf files (from `proto` folder) and generated swagger files (`openapi` folder) into the `api` directory as well for a more compact view of the project.

service/auth/auth.go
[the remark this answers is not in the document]
oxisto: Yes, it follows RFC 7517 to return a JSON Web Key *set*, independently how many keys are in the set. This is documented in the protobuf messages but not in the function. I can add it there as well.

service/auth/auth.go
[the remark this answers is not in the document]
oxisto: I can check, if we can derive the appropriate signing method automatically.

rest/rest_test.go
oxisto: The "logic" seems to be Test*MyType*_*myFunction* for exported types and Test_*myType*_*myFunction* for unexpected types.

service/auth.go
oxisto: > Thx for explanation. Then, I would advocate the `service/common` approach. Or maybe even `pkg` like stated in [this stackoverflow answer](https://stackoverflow.com/a/48872252/16445647).
> 

I am against `pkg` for various reasons that we can discuss while drinking several beers. `service/common` or `service/middleware` is fine for me.

> OT: Looking at the 2nd alternative of this answer, I start thinking about putting the protobuf files (from `proto` folder) and generated swagger files (`openapi` folder) into the `api` directory as well for a more compact view of the project.

Yes. This is something that I wanted to do for a long time anyway. I will create a separate issue for that I think.

service/discovery/discovery.go
```diff
@@ -46,13 +48,16 @@ import (
...
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/codes"
-	"google.golang.org/grpc/credentials/insecure"
 	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
 var log *logrus.Entry
 
+var (
```
garuppel, line 57: Should be const.

garuppel (commented):
First review

service/discovery/discovery.go
garuppel: Delete, it is not used any more.

garuppel: l87/l94 update names in comment

service/auth.go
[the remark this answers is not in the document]
oxisto: Yes, we will mitigate this by moving service/auth.go to service/middleware/auth.go and adding a package comment to service/middleware.

service/auth/pem.go
```diff
@@ -0,0 +1,248 @@
...
+)
+
+var (
+	oidAESCBC         = asn1.ObjectIdentifier{1, 2, 840, 113549, 1, 5, 16, 12}
+	oidHMACWithSHA256 = asn1.ObjectIdentifier{1, 2, 840, 113549, 1, 5, 9}
+	oidPBKDF2         = asn1.ObjectIdentifier{1, 2, 840, 113549, 1, 5, 12}
+	oidPBES2          = asn1.ObjectIdentifier{1, 2, 840, 113549, 1, 5, 13}
+)
+
+var ErrNotECPrivateKey = errors.New("key is not a valid EC private key")
```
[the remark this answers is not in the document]
oxisto: errors.new cannot be const

service/auth/pem.go
```diff
@@ -0,0 +1,248 @@
...
+	"encoding/asn1"
+	"encoding/pem"
+	"errors"
+	"fmt"
+	"io"
+
+	"golang.org/x/crypto/pbkdf2"
+)
+
+var (
```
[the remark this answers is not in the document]
oxisto: they cannot be const, since they are initialised structs, not primitive types.

service/auth/pem.go
```diff
@@ -0,0 +1,248 @@
...
+	EncryptionScheme  EncryptionScheme
+}
+
+// See https://datatracker.ietf.org/doc/html/rfc5958#section-3
+type EncryptionAlgorithmIdentifier struct {
+	Algorithm asn1.ObjectIdentifier
+	Params    PBES2Params
+}
+
+// See https://datatracker.ietf.org/doc/html/rfc5958#section-3
```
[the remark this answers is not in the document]
oxisto: Yes, this was an oversight. Will add godoc compliant comments.

service/discovery/discovery.go
oxisto: See above. cannot be made const.

rest/rest.go
```diff
@@ -125,16 +134,16 @@ func init() {
...
-
 	mux := runtime.NewServeMux()
 
+	for _, o := range serverOpts {
+		o(cors, mux)
+	}
+
 	opts := []grpc.DialOption{grpc.WithTransportCredentials(insecure.NewCredentials())}
 
 	if err := auth.RegisterAuthenticationHandlerFromEndpoint(ctx, mux, fmt.Sprintf("localhost:%d", grpcPort), opts); err != nil {
```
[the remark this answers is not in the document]
oxisto: In the future, we probably want to make this configurable, e.g. only have the orchestrator here but I would tackle this in a separate issue.

Issue #491, opened 2022-02-08
Sometimes gRPC GW plugin generates Orchestrator code with assessment import as alias (`assessment_0`) - and sometimes not


lebogg:
**Describe the bug**
When generating the API files for gRPC, the Gateway part of Orchestrator (`orchestrator.ob.gw.go`) sometimes imports the `api/assessment` with an
alias:
`import(...	assessment_0 "clouditor.io/clouditor/api/assessment ..."`
and sometimes not:
`import (... "clouditor.io/clouditor/api/assessment ..."`

**To Reproduce**
1. Within terminal go to project root
2. Execute `go generate ./...` multiple times until `orchestrator.pb.gw.go` has changed

**Expected behavior**
Consistent import of `assessment` and in particular, importing it without alias since it is not needed.

**Potential source of error**
Maybe something is wrong in `proto/compile_proto_files.go:36` (line for generating Orchestrator-related API code)

**Additional context**
It is no serious bug since nothing will break. But it is not desired either.

PR #498, opened 2022-02-09
Moving protobuf files into `api` folder

oxisto:


oxisto:
@lebogg Maybe this also fixes #491 ?

api/assessment/assessment.proto
```diff
@@ -31,10 +31,10 @@ package clouditor;
 import "google/api/annotations.proto";
 import "google/protobuf/empty.proto";
 import "google/protobuf/timestamp.proto";
-import "evidence.proto";
-import "metric.proto";
+import "api/evidence/evidence.proto";
+import "api/assessment/metric.proto";
 
-option go_package = "api/assessment";
+option go_package = "clouditor.io/clouditor/api/assessment;assessment";
```
lebogg, line 37: Is the ";assessment" at the end intended?

api/assessment/assessment.proto
```diff
@@ -31,10 +31,10 @@ package clouditor;
 import "google/api/annotations.proto";
 import "google/protobuf/empty.proto";
 import "google/protobuf/timestamp.proto";
-import "evidence.proto";
-import "metric.proto";
+import "api/evidence/evidence.proto";
```
lebogg, line 34: here we do not need the full path (with "clouditor.io" prefix)?

proto.go
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright 2016-2020 Fraunhofer AISEC
```
lebogg, line 2:
```suggestion
 * Copyright 2021-2022 Fraunhofer AISEC
```

lebogg: I think we started with Go last year? :)

lebogg (commented):
Seems good. Only some questions to import paths in proto files. 

I downloaded the protobuf files for third party folder (for the newest protoc version). But only struct.proto has changed and only in a comment. Should I update it, nevertheless?

lebogg:
And yes, #491 seems to be fixed now. I also generated it multiple times. No weird alias occurrences anymore :)

api/assessment/assessment.proto
oxisto: That is the syntax for <fullpath>;<relative_path>. This way the imports are resolved correctly according to the full path and the files are placed in the package according to the relative_path.

proto.go
oxisto: The overall project started in 2016, so I would use this as a start year

api/assessment/assessment.proto
lebogg: Nice to know. Do you have a link to DOCs pointing that out?

proto.go
lebogg:
```suggestion
 * Copyright 2016-2022 Fraunhofer AISEC
```

lebogg: I changed it, accordingly.

api/assessment/assessment.proto
oxisto: Looks like it might not be needed any more. https://developers.google.com/protocol-buffers/docs/gotutorial#defining-your-protocol-format. Seems the last part of the path will be used by default.

lebogg: I removed them

lebogg (approved):

PR #502, opened 2022-02-10
Implementing refresh_token flow in CLI

oxisto:
This PR introduces the possibility to refresh the access token in the CLI using a refresh token. This also introduces a new token endpoint in the authentication server, that is compliant to the OAuth 2.0 standard (when used via the REST API). The CLI takes a little shortcut and uses this endpoint directly via gRPC.


cli/session.go
```diff
@@ -71,16 +80,15 @@
 	DefaultSessionFolder = fmt.Sprintf("%s/.clouditor/", home)
 }
 
-func NewSession(url string, opts ...grpc.DialOption) (session *Session, err error) {
+func NewSession(url string) (session *Session, err error) {
```
lebogg, line 83: since CLI is only used on localhost, you removed opts?

service/auth/auth.go
```diff
@@ -279,23 +293,80 @@
...
 
-	if token, err = s.issueToken(u.Username, u.FullName, u.Email, expiry); err != nil {
-		return nil, status.Errorf(codes.Internal, "token issue failed: %v", err)
+	return response, nil
+}
+
+// Token aims to be a OAuth 2.0 compliant token endpoint. Currently, only the refresh_token grant type
+// is supported. This function only takes care of the application behavior, the actual implementation of
+// a HTTP endpoint around this functionality is provided by the gRPC gateway and is modified by the
+// OAuthErrorHandler function, which needs to be registed using runtime.WithErrorHandler in the gRPC
```
lebogg, line 317:
```suggestion
// OAuthErrorHandler function, which needs to be registered using runtime.WithErrorHandler in the gRPC
```

service/auth/auth.go
```diff
@@ -279,23 +293,80 @@
...
+// gateway.
+func (s *Service) Token(_ context.Context, req *auth.TokenRequest) (response *auth.TokenResponse, err error) {
+	var (
+		token  string
+		user   *auth.User
+		claims jwt.RegisteredClaims
+		expiry = time.Now().Add(DefaultPeriodOfValidity)
+	)
+
+	if req == nil {
```
lebogg, line 327: Should we really check a gRPC request for being nil? Can it even be nil? I am pretty sure we didn't do it before/everywhere

service/auth/auth.go
```diff
@@ -279,23 +293,80 @@
...
 	}
 
-	response = &auth.LoginResponse{
+	err = s.storage.Get(&user, "username = ?", claims.Subject)
+	if errors.Is(err, persistence.ErrRecordNotFound) {
+		// User not found, we cannot successfully authenticate this refresh token
+		return nil, status.Error(codes.InvalidArgument, "invalid_grant")
+	} else if err != nil {
+		// A database connection error has occurred, return an an internal error but without any details
+		return nil, status.Errorf(codes.Internal, "internal database error occurred: %v", err)
```
lebogg, line 350: "Without any details" -> Then don't include the error message

service/auth/auth.go
```diff
@@ -387,20 +458,32 @@
...
 			}},
 	)
 	claims.Header["kid"] = DefaultKeyID
 
-	token, err = claims.SignedString(s.apiKey)
+	token, err = claims.SignedString(key)
+	return
+}
+
+func issueRefreshToken(key *ecdsa.PrivateKey, subject string) (token string, err error) {
```
lebogg, line 478: Another non-expert question: Should something happen with the old token? I guess it is kept as a "reference".

lebogg (commented):
Looks good. Only some typos and minor questions

api/authorizer.go
```diff
@@ -139,10 +169,18 @@ func (i *InternalAuthorizer) Token() (*oauth2.Token, error) {
...
-		Username: i.Username,
-		Password: i.Password,
-	})
+	// TODO(oxisto): In the future we should split this into two authorizers
+	if i.token != nil && i.token.RefreshToken != "" {
+		resp, err = i.client.Token(context.TODO(), &auth.TokenRequest{
+			GrantType:    "refresh_token",
+			RefreshToken: i.token.RefreshToken,
+		})
+	} else {
```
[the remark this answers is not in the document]
oxisto: This code is executed at the beginning when the first token is fetched using username and password.

cli/session.go
oxisto: I saw no use case for additional options. You specify the host in the url param. We did not use them, even in the unit tests.

service/auth/auth.go
oxisto: Not sure if I understand your question? which old token?

api/authorizer_test.go
```diff
@@ -0,0 +1,154 @@
...
+		token       *oauth2.Token
+	}
+	tests := []struct {
+		name    string
+		fields  fields
+		want    *oauth2.Token
+		wantErr bool
+	}{
+		{
+			name: "Fetch access token with refresh token",
```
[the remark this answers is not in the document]
oxisto: Added two more flows

PR #516, opened 2022-02-15
Tie-in of metric configuration with Rego evaluation

oxisto:
This PR ties in the metric configuration with the Repo evaluation. It introduces a new interfaces called `MetricConfigurationSource`, which is expected by the `RunEvidence` function. 

This interface can be used to retrieve a metric configuration for a particular metric (and target service). Currently, only the default cloud target service is used, as not every micro-service is aware of the target cloud service (yet). Currently, the assessment service implements this interfaces and fetches the metric configurations from the orchestrator (and caches them). 

In the future, we need a proper way to evict that cache based on an event (such as the configuration of a target cloud service changed).

### DIFF
--- a/service/assessment/assessment.go
+++ b/service/assessment/assessment.go
@@ -223,14 +223,12 @@ func (s *Service) handleEvidence(evidence *evidence.Evidence, resourceId string)
 	}
 
 	for i, data := range evaluations {
-		metricId := data["metricId"].(string)
+		metricId := data.MetricId
 
-		log.Infof("Evaluated evidence with metric '%v' as %v", metricId, data["compliant"])
+		log.Infof("Evaluated evidence with metric '%v' as %v", metricId, data.Compliant)
 
 		// Get output values of Rego evaluation. If they are not given, the zero value is used
-		operator, _ := data["operator"].(string)
-		targetValue := data["target_value"]
-		compliant, _ := data["compliant"].(bool)
+		targetValue := data.TargetValue
 
 		convertedTargetValue, err := convertTargetValue(targetValue)
 		if err != nil {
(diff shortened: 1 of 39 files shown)


service/assessment/assessment.go
```diff
@@ -361,27 +378,72 @@
...
+
+func (s *Service) MetricConfiguration(metric string) (config *assessment.MetricConfiguration, err error) {
+	var (
+		ok    bool
+		cache cachedConfiguration
+	)
+
+	// Lazy init of connection
+	if s.orchestratorStream == nil || s.orchestratorClient == nil {
+		err := s.initOrchestratorStream()
```
lebogg, line 419:
```suggestion
		err = s.initOrchestratorStream()
```

service/assessment/assessment.go
```diff
@@ -361,27 +378,72 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("could not initialize connection to orchestrator: %v", err)
+		}
+	}
+
+	// Retrive our cached entry
```
lebogg, line 425:
```suggestion
	// Retrieve our cached entry
```

service/assessment/assessment.go
```diff
@@ -53,6 +56,16 @@
 	log = logrus.WithField("component", "assessment")
 }
 
+const (
+	// EvictionTime is the time after which an entry in the metric configuration is invalid
+	EvictionTime = time.Minute * 5
```
lebogg, line 61: As you mentioned in the PR Description, an event trigger (sth. like `onMetricConfigurationChanged()`)would be better but for now it should be ok, even if 5 minutes sounds too low for me since metric configurations shouldn't change that often. (Also the eviction time of 5 minutes is as long as the discovery interval^^)

service/assessment/assessment.go
```diff
@@ -76,6 +90,8 @@
 	// Currently, results are just stored as a map (=in-memory). In the future, we will use a DB.
 	results map[string]*assessment.AssessmentResult
 
+	cachedConfigurations map[string]cachedConfiguration
```
lebogg, line 93:
```suggestion
    // cachedConfigurations holds cached metric configurations for faster access with key being the corresponding
    // metric name
	cachedConfigurations map[string]cachedConfiguration
```

lebogg (commented):
Much cleaner representation for rego output with the `Result` struct. Some minor comments and typos.

Issue #535, opened 2022-02-22
HandleEvidence: Make logging message more human-readable


lebogg:
**Describe the bug**
Currently, the log message of [handleEvidence](https://github.com/clouditor/clouditor/blob/f3f4517ac3740742d643b7960e354ef71744d6c9/service/assessment/assessment.go#L225) is, e.g.:
`INFO[0010] Evaluating evidence 1 (1) collected by 1 at seconds:1645574399 nanos:999999999 `

**To Reproduce**
Run engine and assess an evidence.

**Expected behavior**
A human readable message, e.g.:
`INFO[0010] Evaluating evidence with ID: 1 (Resource ID: 1) collected by Tool 1 at 2022-02-22 23:59:59.999999999` or
`INFO[0010] Evaluating evidence (Evidence ID: 1 | Resource ID: 1 | Tool ID: 1 | Time: 2022-02-22 23:59:59.999999999)`

**Additional context**
Use `Timestamp.AsTime()` of the `timestamppb` package

anatheka:
Already fixed in the past.

PR #540, opened 2022-02-25
Improve logging message of `handleEvidence`

lebogg:
- #535 

Before:
`INFO[0010] Evaluating evidence a0527fed-4601-46ea-86e3-f7723fcd5f0d (/subscriptions/.../vmDisk) collected by Clouditor Evidences Collection at seconds:1645574399 nanos:999999999 `
Now:
`INFO[0011] Assessing evidence with ID: a0527fed-4601-46ea-86e3-f7723fcd5f0d | Resource ID: /subscriptions/.../vmDisk | Tool ID: Clouditor Evidences Collection | Time: 2022-02-25 15:04:07.270168626 +0000 UTC  component=assessment
`


oxisto:
Would it make sense to pipe this information into logrus as additional fields? similar to the way it is done for gRPC requests? This would then separate the logic from the display a little bit more.

lebogg:
> Would it make sense to pipe this information into logrus as additional fields? similar to the way it is done for gRPC requests? This would then separate the logic from the display a little bit more.

Oh yea, good idea. In particular with colors, I can imagine it will look much more appealing. I will test it out. Maybe we can even add a separate color for the last component field for separating it from the actual message.

Issue #551, opened 2022-03-02
Implement endpoints for updating and retrieving a certification status


oxisto:
- [ ] Design API endpoint
- [ ] Implement it

PR #560, opened 2022-03-07
API for creating, updating, getting, and deleting a certificate

immqu:
This PR adds an API for creating, updating, getting, and deleting certificates. The data model of certificates is based on ISO17021.


immqu:
fixes #551

api/orchestrator/orchestrator.proto
```diff
@@ -346,4 +376,42 @@ message Requirement {
...
+  string certificate_id = 1;
+	Certificate certificate = 2;
+}
+
+message CreateCertificateRequest { Certificate certificate = 1; }
+
+message RemoveCertificateRequest { string certificate_id = 1; }
+
+message Certificate {
+  string ID = 1;
```
oxisto, line 398: This should be `id` or is there then a problem with Gorm persistance?

immqu: I used ID because that is what gorm uses as the default primary key; I didn't try id though

oxisto: You will never get `ID` past mr. protobuf style guide @lebogg anyway :D

lebogg: Please don't do gitblame on that line :laughing:
`id` seems to work. I will push it.

api/orchestrator/orchestrator.proto
```diff
@@ -206,7 +206,7 @@ service Orchestrator {
 
   // Lists all a metric configurations (target value and operator) for a
   // specific service ID
-  rpc ListMetricConfigurations(ListMetricConfigurationRequest)
+  rpc ListMetrifcConfigurations(ListMetricConfigurationRequest)
```
oxisto, line 209: Typo ;)

persistence/gorm/gorm.go
```diff
@@ -67,17 +67,20 @@ func WithPostgres(host string, port int16) StorageOption {
...
+
 func init() {
 	log = logrus.WithField("component", "storage")
 }
 
 // NewStorage creates a new storage using GORM (which DB to use depends on the StorageOption)
 func NewStorage(opts ...StorageOption) (s persistence.Storage, err error) {
-	g := &storage{
-		config: gorm.Config{
-			Logger: logger.Default.LogMode(logger.Info),
```
oxisto: I would actually prefer logging mode silent, because otherwise gorm logs out all record not found messages, which is quite annoying; but having it configurable is also a good idea; maybe with a default of silent?

internal/testutil/orchestrator.go
```diff
@@ -0,0 +1,33 @@
+package testutil
+
+import (
+	"time"
+
+	"clouditor.io/clouditor/api/orchestrator"
+)
+
+// CreateCertificateMock creates a mock certificate creation request
+func CreateCertificateMock() *orchestrator.Certificate {
```
oxisto, line 10: Are multiple packages using this? otherwise I would move it to the corresponding _test file for now. Also a simple `var` would do. not sure why this should be a function

service/orchestrator/orchestrator.go
```diff
@@ -311,3 +312,100 @@ func (s *Service) informHook(result *assessment.AssessmentResult, err error) {
...
+	if req.CertificateId == "" {
+		return nil, status.Errorf(codes.NotFound, orchestrator.ErrCertIDIsMissing.Error())
+	}
+
+	response = new(orchestrator.Certificate)
+	err = s.storage.Get(response, "Id = ?", req.CertificateId)
+	if errors.Is(err, persistence.ErrRecordNotFound) {
+		return nil, status.Errorf(codes.NotFound, "certificate not found")
+	} else if err != nil {
+		return nil, status.Errorf(codes.Internal, "database error: %s", err)
```
oxisto, line 349: %v is our  practice here I think

service/orchestrator/orchestrator.go
```diff
@@ -311,3 +312,100 @@ func (s *Service) informHook(result *assessment.AssessmentResult, err error) {
...
+// ListCertificates implements method for getting a certificate, e.g. to show its state in the UI
+func (s *Service) ListCertificates(_ context.Context, req *orchestrator.ListCertificatesRequest) (response *orchestrator.ListCertificatesResponse, err error) {
+	if req == nil {
+		return nil, status.Errorf(codes.InvalidArgument, orchestrator.ErrRequestIsNil.Error())
+	}
+
+	response = new(orchestrator.ListCertificatesResponse)
+	err = s.storage.List(&response.Certificates)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "database error: %s", err)
```
oxisto, line 363: %v is our  practice here I think

service/orchestrator/orchestrator.go
```diff
@@ -311,3 +312,100 @@ func (s *Service) informHook(result *assessment.AssessmentResult, err error) {
...
+		return nil, status.Errorf(codes.InvalidArgument, "certificate id is empty")
+	}
+
+	if req.Certificate == nil {
+		return nil, status.Errorf(codes.InvalidArgument, "certificate is empty")
+	}
+
+	count, err := s.storage.Count(req.Certificate, "Certificate_id = ?", req.CertificateId)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "database error: %s", err)
```
oxisto, line 380: %v is our  practice here I think

service/orchestrator/orchestrator.go
```diff
@@ -311,3 +312,100 @@ func (s *Service) informHook(result *assessment.AssessmentResult, err error) {
...
+func (s *Service) RemoveCertificate(_ context.Context, req *orchestrator.RemoveCertificateRequest) (response *emptypb.Empty, err error) {
+	if req.CertificateId == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "certificate id is empty")
+	}
+
+	err = s.storage.Delete(&orchestrator.Certificate{}, "Id = ?", req.CertificateId)
+	if errors.Is(err, persistence.ErrRecordNotFound) {
+		return nil, status.Errorf(codes.NotFound, "certificate not found")
+	} else if err != nil {
+		return nil, status.Errorf(codes.Internal, "database error: %s", err)
```
oxisto, line 407: %v is our  practice here I think

service/orchestrator/orchestrator.go
```diff
@@ -311,3 +312,100 @@ func (s *Service) informHook(result *assessment.AssessmentResult, err error) {
...
+}
+
+// RemoveCertificate implements method for removing a certificate
+func (s *Service) RemoveCertificate(_ context.Context, req *orchestrator.RemoveCertificateRequest) (response *emptypb.Empty, err error) {
+	if req.CertificateId == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "certificate id is empty")
+	}
+
+	err = s.storage.Delete(&orchestrator.Certificate{}, "Id = ?", req.CertificateId)
+	if errors.Is(err, persistence.ErrRecordNotFound) {
```
oxisto, line 404: Just thinking out loud here, could we somehow wrap this into a `handleOrNotFound` function or something? we repeat this pattern quite often (not just here)

oxisto:
@immqu please have a look at the review comments

internal/testutil/orchestrator.go
lebogg: Yes, multiple packages are using it: Tests in orchestrator and gorm. But due to cyclic dependencies I created orchestrator directory/package within testutil.
The idea behind the function is to create multiple mock certificates with different properties when needed. When you have a variable, you would have to make a copy I guess each time?

service/orchestrator/orchestrator.go
lebogg: Good point to think about. Created issue #650

oxisto: Thanks, we can do in a seperate PR, resolving this.

api/orchestrator/orchestrator.proto
```diff
@@ -391,4 +429,52 @@ message Requirement {
...
+}
+
+message CreateCertificateRequest { Certificate certificate = 1; }
+
+message RemoveCertificateRequest { string certificate_id = 1; }
+
+// An ISO17021-based certificate
+message Certificate {
+  string id = 1;
+	string name = 2;
```
oxisto, line 459: Formatting

persistence/gorm/gorm_test.go
```diff
@@ -326,7 +358,7 @@ func Test_storage_Update(t *testing.T) {
 	assert.Equal(t, user.Password, gotUser.Password)
 	assert.Equal(t, user.Email, gotUser.Email)
 
-	// Testing cloud service (A table test now would be better, probably)
+	// Testing cloud service (A entryType test now would be better, probably)
```
oxisto, line 361: ?

persistence/gorm/gorm_test.go
```diff
@@ -1,14 +1,13 @@
...
 import (
-	"fmt"
-	"testing"
-
 	"clouditor.io/clouditor/api/auth"
 	"clouditor.io/clouditor/api/orchestrator"
+	"clouditor.io/clouditor/internal/testutil/orchestratortest"
 	"clouditor.io/clouditor/persistence"
-
+	"fmt"
```
oxisto, line 8: weird import order

service/orchestrator/orchestrator.go
```diff
@@ -257,3 +258,101 @@ func (s *Service) informHook(result *assessment.AssessmentResult, err error) {
...
+
+// CreateCertificate implements method for creating a new certificate
+func (s *Service) CreateCertificate(_ context.Context, req *orchestrator.CreateCertificateRequest) (response *emptypb.Empty, err error) {
+	if req == nil {
+		return nil, status.Errorf(codes.InvalidArgument, orchestrator.ErrRequestIsNil.Error())
+	}
+	if req.Certificate == nil {
+		return nil, status.Errorf(codes.InvalidArgument, orchestrator.ErrCertificateIsNil.Error())
+	}
+	fmt.Println("got request: ", req)
```
oxisto, line 270: Remove `fmt.Println`

service/orchestrator/orchestrator_test.go
```diff
@@ -26,29 +26,31 @@
...
-	"sync"
-	"testing"
-
 	"clouditor.io/clouditor/api/assessment"
 	"clouditor.io/clouditor/api/orchestrator"
 	"clouditor.io/clouditor/internal/testutil/clitest"
+	"clouditor.io/clouditor/internal/testutil/orchestratortest"
 	"clouditor.io/clouditor/persistence/inmemory"
-
+	"context"
```
oxisto, line 34: import order formatting

oxisto (commented):
Some minor formatting issues

lebogg:
@oxisto Dou you have any idea for solving the following GORM problem:
A certificate has a list of states. Both are migrated into GORM and accordingly associated. But we want that GORM to automatically fill the IDs of states (when Updating a certificate). If we let the id field empty, there will be one state only since it is overwritten every time. We also tried to exclude the id field hoping that GORM will solve it. But same problem.
There could be an option via gorm annotations of the state struct for default values. I didn't get it work but anyway, we cannot touch it since it is a generated file. There is a 3rd party package "protoc-gen-gorm" which I would like to avoid.

The only way that comes to my mind is, that we have to implement the lifecycle manager to always increment the state id if a new state for a certificate is created. It would be better though, if we just put the state w/o id and GORM does the rest.

persistence/gorm/gorm.go
```diff
@@ -66,20 +68,28 @@ func WithPostgres(host string, port int16, user string, pw string, db string) St
...
+	// Create storage with default gorm config
 	g := &storage{
 		config: gorm.Config{
 			Logger: logger.Default.LogMode(logger.Silent),
 		},
 	}
 
-	// Init storage
-	log.Println("Creating storage")
+	// Add or override default options
```
lebogg, line 92:
```suggestion
	// Add options and/or override default ones
```

service/orchestrator/orchestrator.go
```diff
@@ -257,3 +258,100 @@ func (s *Service) informHook(result *assessment.AssessmentResult, err error) {
...
+		return nil, status.Errorf(codes.InvalidArgument, orchestrator.ErrCertificateIsNil.Error())
+	}
+
+	// Persist the certificate in our database
+	err = s.storage.Create(req.Certificate)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "could not add certificate to the database: %v", err)
+	}
+
+	return &emptypb.Empty{}, nil
```
lebogg, line 277: He we should return newly created certificate. See also comment in .proto

lebogg (changes requested):
Good stuff. Some minor issues have to be addressed.

api/orchestrator/orchestrator.proto
```diff
@@ -391,4 +429,52 @@ message Requirement {
...
+
+message CreateCertificateRequest { Certificate certificate = 1; }
+
+message RemoveCertificateRequest { string certificate_id = 1; }
+
+// An ISO17021-based certificate
+message Certificate {
+  string id = 1;
+  string name = 2;
+  string service_id = 3;
```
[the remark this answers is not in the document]
oxisto: Actually, it does.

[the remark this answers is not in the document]
lebogg: I'm not sure. Either
- The message `CloudService` should contain at least a `Certificate certificate` for a "Has One" relation or 
- `Certificate` should contain a `CloudService cloudService` for a "Belongs To"?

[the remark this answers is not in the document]
oxisto: I suggest that we will worry about the associations later in a combined issue, since the same applies to metrics as well.

oxisto (approved):

PR #565, opened 2022-03-10
Changed auth URLs back to `/v1/auth`

oxisto:


api/authorizer_test.go
```diff
@@ -118,7 +110,7 @@ func TestNewOAuthAuthorizerFromClientCredentials(t *testing.T) {
 	var config = clientcredentials.Config{
 		ClientID:     "client",
 		ClientSecret: "secret",
-		TokenURL:     "/token",
+		TokenURL:     "/v1/auth/token",
```
lebogg, line 113: You could use here the constants for client id and secret from the testutil package as well. Anyway, the values of the strings doesn't matter for the test to succeed.

lebogg (approved):

PR #573, opened 2022-03-14
Delete import github.com/jinzhu/copier

### LINKED ISSUE #572: Get rid of github.com/jinzhu/copier

garuppel:
Closes #572

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,7 +22,6 @@ require (
 	github.com/google/uuid v1.3.0
 	github.com/grpc-ecosystem/go-grpc-middleware v1.3.0
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.8.0
-	github.com/jinzhu/copier v0.3.5
 	github.com/logrusorgru/aurora/v3 v3.0.0
 	github.com/mitchellh/mapstructure v1.4.3
 	github.com/open-policy-agent/opa v0.38.0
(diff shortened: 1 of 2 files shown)


internal/util/copy.go
```diff
@@ -0,0 +1,49 @@
...
+
+package util
+
+// DeepCopy copies an interface[] array
+func DeepCopy(original []interface{}) []interface{} {
+
+	var target []interface{}
+
+	for i := range original {
+		target = append(target, original[i])
```
oxisto, line 34: isn't this essentially the same as the internal `copy` function? Maybe we are not really in need of a "deep" copy of a slice in our tests anyway

internal/util/copy.go
```diff
@@ -0,0 +1,49 @@
...
+
+	return target
+}
+
+// DeepCopyOfMap copies a map
+func DeepCopyOfMap(originalMap map[string]interface{}) map[string]interface{} {
+	targetMap := make(map[string]interface{})
+
+	for key, value := range originalMap {
+		targetMap[key] = value
```
oxisto, line 45: Wouldn't a deep copy need to check, if value is of type map and then return a deep copy of that instead of the original value?

internal/util/copy.go
garuppel: No, I don't no why, but the internal copy function does not copy that.

internal/util/copy.go
garuppel: Probably, but in that case I don't know if it's very smart to write it our self instead of using a library.

internal/util/copy.go
oxisto: Ok, then lets keep this like this for now

oxisto (approved):
LGTM

PR #582, opened 2022-03-15
Performance and concurrency improvements in assessment

oxisto:
This PR adds a major performance improvement to the policy checking implementation. Instead of creating a new Rego query for each evaluation, we are using a `queryCache` struct to cache the prepared query for each metric (and its metric configuration).

This boots metric evaluation speed by roughly 10x.

### DIFF
--- a/policies/policies.go
+++ b/policies/policies.go
@@ -165,8 +165,18 @@ var cache queryCache = *newQueryCache()
 func RunMap(baseDir string, metric string, m map[string]interface{}, holder MetricConfigurationSource) (result *Result, err error) {
 	var query *rego.PreparedEvalQuery
 
-	// TODO(oxisto): This will only fetch the metric configuration once :(
-	query, err = cache.Get(metric, func(key string) (*rego.PreparedEvalQuery, error) {
+	// We need to check, if the metric configuration has been changed. Any caching of this
+	// configuration will be done by the MetricConfigurationSource.
+	config, err := holder.MetricConfiguration(metric)
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch metric configuration: %w", err)
+	}
+
+	// We build a key out of the metric and its configuration, so we are creating a new Rego implementation
+	// if the metric configuration (i.e. its hash) has changed.
+	var key = fmt.Sprintf("%s-%s", metric, config.Hash())
+
+	query, err = cache.Get(key, func(key string) (*rego.PreparedEvalQuery, error) {
 		var (
 			tx storage.Transaction
 		)
@@ -174,11 +184,6 @@ func RunMap(baseDir string, metric string, m map[string]interface{}, holder Metr
 		// Create paths for bundle directory and utility functions file
 		bundle := fmt.Sprintf("%s/policies/bundles/%s/", baseDir, metric)
 		operators := fmt.Sprintf("%s/policies/operators.rego", baseDir)
-
-		config, err := holder.MetricConfiguration(metric)
-		if err != nil {
-			return nil, fmt.Errorf("could not fetch metric configuration: %w", err)
-		}
 
 		c := map[string]interface{}{
 			"target_value": config.TargetValue.AsInterface(),
--- a/policies/policies.go
+++ b/policies/policies.go
@@ -44,15 +44,24 @@ import (
 
 var (
 	log = logrus.WithField("component", "policies")
-)
 
-// applicableMetrics stores a list of applicable metrics per resourceType
-var applicableMetrics = metricsResourceTypeCache{m: make(map[string][]string)}
+	// applicableMetrics stores a list of applicable metrics per resourceType
+	applicableMetrics = metricsResourceTypeCache{m: make(map[string][]string)}
+
+	cache = newQueryCache()
+)
 
 type metricsResourceTypeCache struct {
 	sync.RWMutex
 	m map[string][]string
 }
+
+type queryCache struct {
+	sync.Mutex
+	cache map[string]*rego.PreparedEvalQuery
+}
+
+type orElseFunc func(key string) (query *rego.PreparedEvalQuery, err error)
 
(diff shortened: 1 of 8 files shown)


service/evidence/evidence_store.go
```diff
@@ -104,14 +105,15 @@
 		req, err = stream.Recv()
 
 		// If no more input of the stream is available, return
-		if err == io.EOF {
+		if errors.Is(err, io.EOF) {
```
lebogg, line 108: `errors.Is` is not necessary here, but also not wrong and good coding practice. According to DOC: "Read must return EOF itself, not an error wrapping EOF, because callers will test for EOF using =="

service/evidence/evidence_store.go
```diff
@@ -123,6 +125,11 @@
 
 		// Send response back to the client
 		err = stream.Send(res)
+
+		// Check for send errors
+		if errors.Is(err, io.EOF) {
```
lebogg, line 130: For me, only the sender can signal an EOF (and not the receiver since it just receives and can't now when EOF is reached). But maybe I oversee sth. in the streaming model.

policies/policies.go
```diff
@@ -96,14 +120,20 @@
...
 
 				data = append(data, runMap)
 			}
 		}
+
+		// Set it and unlock
+		applicableMetrics.m[key] = metrics
+		applicableMetrics.Unlock()
+
+		log.Infof("Resource type %v has the following %v applicable metric(s): %v", key, len(applicableMetrics.m[key]), applicableMetrics.m[key])
```
lebogg, line 134: Keep in mind that the key is a concatenation of the resource types and the logging output may not look as you  would expect, e.g. 
```INFO[0002] Resource type StorageService-NetworkService-Networking-Resource has the following 4 applicable metric(s): [TransportEncryptionEnforced TLSVersion TransportEncryptionAlgorithm TransportEncryptionEnabled]  component=policies```

File: policies/policies.go [hunk 4]
```diff
@@ -117,64 +147,122 @@
...
+		return
 	}
 
-	c := map[string]interface{}{
-		"target_value": config.TargetValue.AsInterface(),
-		"operator":     config.Operator,
+	// Otherwise, the orElse function is executed to fetch the query
+	query, err = orElse(key)
+	if err != nil {
+		return nil, err
```
lebogg, line 184:
```suggestion
		return
```

policies/policies.go
```diff
@@ -117,64 +147,122 @@
...
 	}
 
-	// Convert camelCase metric in under_score_style for package name
-	metric = util.CamelCaseToSnakeCase(metric)
+	// Update the cache
+	qc.cache[key] = query
+	return
+}
+
+var cache queryCache = *newQueryCache()
```
lebogg, line 192:
```suggestion
var cache = *newQueryCache()
```

policies/policies.go
```diff
@@ -117,64 +147,122 @@
 	return data, nil
 }
 
-func RunMap(baseDir string, metric string, m map[string]interface{}, holder MetricConfigurationSource) (result *Result, err error) {
+type queryCache struct {
```
lebogg, line 150: Normally, I would expect that all types, structs and global variables are placed at the top of the code before all the methods and functions (It helps to read the code step by step, though)

service/evidence/evidence_store.go
oxisto: Yes, but it *seems* that the error that is returned here is actually a wrapped gRPC error, so I chose the `errors.Is` approach.

service/evidence/evidence_store.go
oxisto: I was getting EOF gRPC errors in the benchmark tests.

service/assessment/assessment.go
```diff
@@ -479,7 +493,9 @@
 	}
 
 	// Retrieve our cached entry
+	s.confMutex.Lock()
```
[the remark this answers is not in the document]
oxisto: Its a judgement call. In general, this is done by two different mutex structs. `Mutex` only knows `Lock`, which means that this mutex is locked for reading and writing (by all and from all). This is suitable for most scenarios.

There is also the more specialized variant `RWMutex` which allows to separate locking for reading and writing. The idea is that multiple readers can `RLock` the variable in parallel (which is not possible with the regular `Lock`). It will however prevent any write in the meantime. And the other way around, a regular (write) `Lock` will prevent any readers to access it while the writing lock is active. This is especially suited for scenarios where you have far more readers than writers and you want to enable concurrent reading.

policies/policies.go
oxisto: yes, but that is the key that is used, so I think it is ok.

policies/policies.go
oxisto: I actually did that on purpose, since I have no control over what `orElse` returns and I definitely want to return `nil` here and only the error.

service/evidence/evidence_store.go
lebogg: Ok, interesting. In the gRPC [stream example](https://github.com/grpc/grpc-go/blob/94ee3865e17c5645148f8d24a32dc50f40810f13/examples/route_guide/server/server.go#L125) they just return any error and maybe the EOF is hidden there.

lebogg (approved):
Go

service/assessment/benchmark_test.go
```diff
@@ -0,0 +1,158 @@
...
+		err := srv.Serve(sock)
+		if err != nil && err != http.ErrServerClosed {
+			log.Fatalf("Error while creating gRPC server: %v", err)
+		}
+	}()
+	defer srv.Stop()
+
+	wg.Add(n * m * 7)
+
+	var count int64 = 0
```
garuppel, line 56: Move to var block?

garuppel (commented):
LGTM

PR #586, opened 2022-03-16
Separate grpc calls

garuppel:
Decouple handleEvidence and sending to other components

- Add channel for evidences and assessment results
- Move grpc recv/ send evidences from/to Evidence Store in separate go routines
- Move grpc recv/send assessment results from/to Orchestrator in separate go routines
- Update tests

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -10,5 +10,11 @@
 "api":
 - any: ['api/**/*', '!api/evaluation/*']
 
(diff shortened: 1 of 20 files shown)


garuppel:
Where do we want to start the go routines for the communication with the Evidence Store and Orchestrator (initXXX methods)?  Should we start that in the assessment.NewService()?

oxisto (approved):

PR #613, opened 2022-04-05
Update Rego policy implementation during Runtime

### LINKED ISSUE #495: Implement update of Rego policy during runtime

oxisto:
This PR introduces a major overhaul of the policy evaluation engine in the assessment service. 

- [x] Restructuring of Rego policy evaluation
- [x] Support for loading policies into a storage backend and retrieving them from the orchestrator in the assessment
- [x] Implementing RPC methods to update metric implementation
- [x] Broadcast metric events to assessment service
- [x] Moving `metrics` from a global variable to the orchestrator `Service`

As a by-product, this introduces several new classes and functions in the `testutils` package. If wanted, I can extract them into a separate PR:

- `NewInMemoryStorage` is a shortcut to quickly create an in-memory storage for unit tests without checking the error value (it will fail the test on error). It also includes handy varargs for storage initialisation
- `StorageWithError` is a struct that can be used to mock a `persistence.Storage` implementation that produces errors on storage operations
- `ListRecvStreamerOf` implements `grpc.ClientStreams` to mock the streaming of a list of messages and then gracefully EOFs the stream.

Fixes #495


persistence/gorm/gorm.go
```diff
@@ -87,15 +93,17 @@
...
+		&auth.User{},
+		&orchestrator.CloudService{},
+		&assessment.MetricImplementation{},
 	}
-	// Migrate CloudService
-	if err = g.db.AutoMigrate(&orchestrator.CloudService{}); err != nil {
-		err = fmt.Errorf("error during auto-migration: %w", err)
-		return
+
+	for _, typ := range types {
```
lebogg, line 102:
```suggestion
	for _, type := range types {
```

persistence/gorm/gorm.go
```diff
@@ -87,15 +93,17 @@
...
+		if err = g.db.AutoMigrate(typ); err != nil {
```
lebogg, line 103:
```suggestion
		if err = g.db.AutoMigrate(type); err != nil {
```

persistence/gorm/gorm.go
```diff
@@ -87,15 +93,17 @@
 	}
 
 	// After successful DB initialization, migrate the schema
-	// Migrate User
-	if err = g.db.AutoMigrate(&auth.User{}); err != nil {
-		err = fmt.Errorf("error during auto-migration: %w", err)
-		return
+	var types = []interface{}{
```
lebogg, line 96: We could also put everything directly into `AutoMigrate` (variadic function). I can do it in the certificate API PR though.

service/assessment/assessment.go
```diff
@@ -163,6 +175,17 @@
 		o(s)
 	}
 
+	// Default to an in-memory storage, if nothing was explicitly set
```
lebogg, line 178: I think WIthStorage option is missing

lebogg (approved):
Quick review. Seems good!

persistence/gorm/gorm.go
oxisto: `type` is a reserved keyword, that was the reasoning behind `typ`.

persistence/gorm/gorm.go
oxisto: See above. Rejected.

service/assessment/assessment.go
oxisto: Ah, this can be removed, this was a remnant from an old implementation attempt.

PR #615, opened 2022-04-06
Enabling Go 1.18

oxisto:
This PR makes Clouditor ready for 1.18, because we are awaiting some benefits from generics, especially in functions shared by all micro-services.

golangci was updated to its latest release, however some linters are deactivated by default in 1.18 now, because of missing support by the checkers (see Golanghttps://github.com/golangci/golangci-lint/issues/2649). But at least the basic checks should work.


oxisto:
Looks like deep source still compiles with an older Go version and thus it fails when a file contains generics. I have opened a service ticket there. We might consider dropping deep source until they resolve this issue.

lebogg (approved):
Finally :heart:

PR #627, opened 2022-04-14
Fix Azure bootLogging

### LINKED ISSUE #624: Azure Discovery: BootLogging.enabled is always set to false
**Describe the bug**
When running Clouditor and looking at the evidences, all VM instances have bootLogging set to false which isn't right.

**To Reproduce**
- Run clouditor
- list evidences
- Compare bootLogging of VMs with settings in the Azure Portal (BootDiagnostic in Azure).

**Expected behavior**
Set bootLogging to true when it is activated (which is "Enable with managed storage account (recommended)" by default)

### LINKED ISSUE #625: Azure Discovery: BootLogging.retentionPeriod is always set to "0"
**Describe the bug**
When running Clouditor and looking at the evidences, all VM instances have bootLogging.retentionPeriod set to "0"

**To Reproduce**
- Run clouditor
- list evidences

**Expected behavior**
We define retentionPeriod in time. Since Azure defines is in Size (saving logs until 1 GB), I would suggest to map size into time with some heuristic or we should add the option retentionSIZE and adjust the metrics accordingly?

**Additional context**
See https://docs.microsoft.com/en-us/azure/virtual-machines/boot-diagnostics#limitations

### LINKED ISSUE #626: Discovery: Clouditor panics after starting discovery
**Describe the bug**
Panic occurs some time after the Clouditor's discovery is started.

**To Reproduce**
1. Start Clouditor
2. 'cl service discovery start'
3. Wait or do 2 times 'cl service discovery query'

**Expected behavior**
No panics.

**Additional context**
I think the error happens because of missing mutexes.

garuppel:
This PR solves the following Issues: 
- Fix #626   nil pointer derefences in  service/discovery/azure/network.go
- Fix #624


service/discovery/azure/compute.go
```diff
@@ -163,19 +168,14 @@ func (d *azureComputeDiscovery) handleVirtualMachines(vm *compute.VirtualMachine
 				},
 			}},
 		BootLogging: &voc.BootLogging{Logging: &voc.Logging{
-			Enabled:         isBootDiagnosticEnabled(vm),
-			LoggingService:  []voc.ResourceID{voc.ResourceID(bootLogOutput(vm))},
+			Enabled:         isBootDiagnosticEnabled(vmExtended), //TODO(garuppel): Delete if ready
```
garuppel, line 171:
```suggestion
			Enabled:         isBootDiagnosticEnabled(vmExtended), 
```

oxisto:
It seems that the "profile" we use is quite old (2020-09-01). The latest one seems to be from 2021-12-01. 

The alternative seems to be to use the resource manager packages, it seems they are somehow magically always using the latest version. See https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute#section-readme or

oxisto (approved):

service/discovery/azure/network.go
```diff
@@ -247,22 +249,31 @@ func (d *azureNetworkDiscovery) publicIPAddressFromLoadBalancer(lb *network.Load
 	if lb.LoadBalancerPropertiesFormat != nil && lb.LoadBalancerPropertiesFormat.FrontendIPConfigurations != nil {
 		for _, publicIpProperties := range *lb.FrontendIPConfigurations {
 
+			if publicIpProperties.PublicIPAddress == nil {
+				continue
+			}
+
 			publicIpAddressName := getFrontendPublicIPAddressName(*publicIpProperties.PublicIPAddress.ID)
```
lebogg, line 256: Since you check if `PublicIPAddress` is nil in L252 and you can assume that the corresponding ID should be filled, it is maybe safe to use the de-referencing pointer ("*") here.
But I guess the `to.String` function from the Azure package should be used everywhere, right? To avoid nil pointer references?

service/discovery/azure/network.go
```diff
@@ -247,22 +249,31 @@ func (d *azureNetworkDiscovery) publicIPAddressFromLoadBalancer(lb *network.Load
...
 			if publicIpAddressName == "" {
 				continue
 			}
-			publicIPAddress, err := client.Get(context.Background(), getResourceGroupName(*publicIpProperties.ID), publicIpAddressName, "")
 
+			publicIPAddress, err := client.Get(context.Background(), getResourceGroupName(*publicIpProperties.ID), publicIpAddressName, "")
```
lebogg, line 261: see above

service/discovery/azure/network.go
```diff
@@ -247,22 +249,31 @@ func (d *azureNetworkDiscovery) publicIPAddressFromLoadBalancer(lb *network.Load
...
 				log.Infof("Error getting public IP address: %v", err)
 				continue
 			}
 
+			if publicIPAddress.IPAddress == nil {
+				log.Infof("Error getting public IP address: %v", err)
+				continue
+			}
+
 			publicIPAddresses = append(publicIPAddresses, *publicIPAddress.IPAddress)
```
lebogg, line 272: see above

lebogg (approved):

PR #637, opened 2022-04-20
Add medina metrics (Quick Fix)

lebogg:
Add Medina metrics via a `medina_metrics.json` file. Thus, it is added to the embedded filesystem.
In the future we want to modify the code, e.g. the Load function to also use another file system or similar.


garuppel (commented):
:+1:

garuppel (approved):
:+1:

PR #651, opened 2022-04-28
Hotfix for DB username and password using gorm

oxisto:
The GORM option was missing username, password and db name.


garuppel (approved):
OK

PR #670, opened 2022-05-06
Use generic streams function for orchestrator stream

### LINKED ISSUE #669: Use generic streams function for orchestrator stream in assessment service

garuppel:
Closes #669


service/assessment/assessment.go
```diff
@@ -107,7 +105,8 @@ type Service struct {
 	authorizer api.Authorizer
 
 	// grpcOpts contains additional gRPC options that will be appended to all gRPC dial calls. Useful for testing.
-	grpcOpts []grpc.DialOption
+	grpcOptsEvidenceStore []grpc.DialOption
```
oxisto, line 108: Is it really needed to split this?

service/assessment/assessment.go
```diff
@@ -596,3 +570,18 @@ func (svc *Service) recvEventsLoop() {
 		_ = svc.pe.HandleMetricEvent(event)
 	}
 }
+
+// setOrchestratorClient set the orchestrator client
+func (svc *Service) setOrchestratorClient() error {
```
oxisto, line 575: initOrchestratorClient?

service/assessment/assessment.go
garuppel: I've done that for a better testing coverage :) And I thought we might want different grpc DialOptions for different components.

garuppel: TODO: Combine WithAdditionalGRPCOptsXXX and WithXXXAddress.

service/assessment/assessment.go
```diff
@@ -412,103 +422,53 @@ func (s *Service) initEvidenceStoreStream(URL string, additionalOpts ...grpc.Dia
...
 // initOrchestratorStream initializes the stream to the Orchestrator
-func (s *Service) initOrchestratorStream() error {
+func (s *Service) initOrchestratorStream(URL string, additionalOpts ...grpc.DialOption) (stream orchestrator.Orchestrator_StoreAssessmentResultsClient, err error) {
 	log.Infof("Trying to establish a connection to orchestrator service @ %v", s.orchestratorAddress)
 
 	// Establish connection to orchestrator gRPC service
-	conn, err := grpc.Dial(s.orchestratorAddress,
-		api.DefaultGrpcDialOptions(s.orchestratorAddress, s, s.grpcOpts...)...,
-	)
+	err = s.initOrchestratorClient()
```
lebogg, line 438: Check first, if `s.orchestratorClient` is nil? Could be up already, e.g. through `Metrics(..)` function

service/assessment/assessment.go
```diff
@@ -412,103 +422,53 @@ func (s *Service) initEvidenceStoreStream(URL string, additionalOpts ...grpc.Dia
...
 	}
 
+	log.Infof("Connected to Evidence Store")
+
 	return
 }
 
 // initOrchestratorStream initializes the stream to the Orchestrator
-func (s *Service) initOrchestratorStream() error {
+func (s *Service) initOrchestratorStream(URL string, additionalOpts ...grpc.DialOption) (stream orchestrator.Orchestrator_StoreAssessmentResultsClient, err error) {
```
lebogg, line 434: `svc`

service/assessment/assessment.go
```diff
@@ -412,103 +422,53 @@ func (s *Service) initEvidenceStoreStream(URL string, additionalOpts ...grpc.Dia
...
+		return nil, fmt.Errorf("could not set up stream for listening to metric change events: %w", err)
 	}
 
 	go s.recvEventsLoop()
 
-	return nil
+	return
 }
 
 // MetricImplementation implements MetricsSource by retrieving the metric list from the orchestrator.
```
lebogg, line 461:
```suggestion
// Metrics implements MetricsSource by retrieving the metric list from the orchestrator.
```

service/assessment/assessment.go
oxisto: Another suggestion might be to introduce an (internal) type that also stores them together in the `Service` like 
```
type grpcTarget struct {
	target string
	opts   []grpc.DialOption
}
```

service/assessment/assessment.go
```diff
@@ -106,8 +109,8 @@ type Service struct {
 
 	authorizer api.Authorizer
 
-	// grpcOpts contains additional gRPC options that will be appended to all gRPC dial calls. Useful for testing.
-	grpcOpts []grpc.DialOption
+	// grpcTargets contains the target and DialOptions
+	grpcTargets map[string]grpcTarget
```
oxisto, line 113: Having this as a map introduces another layer of errors that needs to be caught (missing keys), etc. Personally, I would recommend just replacing the original `orchestratorAddress` and `evidenceStoreAddress` with the new internal struct.

service/assessment/assessment.go
```diff
@@ -124,17 +127,27 @@ const (
 // ServiceOption is a functional option type to configure the assessment service.
 type ServiceOption func(*Service)
 
-// WithEvidenceStoreAddress is an option to configure the evidence store gRPC address.
-func WithEvidenceStoreAddress(address string) ServiceOption {
+// WithEvidenceStore is an option to configure the evidence store gRPC address.
+func WithEvidenceStore(target string, opts ...grpc.DialOption) ServiceOption {
 	return func(s *Service) {
-		s.evidenceStoreAddress = address
+		opts := grpcTarget{
```
oxisto, line 133: Having this variable named `opts` is confusing.

service/assessment/assessment.go
garuppel: Ok, I thought so we could add easily more services.

oxisto: A good idea in theory but then we would need to also remove `orchestratorClient` and streams etc. so I think we can keep it as two separate "addresses" for now. Let's not break too many things at once ;)

service/assessment/assessment.go
```diff
@@ -78,14 +83,12 @@ type Service struct {
 
 	// evidenceStoreStream sends evidences to the Evidence Store
 	evidenceStoreStreams *api.StreamsOf[evidence.EvidenceStore_StoreEvidencesClient, *evidence.StoreEvidenceRequest]
-	evidenceStoreAddress string
+	evidenceStoreAddress *grpcTarget
```
oxisto, line 86: Is the pointer needed?

service/assessment/assessment.go
```diff
@@ -78,14 +83,12 @@ type Service struct {
...
 
 	// orchestratorStream sends assessment results to the Orchestrator
-	orchestratorStream  orchestrator.Orchestrator_StoreAssessmentResultsClient
+	orchestratorStreams *api.StreamsOf[orchestrator.Orchestrator_StoreAssessmentResultsClient, *orchestrator.StoreAssessmentResultRequest]
 	orchestratorClient  orchestrator.OrchestratorClient
-	orchestratorAddress string
-	// orchestratorChannel stores assessment results for the Orchestrator
-	orchestratorChannel chan *assessment.AssessmentResult
+	orchestratorAddress *grpcTarget
```
oxisto, line 91: Same here

api/orchestrator/orchestrator.pb.gw.go
```diff
@@ -1486,7 +1486,7 @@ func RegisterOrchestratorHandlerServer(ctx context.Context, mux *runtime.ServeMu
 		var stream runtime.ServerTransportStream
 		ctx = grpc.NewContextWithServerTransportStream(ctx, &stream)
 		inboundMarshaler, outboundMarshaler := runtime.MarshalerForRequest(mux, req)
-		rctx, err := runtime.AnnotateIncomingContext(ctx, mux, req, "/clouditor.Orchestrator/RegisterAssessmentTool", runtime.WithHTTPPathPattern("/v1/orchestrator/assessment_tools"))
```
oxisto, line 1489: Seems like old protobuf version ;)

oxisto (commented):
Last small changes

oxisto (approved):

PR #681, opened 2022-05-13
Implement hooks for updated cloud services

immqu:
This PR adds hooks for updates made to Cloud Service objects, so other components can be informed, e.g. when a cloud service's requirements change


oxisto (approved):
Overall we need to rework the hooks and events at some point to reduce duplicated code, but for now let's merge it.

PR #691, opened 2022-05-18
Persisting requirements in storage backend

oxisto:
This PR persist requirements in the storage backend. It also introduces the possibility to inject go struct tags (mainly for GORM) using https://github.com/srikrsna/protoc-gen-gotag. It is used to build an associating between a `Metric` and a `Requirement`.


lebogg (approved):
LGTM :) except another dependency. But what can we do :-O

PR #693, opened 2022-05-19
Add ResourceTypes to AssessmentResult

anatheka:


lebogg (approved):
LGTM

api/assessment/assessment.proto
```diff
@@ -120,9 +119,9 @@ message AssessmentResult {
   // Reference to the resource of the assessed evidence
   string resource_id = 7;
 
-  // Some comments on the reason for non-compliance
-  string non_compliance_comments = 8;
+  // Resource types
```
anatheka, line 122: I thought the newer resource type should have the next field number, in this case 9, for compatibility reasons.

anatheka (commented):
Only a minor thing.

api/assessment/assessment.proto
lebogg: I would guess it's not that important in the development phase

anatheka: I thought we get evidences from another evidence collecting tool and then it would be nice if he not has to change his code.

PR #700, opened 2022-05-22
Using `api.StreamsOf` in discovery service

oxisto:
This PR removes the custom streams implementation from the discovery service and migrates it to `api.StreamsOf`, similar to the other services.


service/discovery/discovery.go
```diff
@@ -100,9 +104,12 @@ const (
 type ServiceOption func(*Service)
 
 // WithAssessmentAddress is an option to configure the assessment service gRPC address.
-func WithAssessmentAddress(address string) ServiceOption {
+func WithAssessmentAddress(address string, opts ...grpc.DialOption) ServiceOption {
 	return func(s *Service) {
-		s.assessmentAddress = address
+		s.assessmentAddress = grpcTarget{
```
anatheka, line 109: In the assessment I have done an additional empty string check. I am not sure if it is really necessary: 
`if address == "" {
			address = DefaultAssessmentAddress
		}`

service/discovery/discovery.go
```diff
@@ -127,10 +134,13 @@ func WithProviders(providersList []string) ServiceOption {
 
 func NewService(opts ...ServiceOption) *Service {
 	s := &Service{
-		assessmentAddress: DefaultAssessmentAddress,
-		resources:         make(map[string]voc.IsCloudResource),
-		scheduler:         gocron.NewScheduler(time.UTC),
-		configurations:    make(map[discovery.Discoverer]*Configuration),
+		assessmentStreams: api.NewStreamsOf[assessment.Assessment_AssessEvidencesClient, *assessment.AssessEvidenceRequest](),
```
anatheka, line 137:
```suggestion
		assessmentStreams: api.NewStreamsOf[assessment.Assessment_AssessEvidencesClient, *assessment.AssessEvidenceRequest](log),
```

service/discovery/discovery_test.go
```diff
@@ -472,96 +445,6 @@ func TestStart(t *testing.T) {
 	}
 }
 
-func TestService_initAssessmentStream(t *testing.T) {
```
anatheka, line 475: I wouldn't delete that completely.

anatheka (approved):
Great! Only some minor things.

service/discovery/discovery.go
oxisto: Should not be necessary I think

PR #703, opened 2022-05-23
Fixed bug when populating applicable metric cache

oxisto:
This fixes a rather intricate bug that occurs, if the Orchestrator stores metrics, which he does not have an implementation for. This can occur if the metric is assessed outside of the Clouditor toolset. Still, the orchestrator needs to know about the metric, even though it does not know its implementation. In this case, if the policy engine tried to build the applicable metric cache, the metric cache was always empty, because if aborted if a metric implementation could not be found. Furthermore, it also ran into a deadlock because of a missing Unlock.

Now, the routine that builds the metric cache differentiates between a generic error, in which case the cache population will be aborted and the case where the metric implementation is not found (e.g., external). In the latter case, the population continues with the next metric.

Alternativly, we might choose to specifically return that the metric implementation is *external* instead of a NotFound code.


service/assessment/assessment_test.go
```diff
@@ -1038,8 +1039,11 @@ func TestHandleEvidence(t *testing.T) {
 				resourceId: "my-resource-id",
 			},
 			wantErr: func(t assert.TestingT, err error, i ...interface{}) bool {
-				assert.Contains(t, err.Error(), "could not get stream to evidence store")
-				return true
+				if !assert.NotEmpty(t, err) {
```
anatheka, line 1042: I would check for the specific error message, otherwise we can not be sure if it really fails at that point.

anatheka, line 1042:
```suggestion
				if assert.Empty(t, err) {
```

oxisto: This changes the semantic (and makes the test fail). We want to assert that the error is NOT empty. And only return false if it is actually empty.

anatheka (approved):

PR #725, opened 2022-06-08
Add Ordering in Persistence

lebogg:
This PR adds the possibility to order the results of ListXXX messages. This is achieved by:
- Adding parameters `orderBy` (column to be sorted) and `asc` (bool, true for ascending ordering and false for descending)
- Whitelisting values for `orderBy` to avoid DB injections


persistence/gorm/gorm.go
```diff
@@ -175,14 +175,22 @@ func (s *storage) Get(r any, conds ...any) (err error) {
 	return
 }
 
-func (s *storage) List(r any, offset int, limit int, conds ...any) error {
+func (s *storage) List(r any, orderBy string, asc bool, offset int, limit int, conds ...any) error {
 	var query = s.db
+	var orderDirection = "asc"
 
 	if limit != -1 {
```
oxisto, line 182: Do we need some kind of escaping here, since `orderBy` is taken from the outside and could potentially be used to inject something? Is gorm takeing care of this?

lebogg: No, a really good point. Im thinking of creating a `validation` function to check against a whitelist including e.g. `id`, `name` etc.

api/request.go
```diff
@@ -0,0 +1,44 @@
+package api
+
+import (
+	"errors"
+	"fmt"
+
+	"google.golang.org/protobuf/proto"
+	"k8s.io/utils/strings/slices"
```
oxisto, line 32: We should probably use https://pkg.go.dev/golang.org/x/exp/slices#Contains instead

api/request.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// ListRequest indicates a proto message being a request for ListXXX RPCs
+type ListRequest interface {
+	GetOrderBy() string
+	GetAsc() bool
+	proto.Message
+}
+
+func ValidateListReq(req ListRequest, responseType any) (err error) {
```
oxisto, line 49: Can we use type constraints instead of the responseType as parameter?

api/request.go
```diff
@@ -0,0 +1,44 @@
...
+	"clouditor.io/clouditor/internal/util"
+)
+
+var (
+	ErrInvalidColumnName = errors.New("column name is invalid")
+	ErrRequestIsNil      = errors.New("request is empty")
+)
+
+// ListRequest indicates a proto message being a request for ListXXX RPCs
+type ListRequest interface {
```
oxisto, line 43: I am torn whether we should have interfaces for the messages or whether we should use protoreflect to retrieve these values (we do it this way in the pagination request handling)

api/request.go
oxisto: Naming: ValidateListRequest?

internal/util/struct.go
```diff
@@ -0,0 +1,29 @@
...
+)
+
+var (
+	// ErrNoStruct indicates the passed argument is not a struct
+	ErrNoStruct = errors.New("no struct")
+)
+
+// GetFieldNames extracts all field names of struct T. Returns error if T is no struct.
+// TODO(lebogg): Only take exported fields
+func GetFieldNames[T any]() (fieldNames []string, err error) {
```
lebogg, line 15: Ideally, I would like to have sth. like `T struct` to avoid strings ect. at compile time already. Any idea? @oxisto

oxisto (changes requested):
Some small last changes necessary

api/request.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func ValidateListRequest[T proto.Message](req SortedListRequest) (err error) {
+	// req must be non-nil
+	if req == nil {
+		err = ErrRequestIsNil
+		return
+	}
+
+	// Avoid DB injections by whitelisting the valid orderBy statements
```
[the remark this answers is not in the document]
lebogg: When you mean to rename `whitelist`, I derived the name from the gorm doc:
"The general rule to avoid SQL injection is don’t trust user-submitted data, you can perform **whitelist** validation to test user input against an existing set of know"

service/paginate.go
```diff
@@ -75,12 +76,15 @@ func PaginateSlice[T any](req api.PaginatedRequest, values []T, opts PaginationO
 }
 
 // PaginateStorage is a helper function that helps to paginate records in persisted storage based on list requests. It
-// parses the necessary informaton out if a paginated request, e.g. the page token and the desired page size and returns
-// a sliced page as well as the next page token.
-func PaginateStorage[T any](req api.PaginatedRequest, storage persistence.Storage, opts PaginationOpts, conds ...interface{}) (page []T, npt string, err error) {
+// parses the necessary information out if a paginated request, e.g. the page token and the desired page size and
+// returns a sliced page as well as the next page token.
+// TODO(lebogg to Oxisto): We could add orderBy and asc to PaginatedRequest interface?
```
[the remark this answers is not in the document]
lebogg: Ok, I did it. But quite fast. Please double check if it is alright :)

openapi/assessment/openapi.yaml
```diff
@@ -161,30 +150,14 @@ components:
...
-                    $ref: '#/components/schemas/GoogleProtobufValue'
+                    $ref: '#/components/schemas/AnyJSONValue'
                 isDefault:
                     type: boolean
                     description: Whether this configuration is a default configuration
                 updatedAt:
                     type: string
                     description: The last time of update
-                    format: date-time
+                    format: RFC3339
```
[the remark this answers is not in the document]
lebogg: Yeeees :) , `protoc-gen-openapi` was not up to date

oxisto (approved):

Issue #563, opened 2022-03-10
Discovery: Azure ARM template discovery needs a mayor update


anatheka:
The ARM discovery needs mayor updates, I will do that in a separate PR.

_Originally posted by @garuppel in https://github.com/clouditor/clouditor/issues/539#issuecomment-1063844424_

The service/discovery/azure/arm_template* files contain the TODOs.

anatheka:
#739 removes the Azure ARM template discovery file and test file for now, as too many changes need to be made.

anatheka:
ARM discovery is removed.

PR #739, opened 2022-07-04
Update Azure SDK

anatheka:
- [x] List compute resources
- [x] List networking resources
- [x] List storage resources
- [x] Azure auth
- [x] Compute tests
- [x] Storage tests
- [x] Networking tests
- [ ] List ARM resources/tests -> Do we also update them now? anatheka: I would update them in a separate PR (Issue #563)
- [ ] If I understand it right, it is not possible to write a test for the NewAuthorizer() function, as the azidentity.NewDefaultAzureCredential() does not returns any error and we are not able access any fields. @oxisto could you check that, please?


internal/util/string.go
```diff
@@ -67,3 +67,13 @@ func marksNewWord(i int, input []rune) bool {
 
 	return false
 }
+
+// Deref dereferences string pointer
```
oxisto, line 71: The comment is not accurate (any more) since it derefes everything. Also the `string.go` should probably be renamed as well.

service/discovery/azure/arm_template.go.tobemigrated
```diff
@@ -34,7 +34,6 @@ import (
 	"clouditor.io/clouditor/api/discovery"
```
oxisto, line 34: We should probably just remove this file completely for now

service/discovery/azure/azure.go
```diff
@@ -123,50 +123,51 @@ func (a *azureDiscovery) authorize() (err error) {
...
-	// Create authorizer from file
-	authorizer, err = auth.NewAuthorizerFromFile(autorest_azure.PublicCloud.ResourceManagerEndpoint)
-	if err == nil {
-		log.Infof("Using authorizer from file")
-		return
+	return l
+}
+
+// safeTimestamp returns either the UNIX timestamp of the time t or 0 if it is nil
+func safeTimestamp(t *time.Time) int64 {
```
oxisto, line 155: Move this to the same package as `Deref`?

service/discovery/azure/azure.go
```diff
@@ -123,50 +123,51 @@ func (a *azureDiscovery) authorize() (err error) {
...
 
-	// Authorizer couldn't be created
-	log.Error(ErrCouldNotAuthenticate)
-	return nil, ErrCouldNotAuthenticate
+	return t.Unix()
 }
 
-func getResourceGroupName(id string) string {
-	return strings.Split(id, "/")[4]
+type Client interface {
```
oxisto, line 167: This probably need to be removed

service/discovery/azure/azure.go
```diff
@@ -123,50 +123,51 @@ func (a *azureDiscovery) authorize() (err error) {
...
+	armsubscription.SubscriptionsClient
+}
+
+type Resource interface {
+	armsubscription.Subscription
```
oxisto, line 172: This probably need to be removed

service/discovery/azure/arm_template.go.tobemigrated
```diff
@@ -85,7 +81,6 @@ func (d *azureARMTemplateDiscovery) discoverARMTemplate() ([]voc.IsCloudResource
 	)
 
 	client := resources.NewGroupsClient(to.String(d.sub.SubscriptionID))
```
oxisto, line 83: We should avoid using the `to` package, since its part of autorest and we want to get rid of that

go.mod
```diff
@@ -4,10 +4,7 @@ go 1.18
 
 // runtime dependencies
 require (
-	github.com/Azure/azure-sdk-for-go v65.0.0+incompatible
-	github.com/Azure/go-autorest/autorest v0.11.24
-	github.com/Azure/go-autorest/autorest/azure/auth v0.5.11
-	github.com/Azure/go-autorest/autorest/to v0.4.0
+	github.com/Azure/go-autorest/autorest v0.11.24 // indirect
```
oxisto, line 7: maybe run a go mod tidy on this? auto rest should be indirect and azidentity should be a direct dependency I guess

go.mod
```diff
@@ -151,4 +144,18 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
+require (
```
oxisto, line 147: is this new block here intentional? do we want to split our dependencies into further groups?

internal/util/string.go
anatheka: Thanks, I forgot to adjust the "string" after I expanded it.

anatheka: I couldn't think of a better filename, suggestions are welcome :wink:

go.mod
anatheka: If I run 'go mod tidy' it deletes everything or does nothing. I don't know why. Could you try that?

go.mod
anatheka: I think that was my IDE. I did not change anything.

oxisto (approved):

PR #754, opened 2022-07-15
Fixes all ports to be `uint16`

oxisto:
Some code was incorrectly using `int16` for ports (which truncates) it or `int` (which is ok but unncessary). This PR addresses most of it. 

This also includes a small ontology update from https://github.com/clouditor/cloud-property-graph/pull/45

### DIFF
--- a/internal/testutil/auth.go
+++ b/internal/testutil/auth.go
@@ -45,7 +45,7 @@ func StartAuthenticationServer() (srv *oauth2.AuthorizationServer, port uint16, 
 		_ = srv.Serve(nl)
 	}()
 
-	port = uint16(nl.Addr().(*net.TCPAddr).Port)
+	port = nl.Addr().(*net.TCPAddr).AddrPort().Port()
 
 	return srv, port, nil
 }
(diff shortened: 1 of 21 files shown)


anatheka (approved):

lebogg (commented):
Changes done LGTM, but I think you missed some:
- engine.go: : Default Ports in the constants section
- cloud_test.go (grpcPort)
- cli.go (grpcPort)
- auth.go (StartAuthenticationServer)

lebogg (changes requested):
see above

oxisto:
> Changes done LGTM, but I think you missed some:
> 
> * engine.go: : Default Ports in the constants section
> * cloud_test.go (grpcPort)
> * cli.go (grpcPort)
> * auth.go (StartAuthenticationServer)

I have converted the rest of the ports to `uint16` and also used the new `netip.AddrPort` which also uses `uint16` for ports for more consistentcy.

internal/testutil/auth.go
```diff
@@ -45,24 +45,24 @@ func StartAuthenticationServer() (srv *oauth2.AuthorizationServer, port int, err
 		_ = srv.Serve(nl)
 	}()
 
-	port = nl.Addr().(*net.TCPAddr).Port
+	port = uint16(nl.Addr().(*net.TCPAddr).Port)
```
lebogg, line 48: Maybe using `netip.AddrPort` here as well instead of cast?

lebogg (commented):
One change missing

lebogg (approved):

PR #767, opened 2022-07-28
Add discoverer component name 

anatheka:


service/discovery/azure/azure.go
```diff
@@ -42,6 +42,12 @@ import (
 	"clouditor.io/clouditor/internal/util"
 )
 
+const (
+	StorageComponent = "storage"
+	ComputeComponent = "compute"
+	NetworkComponent = "network"
+)
```
lebogg, line 49: Maybe we move these one level up s.t. all discoveries use the same constants, e.g. also AWS. What do you think?

lebogg (commented):
LGTM. Only one minor idea.

service/discovery/azure/azure.go
anatheka: You are absolutely right. I will change that.

anatheka: If we change this we run into 'import cycle' errors. For now, I'll leave it like that.

lebogg (approved):

PR #773, opened 2022-08-09
Provide information about failed authentication requests in debug log 

oxisto:
We intentionally do not provide any detailed information about why a request failed to the caller because of security reasons. However, this makes debugging authentication issues extremely hard. This PR adds an additional (debug) log why a request failed. 

@lebogg Do you think debug is ok or should we use trace?

### DIFF
--- a/persistence/gorm/gorm.go
+++ b/persistence/gorm/gorm.go
@@ -26,11 +26,8 @@
 package gorm
 
 import (
-	"context"
 	"errors"
 	"fmt"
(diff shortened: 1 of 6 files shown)


lebogg (commented):
Debug is good. But I think these changes are already done in #774

oxisto:
> Debug is good. But I think these changes are already done in #774

Ah, this PR was supposed to be merged first, you were too fast :P

PR #780, opened 2022-08-17
Excluding reflection API from authentication

oxisto:
This PR excludes the gRPC reflection API from authentication because it does only contain public information anyway. This makes exploring the service with `grpcurl` less of a hassle.


lebogg (approved):
LGTM

PR #787, opened 2022-08-25
Introducing the concept of a `Catalog`

immqu:
This PR is a *major* overhaul of the persistence layer with respect to compliance controls and requirements. `Requirements` were refactored into `Controls`, which also can have sub-controls now. It introduces the concept of a `Catalog`, which holds a `Category` which in turn holds a `Control`. During this change a lot of API paths have changed and are mostly categorised in a hierarchical structure now.

The usage of `service_id` as a foreign key for a `CloudService` was also harmonised into `cloud_service_id`.


api/orchestrator/orchestrator.proto
```diff
@@ -423,31 +460,70 @@ message CloudService {
...
+// could be a Control in a certification catalog.
+message Control {
   string id = 1;
 
   string name = 2;
 
   string description = 3;
 
+  repeated Control controls = 4
+      [ (tagger.tags) = "gorm:\"serializer:json\"" ];
```
lebogg, line 485: TODO: Discuss internally how to do that.

Some thoughts:
- I imagine with json, we store too much directly here? When retrieving a control, it will act like preloading everything. Maybe we even want it that way?
- Using a field `string control_id` will maybe implicitly create the desired nested controls (leaving it empty means that is the root control)
- many2many tag or similar to do the nesting? Dunno if that is working here

api/orchestrator/orchestrator.proto
```diff
@@ -136,13 +136,13 @@ service Orchestrator {
     };
   }
 
-  rpc ListRequirements(ListRequirementsRequest)
-      returns (ListRequirementsResponse) {
+  // List all metrics that are applicable to a certain control
+  rpc ListMetricsForControl(ListMetricsForControlRequest) returns (ListMetricsResponse) {
```
oxisto, line 140: Why do we need this? Isn't this returned by the database?

api/orchestrator/orchestrator.proto
oxisto: This should be a simple belongs to I think https://gorm.io/docs/belongs_to.html

api/orchestrator/orchestrator.proto
```diff
@@ -423,31 +460,70 @@ message CloudService {
...
+  Catalog catalog = 3;
 
-  // category can be used to categorize requirements, e.g., according to a
-  // specific security catalog or something that groups requirements together,
-  // e.g., a security control.
-  string category = 5;
+  string assurance_level = 4;
+}
+
+message CreateTargetOfEvaluationRequest { TargetOfEvaluation targetOfEvaluation = 1; }
```
oxisto, line 507: snakecase. could potentially also just be toe_id

api/orchestrator/orchestrator.proto
```diff
@@ -423,31 +460,70 @@ message CloudService {
...
+
+message RemoveTargetOfEvaluationRequest { string targetOfEvaluation_id = 1; }
```
oxisto, line 509: snakecase

api/orchestrator/orchestrator.proto
```diff
@@ -423,31 +460,70 @@ message CloudService {
...
+
+message GetTargetOfEvaluationRequest { string targetOfEvaluation_id = 1; }
```
oxisto, line 511: snakecase

oxisto:
`Catalog` also needs to be added to the registered types in persistance/gorm/gorm.go

oxisto:
Somehow, I could not start an engine with a postgres db now:
```
Error: could not create storage: error during auto-migration: ERROR: relation "catalogs" does not exist (SQLSTATE 42P01)
```

oxisto:
> Somehow, I could not start an engine with a postgres db now:
> 
> ```
> Error: could not create storage: error during auto-migration: ERROR: relation "catalogs" does not exist (SQLSTATE 42P01)
> ```

Ah. you need to migrate the catalog type first, then control.
```diff
diff --git a/persistence/gorm/gorm.go b/persistence/gorm/gorm.go
index e0015741..72b726aa 100644
--- a/persistence/gorm/gorm.go
+++ b/persistence/gorm/gorm.go
@@ -66,8 +66,8 @@ var DefaultTypes = []any{
        &assessment.Metric{},
        &orchestrator.Certificate{},
        &orchestrator.State{},
-       &orchestrator.Control{},
        &orchestrator.Catalog{},
+       &orchestrator.Control{},
 }

 // StorageOption is a functional option type to configure the GORM storage. E.g. WithInMemory or WithPostgres
```

api/orchestrator/orchestrator.proto
lebogg: I did it this way now (= option 2 w/ `string control_id`). But now we have the following (see also the failing test):
Imagine having a catalog with one control (Cont1234). And this control again has a sub control (Cont1234.1). Which again has a sub control (Cont1234.1.1).
When we now fetch the catalog, the list of controls includes all three controls - although we should only get the one main control. That's the case since they all refer to the same catalog with the corresponding id.

Some options:
- For sub controls we leave the `catalog_id` field empty
- We define a message sub_control that has no field `catalog_id` and use them for sub controls
- (We leave it like that since we get all controls this way. But then the client has to clean up the control mess)

Nevertheless, I think we need some API point for fetching controls, e.g. list/get control. Otherwise we have no chance to get any sub controls due to our restriction to only preload one level deep.

api/orchestrator/orchestrator.proto
immqu: When we load a catalog and its controls from the UI, we still need to load the metrics. So either we list all metrics and filter them in the UI, or we offer the filter directly in the API, what do you think?
Can we also add the option to specify a control in the ListMetrics rpc?

lebogg: "Can we also add the option to specify a control in the ListMetrics rpc?" Thats what I basically did just know ^^. If control id is set I call your implemented function (I just made it unexported)

api/orchestrator/orchestrator.proto
oxisto: > I did it this way now (= option 2 w/ `string control_id`). But now we have the following (see also the failing test): Imagine having a catalog with one control (Cont1234). And this control again has a sub control (Cont1234.1). Which again has a sub control (Cont1234.1.1). When we now fetch the catalog, the list of controls includes all three controls - although we should only get the one main control. That's the case since they all refer to the same catalog with the corresponding id.
> 
> Some options:
> 
> * For sub controls we leave the `catalog_id` field empty
> * We define a message sub_control that has no field `catalog_id` and use them for sub controls
> * (We leave it like that since we get all controls this way. But then the client has to clean up the control mess)
> 
> Nevertheless, I think we need some API point for fetching controls, e.g. list/get control. Otherwise we have no chance to get any sub controls due to our restriction to only preload one level deep.

Only the top most control should have the `catalog_id` field set. In fact, you will probably not need to set the field itself, gorm will do it for you.

oxisto: Additionally, we can also think about changing the `control_id` field to `parent_id` or `parent_control_id` to make it more readable that this only applies to sub-controls.

api/orchestrator/orchestrator.proto
oxisto: > When we load a catalog and its controls from the UI, we still need to load the metrics. So either we list all metrics and filter them in the UI, or we offer the filter directly in the API, what do you think?
> Can we also add the option to specify a control in the ListMetrics rpc?

Technically, this is not really needed from an UI point of view. In the UI we load all metrics once and store them in a global svelte store (as this information rarely changes), so when displaying a catalog we can filter the necessary metrics for the catalog coming out of the store, without triggering a new call to the backend.

oxisto:
There still seems to be a problem when running this with an actual database.

```
ERRO[0000] Error while saving catalog ERROR: insert or update on table "controls" violates foreign key constraint "fk_controls_controls" (SQLSTATE 23503)  component=orchestrator
```

immqu:
> There still seems to be a problem when running this with an actual database.
> 
> ```
> ERRO[0000] Error while saving catalog ERROR: insert or update on table "controls" violates foreign key constraint "fk_controls_controls" (SQLSTATE 23503)  component=orchestrator
> ```

This should be fixed by using the FullSaveAssociations flag in save/update operations: https://github.com/clouditor/clouditor/blob/catalogs/persistence/gorm/gorm.go#L202-L209

api/orchestrator/orchestrator.proto
```diff
@@ -434,34 +526,119 @@ message CloudService {
...
+
+// Control represents a certain Control that needs to be fulfilled. It could be
+// a Control in a certification catalog. It follows the OSCAL model. A
+// requirement in the EUCS terminology, e.g., is represented as the lowest
+// sub-control.
+message Control {
+  // A short name, e.g. OPS-01
+  // TODO(oxisto): name this "id"? or is it too confusing that this id is not
+  // unique? however, it is usually referred to as a control ID
+  string short_name = 1 [ (tagger.tags) = "gorm:\"primaryKey\"" ];
```
oxisto, line 569: Should we just name this "ID" as its commonly referred to in standards and have a comment saying that ID itself is not unique but just together with the other primary keys?

api/orchestrator/orchestrator.proto
```diff
@@ -434,34 +526,119 @@ message CloudService {
...
+  // Reference to the parent category this control belongs to.
+
+  optional string parent_control_short_name = 8;
+  optional string parent_control_category_name = 10;
+  optional string parent_control_category_catalog_id = 9;
+}
+
+// A Target of Evaluation binds a cloud service to a catalog, so the service is
+// evaluated regarding this catalog's controls
+message TargetOfEvaluation {
```
oxisto, line 602: This could probably be modelled as a customised join table. https://gorm.io/docs/many_to_many.html#Customize-JoinTable

Should we extract this into a separate PR to think about this more careful? Maybe once our Gorm expert is back from vacation.

service/orchestrator/orchestrator.go
```diff
@@ -108,12 +110,6 @@ func WithExternalMetrics(f func() ([]*assessment.Metric, error)) ServiceOption {
 	}
 }
```
oxisto, line 112: For MEDINA we need a way to fill the catalogs from the outside, so we need something similar to `WithRequirements`.

service/orchestrator/orchestrator.go
```diff
@@ -403,3 +400,42 @@ func (svc *Service) RemoveCertificate(_ context.Context, req *orchestrator.Remov
 
 	return &emptypb.Empty{}, nil
 }
+
+// LoadControls loads controls definitions from a JSON file.
```
oxisto, line 404:
```suggestion
// loadCatalogs loads controls definitions from a JSON file.
```

service/orchestrator/orchestrator_test.go
```diff
@@ -307,7 +307,7 @@ func TestStoreAssessmentResults(t *testing.T) {
 			},
 		},
 		{
-			name: "Error in stream to server - Recv()-err",
+			name: "Error in stream to server - Recv()-wantErr",
```
oxisto, line 310: probably c/p mistake

service/orchestrator/orchestrator_test.go
```diff
@@ -317,7 +317,7 @@ func TestStoreAssessmentResults(t *testing.T) {
 			wantErrMessage: "rpc error: code = Unknown desc = cannot receive stream request",
 		},
 		{
-			name: "Error in stream to client - Send()-err",
+			name: "Error in stream to client - Send()-wantErr",
```
oxisto, line 320: probably c/p mistake

service/orchestrator/orchestrator_test.go
```diff
@@ -474,7 +474,7 @@ type mockStreamerWithSendErr struct {
 }
 
 func (mockStreamerWithSendErr) Send(*orchestrator.StoreAssessmentResultResponse) error {
-	return errors.New("Send()-err")
+	return errors.New("Send()-wantErr")
```
oxisto, line 477: probably c/p mistake

api/orchestrator/orchestrator.proto
oxisto: I have prepared this now and it seems to work, at last for `CreateTargetOfEvaluation`. The remaining methods need to implemented and tested.

oxisto:
Just a marker for ourselves. We should merge https://github.com/clouditor/clouditor/pull/801 first because this PR depends on changes of that PR.

api/orchestrator/orchestrator.go
```diff
@@ -27,46 +27,54 @@ package orchestrator
 
 import (
 	"context"
-	"database/sql/driver"
 	"errors"
-	"strings"
 )
 
+// keep this file for later target of evaluation
```
oxisto, line 33: Cleanup this file once we are done.

cli/commands/cloud/cloud.go
```diff
@@ -222,19 +219,18 @@ func NewUpdateCloudServiceCommand() *cobra.Command {
 	cmd.PersistentFlags().String("id", "", "the cloud service id to update")
 	cmd.PersistentFlags().StringP("name", "n", "", "the name of the service")
 	cmd.PersistentFlags().StringP("description", "d", "", "an optional description")
-	cmd.PersistentFlags().StringSliceP("requirement-ids", "r", nil, "a list of requirements this cloud service should satisfy")
+	cmd.PersistentFlags().StringSliceP("control-ids", "r", nil, "a list of controls this cloud service should satisfy")
```
oxisto, line 222: This wil not work any more, we should remove this parameter.

persistence/gorm/gorm.go
```diff
@@ -61,13 +62,15 @@ type storage struct {
...
 	&orchestrator.CloudService{},
 	&assessment.MetricImplementation{},
 	&assessment.Metric{},
 	&orchestrator.Certificate{},
 	&orchestrator.State{},
-	&orchestrator.Requirement{},
+	&orchestrator.Catalog{},
+	&orchestrator.Category{},
+	&orchestrator.Control{},
+	&orchestrator.TargetOfEvaluation{},
```
oxisto, line 73: since TargetOfEvaluation is a join table, it is not necessary to auto-migrate it. Not sure if there is any harm in doing it.

persistence/inmemory/inmemory.go
```diff
@@ -14,5 +14,5 @@ func NewStorage() (persistence.Storage, error) {
 	return gorm.NewStorage(
 		gorm.WithInMemory(),
 		gorm.WithMaxOpenConns(1),
-		gorm.WithLogger(logger.Default.LogMode(logger.Silent)))
+		gorm.WithLogger(logger.Default.LogMode(logger.Info)))
```
oxisto, line 17: Set back to silent once we are done.

service/orchestrator/toe.go
```diff
@@ -0,0 +1,130 @@
...
+func (svc *Service) GetTargetOfEvaluation(_ context.Context, req *orchestrator.GetTargetOfEvaluationRequest) (response *orchestrator.TargetOfEvaluation, err error) {
+	if req == nil {
+		return nil, status.Errorf(codes.InvalidArgument, api.ErrRequestIsNil.Error())
+	}
+	if req.CloudServiceId == "" || req.CatalogId == "" {
+		return nil, status.Errorf(codes.NotFound, orchestrator.ErrToEIDIsMissing.Error())
+	}
+
+	response = new(orchestrator.TargetOfEvaluation)
+	err = svc.storage.Get(response, "cloud_service_id = ? AND catalog_id = ?", req.CloudServiceId, req.CatalogId)
```
oxisto, line 63:
```suggestion
	err = svc.storage.Get(response, gorm.WithoutPreload(), "cloud_service_id = ? AND catalog_id = ?", req.CloudServiceId, req.CatalogId)
```

service/orchestrator/toe.go
```diff
@@ -0,0 +1,130 @@
...
+		err = fmt.Errorf("invalid request: %w", err)
+		log.Error(err)
+		err = status.Errorf(codes.InvalidArgument, "%v", err)
+		return
+	}
+
+	res = new(orchestrator.ListTargetsOfEvaluationResponse)
+
+	res.Toes, res.NextPageToken, err = service.PaginateStorage[*orchestrator.TargetOfEvaluation](req, svc.storage,
+		service.DefaultPaginationOpts)
```
oxisto, line 85:
```suggestion
		service.DefaultPaginationOpts,  gorm.WithoutPreload())
```

service/orchestrator/toe_test.go
```diff
@@ -0,0 +1,362 @@
...
+	"clouditor.io/clouditor/persistence/gorm"
+	"github.com/stretchr/testify/assert"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+var AssuranceLevelHigh = "high"
+var AssuranceLevelMedium = "medium"
+
+func Test_CreateTargetOfEvaluation(t *testing.T) {
```
oxisto, line 48:
```suggestion
func TestService_CreateTargetOfEvaluation(t *testing.T) {
```

service/orchestrator/toe_test.go
```diff
@@ -0,0 +1,362 @@
...
+			}
+
+			if tt.want != nil {
+				tt.want(t, gotRes, svc)
+			}
+		})
+	}
+}
+
+func Test_GetTargetOfEvaluation(t *testing.T) {
```
oxisto, line 134:
```suggestion
func TestService_GetTargetOfEvaluation(t *testing.T) {
```

Tests should follow the syntax `TestMyGlobalFunc` if it is a global fund or `TestMyStruct_MyFunc` if it is a method function of `MyStruct`.

service/orchestrator/toe_test.go
```diff
@@ -0,0 +1,362 @@
...
+			// Validate the error via the ErrorAssertionFunc function
+			tt.wantErr(t, err)
+
+			// Validate the response via the ValueAssertionFunc function
+			tt.wantResponse(t, res)
+		})
+	}
+}
+
+func Test_ListTargetsOfEvaluation(t *testing.T) {
```
oxisto, line 237:
```suggestion
func TestService_ListTargetsOfEvaluation(t *testing.T) {
```

service/orchestrator/toe_test.go
```diff
@@ -0,0 +1,362 @@
...
+	assert.Equal(t, 1, len(listTargetsOfEvaluationResponse.Toes))
+
+	// 3rd case: Invalid request
+	_, err = orchestratorService.ListTargetsOfEvaluation(context.Background(),
+		&orchestrator.ListTargetsOfEvaluationRequest{OrderBy: "not a field"})
+	assert.Equal(t, codes.InvalidArgument, status.Code(err))
+	assert.Contains(t, err.Error(), api.ErrInvalidColumnName.Error())
+}
+
+func Test_UpdateTargetOfEvaluation(t *testing.T) {
```
oxisto, line 267:
```suggestion
func TestService_UpdateTargetOfEvaluation(t *testing.T) {
```

service/orchestrator/toe_test.go
```diff
@@ -0,0 +1,362 @@
...
+			CatalogId:      orchestratortest.MockCatalogID,
+			AssuranceLevel: &AssuranceLevelMedium,
+		},
+	})
+	assert.NoError(t, err)
+	assert.NotNil(t, toe)
+	assert.Equal(t, "new assurance level", toe.AssuranceLevel)
+}
+
+func Test_RemoveTargetOfEvaluation(t *testing.T) {
```
oxisto, line 315:
```suggestion
func TestService_RemoveTargetOfEvaluation(t *testing.T) {
```

service/orchestrator/catalogs.go
```diff
@@ -0,0 +1,171 @@
...
+		err = fmt.Errorf("invalid request: %w", err)
+		log.Error(err)
+		err = status.Errorf(codes.InvalidArgument, "%v", err)
+		return
+	}
+
+	res = new(orchestrator.ListControlsResponse)
+
+	// If the category name is set (additional binding), forward it as a condition to the pagination method
+	if req.CategoryName != "" {
```
oxisto, line 159: One probably should specify the category name AND the catalog ID and if both are set, then follow the `if`, since category names are only unique within a catalog.

service/orchestrator/catalogs.json
```diff
@@ -0,0 +1,31 @@
+[
```
oxisto, line 1: More meaningful catalog once we are done. Maybe something similar to the old requirements.json.

service/orchestrator/toe_test.go
```diff
@@ -0,0 +1,362 @@
...
+	// 2nd case: ErrRecordNotFound
+	_, err = orchestratorService.RemoveTargetOfEvaluation(context.Background(), &orchestrator.RemoveTargetOfEvaluationRequest{
+		CloudServiceId: "0000",
+		CatalogId:      "0000",
+	})
+	assert.Error(t, err)
+	assert.Equal(t, status.Code(err), codes.NotFound)
+
+	// 3rd case: Record removed successfully
+	err = orchestratorService.storage.Create(&orchestrator.TargetOfEvaluation{
```
oxisto, line 344: You need to create a catalog and cloud service first, otherwise the foreign key constraints will (correct) fail.

service/orchestrator/toe_test.go
immqu: Are we naming Tests always "TestService" when the tested method is executed _on the service_? Then we should rename a few other test methods as well

oxisto: Tests should follow the syntax `TestMyGlobalFunc` if it is a global func that is being tested (`GlobalFunc`) or `TestMyStruct_MyFunc` if it is a method function of `MyStruct`.

This is defined in some style guide, I don't have a reference for it. Maybe @lebogg has it.

service/orchestrator/toe.go
```diff
@@ -81,9 +80,7 @@ func (svc *Service) ListTargetsOfEvaluation(_ context.Context, req *orchestrator
 	}
 
 	res = new(orchestrator.ListTargetsOfEvaluationResponse)
-
-	res.Toes, res.NextPageToken, err = service.PaginateStorage[*orchestrator.TargetOfEvaluation](req, svc.storage,
-		service.DefaultPaginationOpts)
+	err = svc.storage.List(&res.Toes, "", false, 0, -1, gorm.WithoutPreload())
```
oxisto, line 83: did you try `res.Toes, res.NextPageToken, err = service.PaginateStorage[*orchestrator.TargetOfEvaluation](req, svc.storage, service.DefaultPaginationOpts, gorm.WithoutPreload())`? I think this should work.

service/orchestrator/orchestrator.go
immqu: You mean we get the catalog(s) from the central MEDINA respository, and need to store them in the Orchestrator? I was assuming that we simply pre-define the EUCS in Clouditor. Alternatively, we can simply store it via the CreateCatalog API, right?

oxisto: > You mean we get the catalog(s) from the central MEDINA respository, and need to store them in the Orchestrator? I was assuming that we simply pre-define the EUCS in Clouditor. Alternatively, we can simply store it via the CreateCatalog API, right?

I think we need to retrieve it from the catalog component, but using the `CreateCatalog` API sounds like a reasonable option.

api/orchestrator/orchestrator.proto
immqu: Ok, let's use "id", that way we are also closer to the OSCAL terminology

service/orchestrator/catalogs.json
immqu: I have added a few more fields and more descriptive names, but this is not what you mean I assume?

oxisto: In the old file (see https://github.com/clouditor/clouditor/blob/main/service/orchestrator/requirements.json) we had two exemplary requirements that were connected to real metrics. So I would assume that we would either create our "own" catalog with more or less "fake" controls or maybe choose an appropriate EUCS control that is at least connected to a real metric such as TransportEncryptionEnabled.

service/orchestrator/catalogs.go
```diff
@@ -169,3 +170,55 @@ func (srv *Service) ListControls(ctx context.Context, req *orchestrator.ListCont
 	}
 	return
 }
+
+// LoadCatalogs loads catalog definitions from a JSON file.
```
oxisto, line 174:
```suggestion
// loadCatalogs loads catalog definitions from a JSON file.
```

oxisto:
I fixed the remaining tests but for some reasons, it will still not work with a real Postgres DB:
```
ERRO[0000] Error while saving metrics: ERROR: ON CONFLICT DO UPDATE requires inference specification or constraint name (SQLSTATE 42601)  component=orchestrator
ERRO[0000] Could not load embedded metrics. Will continue with empty metric list: ERROR: ON CONFLICT DO UPDATE requires inference specification or constraint name (SQLSTATE 42601)  component=orchestrator
```

Trying to investigate...

oxisto:
> I fixed the remaining tests but for some reasons, it will still not work with a real Postgres DB:
> 
> ```
> ERRO[0000] Error while saving metrics: ERROR: ON CONFLICT DO UPDATE requires inference specification or constraint name (SQLSTATE 42601)  component=orchestrator
> ERRO[0000] Could not load embedded metrics. Will continue with empty metric list: ERROR: ON CONFLICT DO UPDATE requires inference specification or constraint name (SQLSTATE 42601)  component=orchestrator
> ```
> 
> Trying to investigate...

Should work now.

oxisto (approved):
LGTM

PR #790, opened 2022-08-28
Clearing cached configuration from assessment when metric has changed

oxisto:
We had a rather unfortunate bug in the assessment service, which did clear the metric configuration cache when a metric has changed. This way, the old metric configuration was still received from the cache in the policy evaluation.

Furthermore, we needed to make the policy evaluation aware of the cloud service it is handling the evaluation for, since different cloud services can have different metric configurations. It was still hard-coded to the default service.


lebogg (approved):
LGTM

PR #801, opened 2022-09-12
Various Gorm persistence improvements

oxisto:
This PR adds several improvements to the Gorm model and interface, resulting in a better persistence experience:

- Added an association between `Metric` and `MetricImplementation` to enforce a foreign key `metric_id`.
- For various reasons, I needed to change most / all `service_id` fields to `cloud_service_id`.
- The message `MetricConfiguration` is now correctly modelled as a custom join table between `CloudService` and `Metric`. Due to the way join tables work, `CloudService` now does not have a list of metric configurations any more, but only metrics. However, the preferred way of accessing metric configurations is anyway to use `ListMetricConfigurations` to get a list of all default and changed metric configurations for a particular service.
- Correctly applying the varargs arguments to the underlying Gorm function calls now. We now basically distinguish between three cases:
  - an empty conditions list means no db.Where function is called
  - one condition specified means that it is takes as the query parameter. This will query for the specified primary key
  - otherwise, the first condition will be taken as the query parameter and all others will be taken as additional args.


persistence/persistence.go
```diff
@@ -31,6 +31,7 @@ import (
 
 var (
 	ErrRecordNotFound  = errors.New("record not in the database")
+	ErrConstaintFailed = errors.New("constraint failed")
```
anatheka, line 34:
```suggestion
	ErrConstraintFailed = errors.New("constraint failed")
```

service/orchestrator/metrics.go
```diff
@@ -333,49 +326,19 @@ func (svc *Service) GetMetricConfiguration(_ context.Context, req *orchestrator.
 
 // UpdateMetricConfiguration updates the configuration for a metric, specified by the identifier in req.MetricId.
 func (svc *Service) UpdateMetricConfiguration(_ context.Context, req *orchestrator.UpdateMetricConfigurationRequest) (res *assessment.MetricConfiguration, err error) {
-	var (
-		metric assessment.Metric
-		cld    orchestrator.CloudService
-	)
+	// TODO(oxisto): Validate the request
```
anatheka, line 329: Request validation still missing.

service/orchestrator/metrics_test.go
```diff
@@ -834,16 +833,17 @@ func TestService_UpdateMetricConfiguration(t *testing.T) {
 		name    string
 		fields  fields
 		args    args
-		wantRes *assessment.MetricConfiguration
+		want    assert.ValueAssertionFunc
 		wantErr bool
```
anatheka, line 837: A general point: I would always check the explicit error message. The error can be thrown for all sorts of reasons.

api/orchestrator/orchestrator.proto
```diff
@@ -438,10 +438,8 @@ message CloudService {
 
   Requirements requirements = 4;
 
-  repeated MetricConfiguration metric_configurations = 5
-      [ (tagger.tags) = "gorm:\"many2many:cloud_service_metric_configurations;"
-                        "foreignKey:id;"
-                        "references:service_id,metric_id\"" ];
+  repeated Metric configured_metrics = 5
```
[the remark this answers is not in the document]
oxisto: In theory yes, but currently we are anyway not really backwards compatible so I think its ok. We should probably stick to some fixed numbers at some point

service/orchestrator/metrics.go
oxisto: I would solve this in a separate PR as we finally need something for all methods to properly validate them.

anatheka (approved):
From my side it's ok :+1:

PR #815, opened 2022-10-14
Update GetMetricConfiguration description

anatheka:
- [ ] UpdateMetricConfiguration
    - [x] Add request validation for  metricId and cloudServiceId, Add tests!
    - [ ] isDefault is not updated, if not available in request -> set to false
    - [ ] grpc does not returns some fields (e.g., isDefault, Timestamp)
    - [ ] updatedAt is always set to 1970-01-01T00:00:00Z
- [ ] GetMetricConfiguration: MetricId and cloudServiceId not available after initialization (curl call)


anatheka:
The following should be checked in an separate PR/Issue.

UpdateMetricConfiguration
- [ ] isDefault is not updated, if not available in request -> set to false
- [ ] grpc does not returns some fields (e.g., isDefault, Timestamp)
- [ ] updatedAt is always set to 1970-01-01T00:00:00Z

GetMetricConfiguration: 
- [ ] MetricId and cloudServiceId not available after initialization (curl call)

api/assessment/metric.go
```diff
@@ -142,6 +167,29 @@ func (m *Metric) Validate(opts ...MetricValidationOption) (err error) {
 	return nil
 }
 
+// Validate validates the metric configuration
+func (c *MetricConfiguration) Validate() (err error) {
+
+	// isDefault does not need to be checked.
+	// UpdatedAt does not need to be checked.
+	// MetricId does not need to be checked.
```
oxisto, line 175: why?

api/assessment/metric.go
```diff
@@ -142,6 +167,29 @@ func (m *Metric) Validate(opts ...MetricValidationOption) (err error) {
...
+	// CloudServiceId does not need to be checked.
```
oxisto, line 176: why?

api/orchestrator/orchestrator.go
```diff
@@ -28,21 +28,59 @@ package orchestrator
...
+	ErrCloudServiceIDIsMissing = errors.New("cloud service id is missing")
+	ErrMetricIDIsMissing       = errors.New("metric id is missing")
+	ErrCloudServiceIDIsInvalid = errors.New("cloud service id is invalid")
+	ErrRequestCloudServiceID   = errors.New("error in cloud service id")
 )
 
+// Validate validates the UpdateMetricConfigurationRequest
+func (req *UpdateMetricConfigurationRequest) Validate() error {
+	// Check cloud service ID
+	err := checkServiceID(req.CloudServiceId)
```
oxisto, line 56: don't we have similar functions in the previous validate methods somehow?

api/orchestrator/orchestrator.proto
```diff
@@ -192,7 +192,7 @@ service Orchestrator {
   }
 
   // Retrieves a metric configuration (target value and operator) for a specific
-  // service and metric ID
+  // service and metric ID. If no service ID is specified the default metric configuration is retrieved.
```
oxisto, line 195: That is not true. If It is so, then it is a bug. This call always needs a cloud service ID. If no specific metric configuration exists for the service, then the default one is returned, otherwise the "overwritten" one.

api/assessment/metric.go
```diff
@@ -142,8 +170,53 @@ func (m *Metric) Validate(opts ...MetricValidationOption) (err error) {
...
+		return ErrMetricConfigurationTargetValueMissing
+	}
+
+	if c.MetricId == "" {
+		return ErrMetricIdMissing
+	}
+
+	err = CheckCloudServiceID(c.CloudServiceId)
+	if err != nil {
+		return fmt.Errorf("%w", err)
```
oxisto, line 197: We can just return the error here without any wrapping I think

api/orchestrator/orchestrator.go
```diff
@@ -28,21 +28,57 @@ package orchestrator
...
+	ErrToEIDIsMissing        = errors.New("toe ID is empty")
+	ErrRequestCloudServiceID = errors.New("error in cloud service id")
 )
 
+// Validate validates the UpdateMetricConfigurationRequest
+func (req *UpdateMetricConfigurationRequest) Validate() error {
+	// Check cloud service ID
+	err := assessment.CheckCloudServiceID(req.CloudServiceId)
+	if err != nil {
+		return fmt.Errorf("%s: %w", ErrRequestCloudServiceID, err)
```
oxisto, line 55: I would also just return the `err` here without any wrapping, I think this is necessary. and then we can get rid of ErrRequestCloudServiceID

api/orchestrator/orchestrator.go
```diff
@@ -28,21 +28,57 @@ package orchestrator
...
+
+	return nil
+}
+
+// Validate validates the GetMetricConfigurationRequest
+func (req *GetMetricConfigurationRequest) Validate() error {
+	// Check cloud service ID
+	err := assessment.CheckCloudServiceID(req.CloudServiceId)
+	if err != nil {
+		return fmt.Errorf("%s: %w", ErrRequestCloudServiceID, err)
```
oxisto, line 71: no wrap

api/orchestrator/orchestrator.go
```diff
@@ -28,21 +28,57 @@ package orchestrator
...
+	}
+
+	// Check metric ID
+	if req.MetricId == "" {
+		return fmt.Errorf("%s: %w", assessment.ErrMetricIdMissing, err)
```
oxisto, line 76: are you wrapping an previous error here?

api/orchestrator/orchestrator.go
```diff
@@ -28,21 +28,57 @@ package orchestrator
...
+func (req *UpdateMetricConfigurationRequest) Validate() error {
+	// Check cloud service ID
+	err := assessment.CheckCloudServiceID(req.CloudServiceId)
+	if err != nil {
+		return fmt.Errorf("%s: %w", ErrRequestCloudServiceID, err)
+	}
+
+	// Check metric ID
+	if req.MetricId == "" {
+		return fmt.Errorf("%s: %w", assessment.ErrMetricIdMissing, err)
```
oxisto, line 60: are you wrapping an previous error here?

oxisto: we should probably just return ErrMetricIdMissing

oxisto (approved):

PR #820, opened 2022-10-18
Improved AWS and Azure discoverer

oxisto:
This PR improves the discoverers, mainly for AWS and Azure. It introduces discovery of AWS network interfaces and volumes.


oxisto:
Blocked by #821

anatheka:
@oxisto I have fixed the tests and changed the generation for the storage blob and files client. If this fits, I can do the same for the others in Azure.

anatheka:
I have added the missing labels to K8S, except the label for the volume assigned to a pod. The volume label we only get directly from the related storage. The question is, if that is than the right label or if we should use the pod label instead.

oxisto:
> @oxisto I have fixed the tests and changed the generation for the storage blob and files client. If this fits, I can do the same for the others in Azure.

Ah good! The tests still seem to fail because I moved some of the parts related to Azure Disks from storage to compute and now the tests cannot find the necessary mocks.

anatheka:
> @oxisto I have fixed the tests and changed the generation for the storage blob and files client. If this fits, I can do the same for the others in Azure.

Should I refactor the Azure client generation for compute and network as well?

service/discovery/aws/compute.go
```diff
@@ -253,11 +364,45 @@ func (*computeDiscovery) getNameOfVM(vm *typesEC2.Instance) string {
...
+		if aws.ToString(tag.Key) == "Name" {
+			return aws.ToString(tag.Value)
+		}
+	}
+
+	// If no tag with 'name' was found, return instanceId instead
+	return aws.ToString(vm.VolumeId)
+}
+
+// getNameOfVolume returns the name if exists (i.e. a tag with key 'name' exists), otherwise instance ID is used
```
oxisto, line 379: wrong documentation text

service/discovery/aws/compute.go
```diff
@@ -253,11 +364,45 @@ func (*computeDiscovery) getNameOfVM(vm *typesEC2.Instance) string {
 	return aws.ToString(vm.InstanceId)
 }
 
-// addARNToVM generates the ARN of a VM instance
-func (d computeDiscovery) addARNToVM(vm *typesEC2.Instance) voc.ResourceID {
+// getNameOfVolume returns the name if exists (i.e. a tag with key 'name' exists), otherwise instance ID is used
+func (*computeDiscovery) getNameOfVolume(vm *typesEC2.Volume) string {
```
oxisto, line 368: this is probably absolute now that we have `nameOrId`

service/discovery/azure/azure.go
```diff
@@ -82,6 +85,12 @@ type azureDiscovery struct {
 	cred                azcore.TokenCredential
 	clientOptions       arm.ClientOptions
 	discovererComponent string
+	client              client
+}
+
+type client struct {
```
oxisto, line 91: I would suggest `clients`

service/discovery/azure/azure.go
```diff
@@ -82,6 +85,12 @@ type azureDiscovery struct {
 	cred                azcore.TokenCredential
 	clientOptions       arm.ClientOptions
 	discovererComponent string
+	client              client
```
oxisto, line 88: `clients`

service/discovery/azure/storage.go
```diff
@@ -523,3 +378,35 @@ func accountName(id string) string {
 	splitName := strings.Split(id, "/")
 	return splitName[8]
 }
+
+// initializeBlobContainerClient creates the client if not already exists
+func (d *azureStorageDiscovery) initializeBlobContainerClient() (err error) {
```
oxisto, line 383: I think we just used `initXXX` somewhere else

service/discovery/azure/storage.go
```diff
@@ -523,3 +378,35 @@ func accountName(id string) string {
...
+		err = fmt.Errorf("could not get new blob containers client: %w", err)
+		log.Debug(err)
+		return err
+	}
+
+	return
+}
+
+// initializeFileStorageClient creates the client if not already exists
+func (d *azureStorageDiscovery) initializeFileStorageClient() (err error) {
```
oxisto, line 399: same as above

service/discovery/discovery.go
```diff
@@ -69,6 +69,16 @@ type grpcTarget struct {
 	opts   []grpc.DialOption
 }
 
+const DiscoveryEventTypeDiscovererStart = 0
+const DiscoveryEventTypeDiscovererFinished = 1
+
+type DiscoveryEvent struct {
```
oxisto, line 75: missing documentation

oxisto (commented):
Some review comments, some of which I should probably address myself :)

service/discovery/azure/storage.go
anatheka: You mean the naming or do you wanted to use the other generic function? 

Is it possible to change these methods to generic methods?

oxisto: I was referring to the naming. Would be interesting if these could be generic functions.

anatheka (approved):
:+1:

PR #822, opened 2022-10-18
Ensuring the existance of the `.clouditor` directory when starting the engine

### LINKED ISSUE #786: Creating API Key fails since ".clouditor" directory doesn't exist
**Describe the bug**
```
INFO[0000] API key does not exist at the default location yet. We will create a new one  component=auth
ERRO[0000] Error while saving the new API key: error while opening the file: open /home/lebogg/.clouditor/api.key: no such file or directory  component=auth
```

**To Reproduce**
Start engine with non-existing folder ".clouditor" in home

**Expected behavior**
Clouditor creates folder in home when it not exist or in  "/root" in the case of root (e.g. using Docker)

oxisto:
Fixes #786 by checking for the existance of the directory and creating it, if necessary.

### DIFF
--- a/internal/auth/key_loader.go
+++ b/internal/auth/key_loader.go
@@ -31,7 +31,6 @@ import (
 	"crypto/rand"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"os/user"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


anatheka (approved):
Great!

PR #834, opened 2022-10-26
Removing hard-coded test token

oxisto:
Even though it is only used for testing, this PR replaces the hard-coded test JWT with one that gets generated. This should silence all automated tools that look for "leaked" keys.


lebogg (approved):
LGTM.

PR #839, opened 2022-10-28
Update Catalog descriptions

anatheka:


oxisto (approved):

oxisto (commented):
Better. But not sure if anyone who reads this for the first time will know exactly what is happening. We should probably also document the implementation in the `service` package what is happening with preload, etc. as well.

anatheka:
> Better. But not sure if anyone who reads this for the first time will know exactly what is happening. We should probably also document the implementation in the `service` package what is happening with preload, etc. as well.
That is right, maybe we should do that also in this PR.

anatheka:
I updated the service descriptions, but I don't know what is happening with preload, etc. Someone else must describe that. 
Furthermore, where do we want describe the catalog structure anyway? Especially with the two control levels, this is not quite trivial. README, proto files, ...?

PR #840, opened 2022-10-28
Update Ontology files

anatheka:
Fix in ontology:
- [x] VirtualMachine has list of BlockStorages
- [x] Compute has a list of NetworkInterfaces
- [x] LoggingService has multiple Storage
- [x] Delete Metric
- [x] Delete RiskImpact
- [x] Delete RequirementCatalog
- [x] Delete Requirement


anatheka:
I would say we ignore the DeepSource issue, what do you think?

service/discovery/azure/compute.go
```diff
@@ -154,7 +154,7 @@ func (*azureComputeDiscovery) handleFunction(function *armappservice.Site) voc.I
 				},
 				Labels: labels(function.Tags),
 			},
-			NetworkInterface: []voc.ResourceID{},
+			NetworkInterfaces: []voc.ResourceID{},
```
oxisto, line 157: Do we have to adjust any Rego rules? I think the related-resources stuff is not yet merged in, so we definitely need to keep this in mind once we do.

service/discovery/azure/compute.go
```diff
@@ -221,9 +221,9 @@ func (*azureComputeDiscovery) handleVirtualMachines(vm *armcompute.VirtualMachin
 				},
 				Labels: labels(vm.Tags),
 			},
-			NetworkInterface: []voc.ResourceID{},
+			NetworkInterfaces: []voc.ResourceID{},
 		},
-		BlockStorage:      []voc.ResourceID{},
+		BlockStorages:     []voc.ResourceID{},
```
oxisto, line 226: what comes to my mind though, is that actually the word `storage` has no plural form as far as I know. Should we just stick with `BlockStorage` + array? Might affect other properties as well but might be better comprehensible.

voc/access_restriction.go
```diff
@@ -29,6 +31,6 @@ type AccessRestriction struct {
 	*Authorization
 }
 
-func (*AccessRestriction) Type() string {
+func (a *AccessRestriction) Type() string {
```
oxisto, line 34: If it is not too much work I would prefer leaving the receiver our in the Type() functions. I think we cannot ignore deep source issues just for specific files and I think the way without receiver is actually cleaner.

voc/anomaly_detection.go
```diff
@@ -25,8 +27,10 @@
 
 package voc
 
+// AnomalyDetection analyzes the activity of a NetworkService (which includes DatabaseServices)
 type AnomalyDetection struct {
 	*Auditing
-	ApplicationLogging *ApplicationLogging `json:"applicationLogging"`
-	Enabled            bool                `json:"enabled"`
+	ApplicationLogging	*ApplicationLogging 	`json:"applicationLogging"`
```
oxisto, line 33: we should probably run `go fmt` on these files after generating them? Otherwise its hard to see the actual changes in the structs.

voc/application.go
```diff
@@ -25,9 +27,12 @@
 
 package voc
 
+// Application This encapsulates the whole (source) code of an application.
```
oxisto, line 30: maybe a "hack" to have slightly better auto-documented structs. Can we have something like `$entity is an entity in our Cloud ontology. $description_sentence`? Otherwise we could have some kind of convention that we should have `$Entity is...` descriptions in the ontology, but we will probably break it at some point.

voc/automatic_updates.go
```diff
@@ -27,9 +29,12 @@ package voc
...
 
+// AutomaticUpdates This feature is, e.g., available on some VM services to automatically update their software. It ensures that a resource is protected from tampering with its state.
 type AutomaticUpdates struct {
 	*Integrity
-	Enabled      bool          `json:"enabled"`
-	Interval     time.Duration `json:"interval"`
-	SecurityOnly bool          `json:"securityOnly"`
+	Enabled	bool	`json:"enabled"`
+	SecurityOnly	bool	`json:"securityOnly"`
+	// The interval refers to the update interval in days.
```
oxisto, line 37: woah nice! How did you do that now?

voc/logging_service.go
```diff
@@ -25,7 +27,9 @@
 
 package voc
 
+// LoggingService A logging-as-a-service offering, e.g. for analyzing logs; has a Storage resource that stores the logs
 type LoggingService struct {
 	*NetworkService
-	Storage []ResourceID `json:"storage"`
+	Storage	ResourceID	`json:"storage"`
```
oxisto, line 33: Do we intentionally want only a single storage object here? Maybe also keep the array but the "storage" name?

voc/metric.go
```diff
@@ -25,9 +27,8 @@
 
 package voc
 
-type LogOutput struct {
-	*Functionality
-	Logging []ResourceID `json:"logging"`
-	Call    string       `json:"call"`
-	Value   string       `json:"value"`
+type Metric struct {
```
oxisto, line 30: I think this was actually a left over from a MEDINA experiment and can be deleted.

voc/requirement.go
```diff
@@ -25,6 +27,8 @@
 
 package voc
 
-type OSLog struct {
-	*Log
+type Requirement struct {
```
oxisto, line 30: can also be deleted

voc/requirement_catalog.go
```diff
@@ -0,0 +1,34 @@
...
+// $$ /      $$ |$$ /  $$ |$$ |  $$ |$$ /  $$ |$$ |  $$ |    $$ /  $$ |$$ | \__|
+// $$ |      $$ |$$ |  $$ |$$ |  $$ |$$ |  $$ |$$ |  $$ |$$\ $$ |  $$ |$$ |
+// \$$$$$$\  $$ |\$$$$$   |\$$$$$   |\$$$$$$  |$$ |  \$$$   |\$$$$$   |$$ |
+//  \_______|\__| \______/  \______/  \_______|\__|   \____/  \______/ \__|
+//
+// This file is part of Clouditor Community Edition.
+
+package voc
+
+type RequirementCatalog struct {
```
oxisto, line 30: delete

voc/risk_impact.go
```diff
@@ -26,5 +28,7 @@
 package voc
 
 type RiskImpact struct {
```
oxisto, line 30: delete

service/discovery/azure/compute.go
anatheka: Yes, that's true. You mean we call that "BlockStorageArray". Then I would adjust that in general.

voc/access_restriction.go
anatheka: True, this is the simplest solution :+1:  That's easy to change.

voc/anomaly_detection.go
anatheka: I forgot that.  It would be cool if you could do this directly in the owl2java code, but I don't know how.

voc/application.go
anatheka: Very good idea.

voc/logging_service.go
anatheka: Than we have to change that in the Ontology.

service/discovery/azure/compute.go
oxisto: No I meant the name `BlockStorage` and array as data type

voc/logging_service.go
oxisto: Yes, ontology change. I think its reasonable to assume that a logging service has multiple locations where things are stored.

voc/requirement.go
anatheka: I deleted the Requirement in the Ontology.

voc/metric.go
anatheka: I deleted the Metric in the Ontology.

voc/requirement_catalog.go
anatheka: I deleted the RequirementCatalog in the Ontology.

voc/risk_impact.go
anatheka: I delete the RiskImpact in the Ontology.

service/discovery/azure/compute.go
anatheka: What should I do with that. We also could do it in singular. I can live well with both :woman_shrugging:

service/discovery/azure/compute.go
anatheka: For now I changed it as follows: Strings that contain the word "storage" (case insensitive) aren't converted to plural. We should decide the comment above about the plural or singular version.

voc/automatic_updates.go
anatheka: That was a cramp and not easy at all at this confused format. 

How I did that? I have absolutely no idea. But now it works :)

service/discovery/azure/compute.go
oxisto: I think in general it should be plural, the only issue is that `storage` has no plural :)

anatheka:
Ready for the next review!

Issue #831, opened 2022-10-25
Refine target of evaluation / selection of controls


oxisto:
**Is your feature request related to a problem? Please describe.**

I want to differentiate between

- Controls, that are under continuous monitoring (e.g. by Clouditor)
- Controls, that I monitor manually, e.g., by a questionnaire
- Controls, that are delegated to somebody else, e.g. the cloud provider that is "below" me

The default for controls when a new target of evaluation is created is "manually monitored".

**Describe the solution you'd like**

Probably three different lists in the `TargetOfEvaluation`

**Additional context**

Also needs some UI elements.

PR #852, opened 2022-11-08
Add toe hooks

anatheka:


anatheka:
@oxisto do we block the PR until you are ready or just merge it now?

anatheka:
We could also think about to use genercis for the hook stuff.

oxisto:
> @oxisto do we block the PR until you are ready or just merge it now?

I would wait until I implement #831

api/orchestrator/orchestrator.proto
```diff
@@ -824,3 +824,14 @@ message State {
   // Reference to the certificate
   string certificate_id = 5;
 }
+
+// ToeStatus represents the status of the Target of Evaluation. 
+// With that status the ToE hook function can be informed about the reason of its call. 
+message ToeStatus {
```
oxisto, line 830: I think the other message are named similar to `ChangeEvent`, status might be a bit misleading.
Also `0` must be `_UNSPECIFIED` according to the style guide and the rest of the enums should have similar names like the enum itself. See https://developers.google.com/protocol-buffers/docs/proto3#enum

We probably don't follow this everywhere (yet).

service/orchestrator/toe.go
```diff
@@ -153,9 +156,32 @@ func (svc *Service) RemoveTargetOfEvaluation(_ context.Context, req *orchestrato
 		return nil, status.Errorf(codes.Internal, "database error: %v", err)
 	}
 
+	go svc.informToeHooks(ctx, nil, orchestrator.ToeStatus_TOE_REMOVE, nil)
 	return &emptypb.Empty{}, nil
 }
 
+// informToeHooks informs the registered hook functions about create, update or remove of the Target of Evaluation
+func (s *Service) informToeHooks(ctx context.Context, toe *orchestrator.TargetOfEvaluation, status orchestrator.ToeStatus_Status, err error) {
```
oxisto, line 164: Would it make sense to call this hook also from the update monitoringcontrol status?. Since in this case the "contents" / scope of the ToE has changed? I think this is needed for MEDINA, where we we need to find out if a control has changed to continuously monitored?

api/orchestrator/orchestrator.go
```diff
@@ -33,6 +33,7 @@ import (
 )
 
 type CloudServiceHookFunc func(ctx context.Context, cld *CloudService, err error)
+type ToeHookFunc func(ctx context.Context, toe *TargetOfEvaluation, status ToeStatus_Status, err error)
```
oxisto, line 36: I think we only used "Toe" in the URLs, in the code itself I would use the "long" version TargetOfEvaluation.

api/orchestrator/orchestrator.proto
```diff
@@ -708,22 +708,22 @@ message ListTargetsOfEvaluationRequest {
...
 }
 
 message ListTargetsOfEvaluationResponse {
-  repeated TargetOfEvaluation toes = 1;
+  repeated TargetOfEvaluation target_of_evaluation = 1;
   string next_page_token = 2;
 }
 
 message UpdateTargetOfEvaluationRequest {
   string cloud_service_id = 1;
```
oxisto, line 724: According to our style guide we can directly use the embedded message instead of repeating cloud_service_id and catalog_id. This is something we missed before, but if we are any way changing he message, we might as well correct it here.

api/orchestrator/orchestrator.proto
```diff
@@ -824,3 +824,25 @@ message State {
   // Reference to the certificate
   string certificate_id = 5;
 }
+
+// TargetOfEvaluationChangeEvent represents a change of either a Target of Evaluation or Control Monitoring Status and the TargetOfEvaluation hook function can be informed about the reason of its call. 
+message TargetOfEvaluationChangeEvent {
+  // Type represents the type of the change event. Type can be a change event regarding the Target of Evaluation or the Control Monitoring Status.
+  enum Type {
+    Type_UNSPECIFIED = 0;
```
oxisto, line 832: Type_ should be UPPERCASE

api/orchestrator/orchestrator.proto
```diff
@@ -824,3 +824,25 @@ message State {
...
+
+    TargetOfEvaluation_CREATED = 1;
```
oxisto, line 834:
```suggestion
    TYPE_TARGET_OF_EVALUATION_CREATED = 1;
```

oxisto: the other ones similar

oxisto (commented):
minor enum things, otherwise it looks good.

api/orchestrator/orchestrator.proto
```diff
@@ -824,3 +821,25 @@ message State {
...
+message TargetOfEvaluationChangeEvent {
+  // Type represents the type of the change event. Type can be a change event regarding the Target of Evaluation or the Control Monitoring Status.
+  enum Type {
+    TYPE_UNSPECIFIED = 0;
+
+    TYPE_TARGET_OF_EVALUATION_CREATED = 1;
+    TYPE_TARGET_OF_EVALUATION_UPDATED = 2;
+    TYPE_TARGET_OF_EVALUATION_REMOVED = 3;
+    
+    TYPE_CONTROL_MONITORING_STATUS_UPDATED = 4;
```
oxisto, line 835: Did you forgot to actually trigger the hook in UpdateControlMonitoringStatus?

oxisto (approved):

PR #853, opened 2022-11-10
Add evaluation

anatheka:
This PR adds the `evaluation` service. This service aggregates assessment results and provides an overview if a control catalog is fulfilled or not.


api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,133 @@
...
+
+
+ option go_package = "clouditor.io/clouditor/api/evaluation";
+
+ // This service represents the Evaluation Interface
+ service Evaluation {
+    // StartEvaluation evaluates all assessment results of an service based on its Target of Evaluation (binding of a cloud service to a catalog). The assessment results are evaluated regarding the contol ID.
+    rpc StartEvaluation(StartEvaluationRequest) returns (StartEvaluationResponse) {
+    option (google.api.http) = {
+        post : "/v1/evaluation/evaluate/{category_name}/{control_id}/start"
```
oxisto, line 45: Do we really want to start the evaluation on an individual control level? We might need to evaluate all (selected) controls in a target of evaluation? Is this blocked by the PR that sets up the "continuously monitored" status?

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,133 @@
...
+        COMPLIANT = 1;
+        NOT_COMPLIANT = 2;
+        PENDING = 3;
+    }
+    
+    // Evaluation status
+    EvaluationStatus status = 2;
+    
+    // Reference to the control the evaluation was based on
+    string control = 3;
```
oxisto, line 119: do we need the category name here as well?

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,133 @@
...
+
+    // Time of evaluation
+    google.protobuf.Timestamp timestamp = 4
+      [ (tagger.tags) = "gorm:\"serializer:timestamppb;type:time\"" ];
+
+    // The cloud service this evaluation result belongs to
+    string cloud_service_id = 5;
```
oxisto, line 126: isn't this already included in the toe?

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,133 @@
...
+    string order_by = 12;
+    bool asc = 13;
+  }
+  message ListEvaluationResultsResponse {
+    repeated EvaluationResult results = 1;
+    string next_page_token = 2;
+  }
+
+message StartEvaluationRequest {
+    TargetOfEvaluation toe = 1;
```
oxisto, line 81: toe or target_of_evaluation? I think we only used toe in the URL but not in the fields. but not sure.

internal/testutil/prototest/prototest.proto
```diff
@@ -0,0 +1,44 @@
...
+// This file is part of Clouditor Community Edition.
+
+syntax = "proto3";
+package prototest;
+
+option go_package = "clouditor.io/clouditor/internal/testutil/prototest";
+
+// This is a test message. If a test needs a proto message, that one can be used. 
+// This has been created because of import cycles in the package util.
+message TestStruct {
```
oxisto, line 33: not entirely sure why this is needed :D

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,429 @@
...
+		// TODO(anatheka): Do we need that? Or do we let it running?
+		// Delete evaluation entry, it is no longer needed if we don't get the assessment results from the orchestrator
+		s.evaluationMutex.Lock()
+		delete(s.evaluation, createSchedulerTag(req.Toe.CloudServiceId, req.ControlId))
+		s.evaluationMutex.Unlock()
+		return
+	}
+
+	// Get mapping assessment results related to the metric
+	mappingList := getMapping(assessmentResults, metrics)
```
oxisto, line 265: I would extract this into the `ListAssessmentResults` call of the orchestrator.

api/evaluation/evaluation.proto
anatheka: Yes, that is blocked by that. I just wanted to do it first with an individual control and then we'll update it when the PR is ready.

anatheka: Not necessarily, but it makes it easier. But I remove the category when the other PR is ready.

api/evaluation/evaluation.proto
anatheka: :+1:

api/evaluation/evaluation.proto
anatheka: You're right. I will change that.

internal/testutil/prototest/prototest.proto
anatheka: That is needed for internal/util/struct_test.go. There the Certificate struct was used. And it came to an import     errors when using a function in the internal/util/deref.go.

service/evaluation/evaluation.go
anatheka: @oxisto should I start a PR for that or do you have something where you can add it?

api/evaluation/evaluation.proto
anatheka: Oh, I was in the wrong line. I don 't know if we need the category name, but we can add it.

anatheka:
Is it possible that a control has sub-controls and metrics?

anatheka:
Do we want the whole assessment result in evaluationResult.FailingAssessmentResults or only the ID?

anatheka:
> How to fix that error?
> 
> `cannot use req (variable of type *"clouditor.io/clouditor/api/evaluation".StopEvaluationRequest) as service.IncomingRequest value in argument to service.ValidateRequest: *"clouditor.io/clouditor/api/evaluation".StopEvaluationRequest does not implement service.IncomingRequest (missing method Validate) [service/evaluation/evaluation.go](https://github.com/clouditor/clouditor/blob/60d8035411410a14e73fa7b63f273ff5c1b6a80f/service/evaluation/evaluation.go#L280-L280)`

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,145 @@
...
+ import "api/orchestrator/orchestrator.proto";
+ import "tagger/tagger.proto";
+ import "validate/validate.proto";
+
+ option go_package = "clouditor.io/clouditor/api/evaluation";
+
+ // This service represents the Evaluation Interface
+ service Evaluation {
+    // StartEvaluation evaluates all assessment results of an service based on its Target of Evaluation (binding of a cloud service to a catalog). The assessment results are evaluated regarding the contol ID.
+    rpc StartEvaluation(StartEvaluationRequest) returns (StartEvaluationResponse) {
```
oxisto, line 43: Is the evaluation then executed regularly on its own after this call? This would be good  to document as well.

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,145 @@
...
+    option (google.api.http) = {
+        post : "/v1/evaluation/evaluate/start"
+        body : "*"
+      };
+     } 
+
+    // StopEvaluation stop the evaluation for the given Cloud Service
```
oxisto, line 50:
```suggestion
    // StopEvaluation stops the evaluation for the given target of evaluation
```

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,145 @@
...
+ import "validate/validate.proto";
+
+ option go_package = "clouditor.io/clouditor/api/evaluation";
+
+ // This service represents the Evaluation Interface
+ service Evaluation {
+    // StartEvaluation evaluates all assessment results of an service based on its Target of Evaluation (binding of a cloud service to a catalog). The assessment results are evaluated regarding the contol ID.
+    rpc StartEvaluation(StartEvaluationRequest) returns (StartEvaluationResponse) {
+    option (google.api.http) = {
+        post : "/v1/evaluation/evaluate/start"
```
oxisto, line 45: path parameters in post-query are missing

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,145 @@
...
+        post : "/v1/evaluation/evaluate/{category_name}/{control_id}/stop"
+        body : "target_of_evaluation"
+        };
+    } 
+
+    // List all evaluation results. Part of the public API, also exposed as REST.
+    rpc ListEvaluationResults(ListEvaluationResultsRequest)
+        returns (ListEvaluationResultsResponse) {
+        option (google.api.http) = {
+            get : "/v1/evaluation/results"
```
oxisto, line 62: additional bindings for filtering per cloud service?

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,145 @@
...
+    string order_by = 12;
+    bool asc = 13;
+  }
+  message ListEvaluationResultsResponse {
+    repeated EvaluationResult results = 1;
+    string next_page_token = 2;
+  }
+
+message StartEvaluationRequest {
+    clouditor.orchestrator.v1.TargetOfEvaluation target_of_evaluation = 1
```
oxisto, line 85: do we really want the complete message here and not just the service ID and the catalog ID?

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,145 @@
...
+    optional int32 interval = 2;
+}
+
+message StartEvaluationResponse {
+    bool status = 1; 
+    string status_message = 2; 
+}
+
+message StopEvaluationRequest {
+    clouditor.orchestrator.v1.TargetOfEvaluation target_of_evaluation = 1
```
oxisto, line 97: see comment above

service/orchestrator/orchestrator.go
```diff
@@ -53,7 +53,7 @@ var log *logrus.Entry
 
 var DefaultMetricsFile = "metrics.json"
 
-var DefaultCatalogsFile = "catalogs.json"
+var DefaultCatalogsFile = "demo_catalogs.json"
```
oxisto, line 56: we do not want to store the whole EUCS here ;)

oxisto (commented):
First round of comments, will look at the actual evaluation implementation later.

api/evaluation/evaluation.proto
anatheka: You are right, after the other changes, we can get the ToE from the orchestrator. :+1:

service/orchestrator/orchestrator.go
anatheka: But we are not storing the whole EUCS catalog here. I only changed the naming from catalogs to demo_catalogs.json. The content is the same as before. Or do I miss something?

oxisto: You added the whole EUCS in service/orchestrator/eucs_catalogs.json

anatheka: Oh, you are right. I don't know what happened.... I will change it.

anatheka:
I doesn't like the WaitGroups for the sub-controls. Currently it is as follows: 
If the evaluation for a sub-control is started it starts as one job. That is easy. 
If the evaluation for a parent control is started all sub-controls are started as well and the parent control waits with its evaluation until all the sub-controls are evaluated. This is done with the WaitGroups. 

I see a problems with this approach: Maybe we want to change the evaluated sub-controls. Than we have to restart all jobs for the sub-controls and the parent control (without the affected sub-control). Otherwise we will stuck in the WaitGroup.

Maybe we should start the parent and sub-controls individually and use for the parent control evaluation only the last evaluation results of the sub-controls. Than we have to wait a moment for the evaluation of a parent control and maybe we use "old" results (depends on the time interval what old means), but I think that would be a better approach.

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,853 @@
...
+	// Scheduler tags must be unique to find the jobs by tag name
+	s.scheduler.TagsUnique()
+
+	// Add the controls_in_scope of the target_of_evaluation including their sub-controls to the scheduler
+	for _, c := range toe.GetControlsInScope() {
+		var controls []*orchestrator.Control
+
+		// TODO(anatheka): I think we do not need that anymore as we are using the controls from the ToE and these should be ok.
+		// We must get the control from the orchestrator, as we do not know if the request is correct and the parent_ids are probably missing
+		control, err := s.getControl(c.GetCategoryCatalogId(), c.GetCategoryName(), c.GetId())
```
oxisto, line 223: As discussed in the call: Only use the sub controls specified in the `controlsInScope` and NOT fetch any (sub)controls from orchestrator.

oxisto:
Currently blocked, until we merge https://github.com/clouditor/clouditor/pull/976

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,164 @@
...
+message EvaluationResult {
+  // Evaluation result id
+  string id = 1
+    [(validate.rules).string = {ignore_empty : true, uuid : true}]; 
+
+  // The Cloud Service ID the evaluation belongs to
+  string cloud_service_id = 2
+  [(validate.rules).string = {ignore_empty : true, uuid : true}]; 
+
+  // The category the control belongs to
```
oxisto, line 129: ?

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,164 @@
...
+
+  // The Cloud Service ID the evaluation belongs to
+  string cloud_service_id = 2
+  [(validate.rules).string = {ignore_empty : true, uuid : true}]; 
+
+  // The category the control belongs to 
+  string category_name = 3
+    [(validate.rules).string.min_len = 1]; 
+
+  // The catalog the control belongs to
```
oxisto, line 133: the control?

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,164 @@
...
+  // The catalog the control belongs to 
+  string catalog_id = 4
+    [(validate.rules).string.min_len = 1]; 
+
+  // The evaluated resource 
+  string resource_id = 5
+    [(validate.rules).string.min_len = 1]; 
+  
+    enum EvaluationStatus {
+    STATUS_UNSPECIFIED = 0;
```
oxisto, line 142: enum style: should be EVALUATION_STATUS_UNSPECIFIED and the others should begin with EVALUATION_STATUS

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,164 @@
...
+  // The control id the evaluation was based on
+  string control_id = 7
+    [(validate.rules).string.min_len = 1];
+  
+  // Time of evaluation
+  google.protobuf.Timestamp timestamp = 8
+    [(tagger.tags) = "gorm:\"serializer:timestamppb;type:time\"" ];
+
+  // List of assessment results because of which the evaluation status is not 'compliant'
+  repeated string failing_assessment_results_id = 9 [
```
oxisto, line 161: result_ids?

cmd/engine/engine.go
```diff
@@ -255,7 +259,7 @@ func doCmd(_ *cobra.Command, _ []string) (err error) {
 			}),
 	)
 
-	orchestratorService = service_orchestrator.NewService(service_orchestrator.WithStorage(db))
+	orchestratorService = service_orchestrator.NewService(service_orchestrator.WithStorage(db), service_orchestrator.WithCatalogsFile("eucs_catalogs.json"))
```
oxisto, line 262: remove hardcoded catalog path.

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,845 @@
...
+	"time"
+
+	"clouditor.io/clouditor/api"
+	"clouditor.io/clouditor/api/assessment"
+	"clouditor.io/clouditor/api/evaluation"
+	"clouditor.io/clouditor/api/orchestrator"
+	"clouditor.io/clouditor/internal/util"
+	"clouditor.io/clouditor/persistence"
+	"clouditor.io/clouditor/persistence/inmemory"
+	"clouditor.io/clouditor/service"
```
oxisto, line 43: Formatting looks odd

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,845 @@
...
+	// defaultinterval is the default interval time for the scheduler. If no interval is set in the StartEvaluationRequest, the default value is taken.
+	defaultInterval int = 5
+)
+
+type grpcTarget struct {
+	target string
+	opts   []grpc.DialOption
+}
+
+type WaitGroup struct {
```
oxisto, line 71: unexport?

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,845 @@
...
+		}
+
+		// parentSchedulerTag is the tag for the parent control (e.g., OPS-13)
+		parentSchedulerTag = getSchedulerTag(toe.GetCloudServiceId(), control.GetId())
+
+		// Add number of sub-controls to the WaitGroup. For the control (e.g., OPS-13) we have to wait until all the sub-controls (e.g., OPS-13.1) are ready.
+		// Control is a parent control if no parentControlId exists.
+		s.wg[parentSchedulerTag] = &WaitGroup{
+			wg:      &sync.WaitGroup{},
+			wgMutex: sync.Mutex{},
```
oxisto, line 255: What's the point of the mutex *inside* the map? Shouldn't the mutex protect the map? The wait group itself should already be thread-safe. Not quite sure what you want to achieve here :)

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,845 @@
...
+	)
+
+	// Validate request
+	err = service.ValidateRequest(req)
+	if err != nil {
+		err = status.Errorf(codes.InvalidArgument, "%v", err)
+		return
+	}
+
+	// Get Target of Evaluation. The Target of Evaluation is retrieved to get the controls_in_scope, which are then stopped from the evaluation.
```
oxisto, line 306: Wouldn't it be easier to have a tag for the ToE itself (cloudservice+catalog) instead of looping over all controls again? Seems fragile.

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,845 @@
...
+	schedulerTag := getSchedulerTag(toe.GetCloudServiceId(), c.GetId())
+
+	// Regarding the control level the specific method is called every X minutes based on the given interval. We have to decide if a sub-control is started individually or a parent control that has to wait for the results of the sub-controls.
+	// If a parent control with its sub-controls is started, the parentSchedulerTag is empty and the ParentControlId is not set.
+	// If a sub-control is started individually the parentSchedulerTag is empty and ParentControlId is set.
+	if parentSchedulerTag == "" && c.ParentControlId == nil { // parent control
+		_, err = s.scheduler.
+			Every(interval).
+			Minute().
+			Tag(schedulerTag).
```
oxisto, line 456: Here I would add an additional tag about the ToE so we can quickly stop the evaluation again without knowing all the controls

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,845 @@
...
+		evalResult *evaluation.EvaluationResult
+	)
+
+	log.Infof("Start control evaluation for Cloud Service '%s', Catalog ID '%s' and Control '%s'", toe.CloudServiceId, toe.CatalogId, controlId)
+
+	// Wait till all sub-controls are evaluated
+	s.wg[schedulerTag].wg.Wait()
+
+	// For the next iteration set wg again to the number of sub-controls
+	s.wg[schedulerTag].wgMutex.Lock()
```
oxisto, line 490: Lock is not necessary here in my opinion

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,845 @@
...
+
+		// Create evaluation result
+		eval = &evaluation.EvaluationResult{
+			Id:                         uuid.NewString(),
+			Timestamp:                  timestamppb.Now(),
+			CategoryName:               categoryName,
+			ControlId:                  controlId,
+			CloudServiceId:             toe.GetCloudServiceId(),
+			CatalogId:                  toe.GetCatalogId(),
+			ResourceId:                 result[0].GetResourceId(), // It does not matter which element is used here, since they all have the same resource_id.
```
oxisto, line 626: the cleaner solution would be to use the key of the `assessmentResultsMap`

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,845 @@
...
+		}
+	} else {
+		log.Debugf("no metrics are available for the given control")
+	}
+
+	// Here the actual evaluation takes place. For every resource_id we check if the asssessment results are compliant. If the latest assessment result per resource_id is not compliant the whole evaluation status is set to NOT_COMPLIANT. Furthermore, all non-compliant assessment_result_ids are stored in a separate list.
+
+	// Get a map of the assessment results, so that we have all assessment results for a specific resource_id and metric_id together for evaluation
+	assessmentResultsMap := getAssessmentResultMap(assessmentResults)
+	for _, result := range assessmentResultsMap {
```
oxisto, line 596: better name would probably be `results`, since its a list, otherwise its confusing.

api/evaluation/evaluation.proto
anatheka: My thought: The evaluation is based on a specific control, but the control_id is only unique in combination with the catalog_id and category_name. So, we need all three informations and the category_name and catalog_id belong to the control_id.

service/evaluation/evaluation.go
anatheka: Thatś right, I think this is an old mutex from the beginning that is now obsolete.  I will change that and try :)

api/evaluation/evaluation.proto
anatheka: What do you mean?

oxisto: maybe we should write something like 
```suggestion
  // The category the control (that was evaluated) belongs to 
```

Or something like that

api/evaluation/evaluation.proto
oxisto: Yes that makes sense. I was just referring to the docstring and was a bit confused why a control catalog was here.

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,165 @@
...
+  // The Cloud Service ID the evaluation belongs to
+  string cloud_service_id = 2
+  [(validate.rules).string = {ignore_empty : true, uuid : true}]; 
+
+  // The control id the evaluation was based on
+  string control_id = 3
+    [(validate.rules).string.min_len = 1];
+  
+  // The category the control belongs to 
+  string category_name = 4
```
oxisto, line 134: A better name would probably be:
```suggestion
  string control_category_name = 4
```

Then its clear that it is the category of the control and not a category of the eval result

service/evaluation/evaluation.go
anatheka: See the answer above. Now I am adding a scheduler for each ToE and remove it when stopping the evaluation for the specific ToE.

service/evaluation/evaluation.go
oxisto: I think both approaches work, so what ever you prefer.

service/evaluation/evaluation_test.go
```diff
@@ -0,0 +1,2145 @@
...
+				wantControl.Controls[0].Metrics = nil
+
+				if !proto.Equal(gotControl, wantControl) {
+					t.Errorf("Service.GetControl() = %v, want %v", gotControl, wantControl)
+					wantControl.Controls[0].Metrics = tmpMetrics
+					return false
+				}
+
+				wantControl.Controls[0].Metrics = tmpMetrics
+				return true
```
anatheka, line 1284: Does anyone have an idea on how to do this better?

oxisto: I am not quite sure what you want to achieve?

oxisto:
> TODO: Add the request filter, do we have some guideline for that?

Should be straightforward, since all calls have the cloud service ID in the request. So calling CheckAccess should be enough.

cmd/engine/engine.go
```diff
@@ -272,8 +276,20 @@ func doCmd(_ *cobra.Command, _ []string) (err error) {
 			},
 		),
 	)
+
 	evidenceStoreService = service_evidenceStore.NewService(service_evidenceStore.WithStorage(db))
 
+	evaluationService = service_evaluation.NewService(
+		service_evaluation.WithOAuth2Authorizer(
```
oxisto, line 283: WithStorage(db) is missing here

internal/testdata/testdata.go
```diff
@@ -7,6 +7,8 @@ import (
 )
 
 const (
+	MockOrchestratorAddress = "localhost:9090"
```
oxisto, line 10: shouldn't the mock orchestrator address something like "bufnet"?

internal/testutil/prototest/prototest.proto
```diff
@@ -41,4 +41,4 @@ message TestStruct {
         PENDING = 3;
```
oxisto, line 41: This file is weirdly formatted

internal/util/slice.go
```diff
@@ -0,0 +1,38 @@
...
+// $$ /      $$ |$$ /  $$ |$$ |  $$ |$$ /  $$ |$$ |  $$ |    $$ /  $$ |$$ | \__|
+// $$ |      $$ |$$ |  $$ |$$ |  $$ |$$ |  $$ |$$ |  $$ |$$\ $$ |  $$ |$$ |
+// \$$$$$$\  $$ |\$$$$$   |\$$$$$   |\$$$$$$  |$$ |  \$$$   |\$$$$$   |$$ |
+//  \_______|\__| \______/  \______/  \_______|\__|   \____/  \______/ \__|
+//
+// This file is part of Clouditor Community Edition.
+
+package util
+
+func RemoveIndexFromSlice[T any](slice []T, index int) []T {
```
oxisto, line 28: We can replace this with https://pkg.go.dev/golang.org/x/exp/slices#Delete from the experimental `slices` package, which should already be imported.

oxisto: And it seems to be unused anyway.

internal/util/slice_test.go
oxisto, line 1: Can be removed

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,783 @@
...
+
+const (
+	// DefaultOrchestratorAddress specifies the default gRPC address of the orchestrator service.
+	DefaultOrchestratorAddress = "localhost:9090"
+
+	// defaultinterval is the default interval time for the scheduler. If no interval is set in the StartEvaluationRequest, the default value is taken.
+	defaultInterval int = 5
+)
+
+type grpcTarget struct {
```
oxisto, line 66: We should definitely move this type to `service` or `internal/service` because we define it like 4 times. I will prepare this in a separate PR.

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,783 @@
...
+func (s *Service) SetAuthorizer(auth api.Authorizer) {
+	s.authorizer = auth
+}
+
+// Authorizer implements UsesAuthorizer
+func (s *Service) Authorizer() api.Authorizer {
+	return s.authorizer
+}
+
+// StartEvaluation is a method implementation of the evaluation interface: It periodically starts the evaluation of a cloud service and the given controls_in_scope (e.g., EUCS OPS-13, EUCS OPS-13.2) in the target_of_evaluation. If no inteval time is given, the default value is used.
```
oxisto, line 169:
```suggestion
// StartEvaluation is a method implementation of the evaluation interface: It periodically starts the evaluation of a cloud service based on its given controls_in_scope (e.g., EUCS OPS-13, EUCS OPS-13.2) in the target_of_evaluation. If no interval time is given, the default value is used.
```

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,783 @@
...
+		toe                *orchestrator.TargetOfEvaluation
+		toeTag             string
+	)
+
+	// Validate request
+	err = service.ValidateRequest(req)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "%v", err)
+	}
+
```
oxisto, line 183: Add `CheckAccess` here

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,783 @@
...
+		parentSchedulerTag string
+		interval           int
+		toe                *orchestrator.TargetOfEvaluation
+		toeTag             string
+	)
+
+	// Validate request
+	err = service.ValidateRequest(req)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "%v", err)
```
oxisto, line 181: You should directly return the `err` here, it's already wrapped in a gRPC error. That is even documented in the function ;)

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,783 @@
...
+		return nil, status.Errorf(codes.Internal, "%s", err)
+	}
+
+	// Get Target of Evaluation. The Target of Evaluation is retrieved to get the controls_in_scope, which are then evaluated.
+	toe, err = s.orchestratorClient.GetTargetOfEvaluation(context.Background(), &orchestrator.GetTargetOfEvaluationRequest{
+		CloudServiceId: req.GetCloudServiceId(),
+		CatalogId:      req.GetCatalogId(),
+	})
+	if err != nil {
+		err = fmt.Errorf("could not get target of evaluation: %v", err)
```
oxisto, line 205:
```suggestion
		err = fmt.Errorf("could not get target of evaluation: %w", err)
```

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,783 @@
...
+	if req.GetInterval() == 0 {
+		interval = defaultInterval
+	} else {
+		interval = int(req.GetInterval())
+	}
+
+	// Get orchestrator client. The orchestrator client is used to retrieve necessary information from the Orchestrator, such as assessment_results, controls or targets_of_evaluation.
+	err = s.initOrchestratorClient()
+	if err != nil {
+		err = fmt.Errorf("could not set orchestrator client: %v", err)
```
oxisto, line 194:
```suggestion
		err = fmt.Errorf("could not set orchestrator client: %w", err)
```

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,783 @@
...
+		CloudServiceId: req.GetCloudServiceId(),
+		CatalogId:      req.GetCatalogId(),
+	})
+	if err != nil {
+		err = fmt.Errorf("could not get target of evaluation: %v", err)
+		log.Error(err)
+		return nil, status.Errorf(codes.Internal, "%s", err)
+	}
+
+	// toeTag is the tag for the scheduler that contains the scheduler jobs for the specific ToE
```
oxisto, line 210: toe tag is maybe not a really suitable name for a variable ;) https://en.wikipedia.org/wiki/Toe_tag

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,783 @@
...
+			controls []*orchestrator.Control
+		)
+
+		// If the control does not have sub-controls don't start a scheduler.
+		if control.ParentControlId != nil {
+			continue
+		}
+
+		// Collect control including sub-controls from the controls_in_scope slice to start them later as scheduler jobs
+		controls = append(controls, control)
```
oxisto, line 241: Can be merged into one line:
```
controls = append(controls, control, control.GetControls()...)
```

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,783 @@
...
+		controls = append(controls, control.GetControls()...)
+
+		// parentSchedulerTag is the tag for the parent control (e.g., OPS-13)
+		parentSchedulerTag = getSchedulerTag(toe.GetCloudServiceId(), toe.GetCatalogId(), control.GetId())
```
oxisto, line 245: shouldn't the category name also be included? on the other hand, you probably won't need the cloud service ID because your schedulers are already limited to a ToE

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,783 @@
...
+				parentSchedulerTag = getSchedulerTag(toe.GetCloudServiceId(), toe.GetCatalogId(), control.GetParentControlId())
+			}
+			err = s.addJobToScheduler(control, toe, parentSchedulerTag, interval)
+			// We can return the error as it is
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		log.Infof("Evaluate Control ID '%s' for Cloud Service '%s' every 5 minutes...", control.GetId(), toe.GetCloudServiceId())
```
oxisto, line 269:
```suggestion
		log.Infof("Scheduled to evaluate control ID '%s' for cloud service '%s' every 5 minutes...", control.GetId(), toe.GetCloudServiceId())
```

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,783 @@
...
+
+	var toeTag string
+
+	// Validate request
+	err = service.ValidateRequest(req)
+	if err != nil {
+		err = status.Errorf(codes.InvalidArgument, "%v", err)
+		return
+	}
+
```
oxisto, line 292: TODO: Add CheckAccess

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,783 @@
...
+// StopEvaluation is a method implementation of the evaluation interface: It stops the evaluation for a TargetOfEvaluation.
+func (s *Service) StopEvaluation(_ context.Context, req *evaluation.StopEvaluationRequest) (resp *evaluation.StopEvaluationResponse, err error) {
+
+	var toeTag string
+
+	// Validate request
+	err = service.ValidateRequest(req)
+	if err != nil {
+		err = status.Errorf(codes.InvalidArgument, "%v", err)
```
oxisto, line 289: directly return err

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,783 @@
...
+		return
+	}
+
+	// toeTag is the tag for the scheduler that contains the scheduler jobs for the specific ToE
+	toeTag = getToeTag(req.GetCloudServiceId(), req.GetCatalogId())
+
+	// Stop scheduler for given Cloud Service
+	s.scheduler[toeTag].Stop()
```
oxisto, line 297: This will panic, if the scheduler for this "toe tag" does not exist

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,783 @@
...
+	// For the next iteration set wg again to the number of sub-controls
+	s.wg[schedulerTag].Add(len(subControls))
+
+	evaluations, err := s.ListEvaluationResults(context.Background(), &evaluation.ListEvaluationResultsRequest{
+		FilteredCloudServiceId: &toe.CloudServiceId,
+		FilteredSubControls:    util.Ref(fmt.Sprintf("%s.", controlId)), // We only need the sub-controls for the evaluation.
+		LatestByResourceId:     util.Ref(true),
+	})
+	if err != nil {
+		err = fmt.Errorf("error list evaluation results: %v", err)
```
oxisto, line 477: always use `%w` when using `fmt.Errorf`

service/evaluation/evaluation_test.go
anatheka: For the test I have to delete the Metrics. And if I do the following, the orchestratortest.MockControl1 is changed for all following tests and some tests fail. 
`
wantControl := orchestratortest.MockControl1
tmpMetrics := wantControl.Controls[0].Metrics
wantControl.Controls[0].Metrics = nil
`

I have no idea how to copy orchestratortest.MockControl1 without changing it permanently.

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,783 @@
...
+
+	return metricIds
+}
+
+// getAllMetricsFromControl returns all metrics from a given controlId
+// For now a control has either sub-controls or metrics. If the control has sub-controls, get also all metrics from the sub-controls.
+func (s *Service) getAllMetricsFromControl(catalogId, categoryName, controlId string) (metrics []*assessment.Metric, err error) {
+	var subControlMetrics []*assessment.Metric
+
+	control, err := s.getControl(catalogId, categoryName, controlId)
```
oxisto, line 642: This seems to be very inefficient. You are issuing a RPC for each individual control inside each evaluation cycle. Wouldn't it make more sense to just fetch the complete catalog once (per evaluation cycle) and then use this information instead of requesting information per-control?

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,783 @@
...
+// getToeTag creates a tag for the scheduler in the format 'cloud_service_id-catalog_id' ('00000000-0000-0000-0000-000000000000-EUCS')
+func getToeTag(cloudServiceId, catalogId string) string {
+	if cloudServiceId == "" || catalogId == "" {
+		return ""
+	}
+	return fmt.Sprintf("%s-%s", cloudServiceId, catalogId)
+}
+
+// getSchedulerTag creates the scheduler tag for a given cloud_service_id and control_id.
+func getSchedulerTag(cloudServiceId, catalogId, controlId string) string {
```
oxisto, line 733: should be catalog ID, categoryName and control ID instead

api/evaluation/evaluation.go
```diff
@@ -0,0 +1,26 @@
+// Copyright 2022 Fraunhofer AISEC
```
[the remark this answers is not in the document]
anatheka: :) I've started last year.

service/evaluation/evaluation.go
anatheka: That's a good point.

service/evaluation/evaluation.go
oxisto: I meant, directly return error without any wrapping

anatheka:
> > TODO: Add the request filter, do we have some guideline for that?
> 
> Should be straightforward, since all calls have the cloud service ID in the request. So calling CheckAccess should be enough.

Thanks, but I menat the filterings for the ListEvaluationResults endpoint.

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,859 @@
...
+func (s *Service) SetAuthorizer(auth api.Authorizer) {
+	s.authorizer = auth
+}
+
+// Authorizer implements UsesAuthorizer
+func (s *Service) Authorizer() api.Authorizer {
+	return s.authorizer
+}
+
+// StartEvaluation is a method implementation of the evaluation interface: It periodically starts the evaluation of a cloud service and the given controls_in_scope (e.g., EUCS OPS-13, EUCS OPS-13.2) in the target_of_evaluation. If no inteval time is given, the default value is used.
```
lebogg, line 180:
```suggestion
// StartEvaluation is a method implementation of the evaluation interface: It periodically starts the evaluation of a cloud service and the given controls_in_scope (e.g., EUCS OPS-13, EUCS OPS-13.2) in the target_of_evaluation. If no interval time is given, the default value is used.
```

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,859 @@
...
+	// map[catalog_id][category_name-control_id]*orchestrator.Control
+	catalogControls map[string]map[string]*orchestrator.Control
+}
+
+func init() {
+	log = logrus.WithField("component", "evaluation")
+}
+
+// ServiceOption is a function-style option to configure the Evaluation Service
+type ServiceOption func(*Service)
```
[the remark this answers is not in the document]
oxisto: We have a generic `service.Option[T]` for this already in `service/options.go`.

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,165 @@
...
+    [(validate.rules).string.min_len = 1];
+
+  // Optional. Latest results grouped by resource_id and metric_id.
+  optional bool latest_by_resource_id = 4;
+
+  int32 page_size = 10;
+  string page_token = 11;
+  string order_by = 12;
+  bool asc = 13;
+}
```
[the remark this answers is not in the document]
oxisto: The idea was to always have the pagination fields beginning at 10 and all the other fields before that, starting at one. So for me this ordering is the way it should be.

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,165 @@
...
+  EvaluationStatus status = 7 
+    [(validate.rules).enum.defined_only = true];
+  
+  
+  // Time of evaluation
+  google.protobuf.Timestamp timestamp = 8
+    [(tagger.tags) = "gorm:\"serializer:timestamppb;type:time\"" ];
+
+  // List of assessment results because of which the evaluation status is not 'compliant'
+  repeated string failing_assessment_result_ids = 9 [
```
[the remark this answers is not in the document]
oxisto: That would probably be too much. I would also propose just listing the failing ones here.

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,165 @@
...
+  // The control id the evaluation was based on
+  string control_id = 3
+    [(validate.rules).string.min_len = 1];
+  
+  // The category the evaluated control belongs to
+  string control_category_name = 4
+    [(validate.rules).string.min_len = 1]; 
+
+  // The catalog the evaluated control belongs to
+  string control_catalog_id = 5
```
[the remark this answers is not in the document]
anatheka: https://github.com/clouditor/clouditor/pull/853#discussion_r1148008979 :)

I would let it as it is. The important information is the control_id on which the evaluation is based. The category and catalog are only additional information for the unique control.

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,165 @@
...
+
+  // The interval time in minutes the evaluation executes periodically. The default interval is set to 5 minutes.
+  optional int32 interval = 3
+    [ (validate.rules).int32.gt = 0 ];
+}
+
+message StartEvaluationResponse {
+  bool status = 1; 
+  string status_message = 2; 
+}
```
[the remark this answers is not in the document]
anatheka: Now it is analogous to StartDiscoveryResponse.

api/evaluation/evaluation.proto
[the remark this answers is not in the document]
anatheka: Regarding the last 4 fields: I have also considered whether to make them optional. But I think that this is not necessary. What advantages would that bring us?

api/evaluation/evaluation.proto
```diff
@@ -0,0 +1,165 @@
...
+  // The category the evaluated control belongs to
+  string control_category_name = 4
+    [(validate.rules).string.min_len = 1]; 
+
+  // The catalog the evaluated control belongs to
+  string control_catalog_id = 5
+    [(validate.rules).string.min_len = 1]; 
+
+  // The evaluated resource 
+  string resource_id = 6
```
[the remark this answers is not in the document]
anatheka: It should be like this.

api/evaluation/evaluation.proto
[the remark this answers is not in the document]
anatheka: It shouldn't be so much, because we are only using the last assessment results for the evaluation. I think it is a good idea to add the not failing assessment results as well. I could imagine that an auditor would want to see this information.

internal/testutil/prototest/prototest.proto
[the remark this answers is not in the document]
anatheka: Maybe. I don't know anymore. But I don't want to write it in the proto.go. Do you have better ideas where to generate it?

service/evaluation/evaluation.go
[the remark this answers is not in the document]
lebogg: Exactly

api/evaluation/evaluation.proto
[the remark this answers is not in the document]
lebogg: But when we, e.g., would have a control requiring all VMs to "have automatic updates enabled" the evaluation result belongs to all VMs (=multiple resources), doesn't it?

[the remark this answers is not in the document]
anatheka: For now we remove the resource_id.

api/evaluation/evaluation.proto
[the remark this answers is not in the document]
anatheka: For now, we do not add the not failling assessment results. 

Check the list creation.

internal/testutil/prototest/prototest.proto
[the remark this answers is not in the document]
anatheka: Can be removed.

api/evaluation/evaluation.proto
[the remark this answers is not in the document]
oxisto: Ah it seems you are creating one evaluation result per resource, so we need to keep it for now.

service/evaluation/evaluation.go
```diff
@@ -0,0 +1,859 @@
...
+	"github.com/sirupsen/logrus"
+)
+
+var (
+	log                      *logrus.Entry
+	ErrCatalogIdIsMissing    = errors.New("catalog_id is missing")
+	ErrCategoryNameIsMissing = errors.New("category_name is missing")
+	ErrControlIdIsMissing    = errors.New("control_id is missing")
+	ErrControlNotAvailable   = errors.New("control not available")
+)
```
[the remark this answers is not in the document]
oxisto: We can do that separately for all errors if we want at some point

oxisto (approved):

PR #872, opened 2022-11-21
Update description for GetControl()

anatheka:
The description is not correct. Only if one hierarchy level exists for the control the metrics are returned. In case of two hierarchy levels for controls (e.g., control OPS-13 and subcontrols OPS-13.1, OPS-13.2, OPS-13.3) only the control OPS-13 with its subcontrols OPS-13.1, OPS-13.2, OPS-13.3 is returned .


anatheka:
TODO: generate proto files

anatheka:
> TODO: generate proto files

oxisto (approved):

PR #874, opened 2022-11-22
Fix `PermissionDenied` for `listAssessmentResults` in case of `AuthorizationStrategyAllowAll`

### LINKED ISSUE #873: ListAssessmentResult fails with filtered_cloud_service_id
**Describe the bug**
ListAssessmentResult does not filter by the Cloud Service ID.

Without filtered_cloud_service_id everything works
`grpcurl --plaintext -H "authorization: Bearer $TOKEN" localhost:9090 clouditor.Orchestrator.ListAssessmentResult`  assessmentResults

With filtered_cloud_service_id a PermissionDenied error returns
```
grpcurl --plaintext -d '{"filtered_cloud_service_id":"00000000-0000-0000-0000-000000000000"}' -H "authorization: Bearer $TOKEN" localhost:9090 clouditor.Orchestrator.ListAssessmentResults
ERROR:
  Code: PermissionDenied
  Message: access denied
```

lebogg:
Fixes #873.
In  `all, allowed = svc.authz.AllowedCloudServices(ctx)`, `all` is true and the list`allowed` nil if the strategy is `AuthorizationStrategyAllowAll`. And if `all` is true, the list shouldn't be checked. I adjusted the if condition accordingly and changed a bit the comments of `AuthorizationStrategyAllowAll` interface methods.

### DIFF
--- a/service/orchestrator/assessment_results.go
+++ b/service/orchestrator/assessment_results.go
@@ -45,11 +45,15 @@ func (svc *Service) ListAssessmentResults(ctx context.Context, req *assessment.L
 	var all bool
 
 	// Check, if the current filter is valid according to the authorization strategy. Omitting the cloud service ID is
-	// only allowed, if one can access all the cloud services. Furthermore, the content of the filtered cloud service ID
-	// must be in the list of allowed cloud service IDs.
+	// only allowed, if one can access *all* the cloud services.
 	all, allowed = svc.authz.AllowedCloudServices(ctx)
-	if !all && (req.FilteredCloudServiceId == "" ||
-		(req.FilteredCloudServiceId != "" && !slices.Contains(allowed, req.FilteredCloudServiceId))) {
+	if !all && req.FilteredCloudServiceId == "" {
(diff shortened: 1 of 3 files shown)


service/orchestrator/assessment_results.go
```diff
@@ -47,8 +48,8 @@ func (svc *Service) ListAssessmentResults(ctx context.Context, req *assessment.L
 	// only allowed, if one can access all the cloud services. Furthermore, the content of the filtered cloud service ID
 	// must be in the list of allowed cloud service IDs.
 	all, allowed = svc.authz.AllowedCloudServices(ctx)
-	if (req.FilteredCloudServiceId == "" && !all) ||
-		(req.FilteredCloudServiceId != "" && !slices.Contains(allowed, req.FilteredCloudServiceId)) {
+	if !all && (req.FilteredCloudServiceId == "" ||
```
oxisto, line 51: Should be solid now. Could we actually add more tests to this since we obviously did not catch this mistake with our current tests?

oxisto (approved):

PR #877, opened 2022-11-25
Add assessment result filter to ListAssessmentResult

anatheka:
This PR adds additional filter to ListAssessmentResult:
- compliant field in assessment_result
- metric_id field in assessment_result -> we should change that to a list of metric_ids


api/assessment/assessment.proto
```diff
@@ -70,8 +70,17 @@ service Assessment {
 };
 
 message ListAssessmentResultsRequest {
+  // Optionally, a filter for the cloud service id.  
   string filtered_cloud_service_id = 1;
 
+  enum Compliant {
```
oxisto, line 76: can't we just use an `optional bool` here?

api/orchestrator/orchestrator.proto
```diff
@@ -107,6 +107,10 @@ service Orchestrator {
         get : "/v1/orchestrator/cloud_services/{filtered_cloud_service_id}/"
               "assessment_results"
       }
+      additional_bindings {
```
oxisto, line 110: I think we don't need additional bindings here, this should automatically be part of the query parameters, correct, @lebogg?

api/assessment/assessment.proto
anatheka: I thought so too. But, you want either all assessment results or you want to filter for compliant=true or compliant=false. If you don't set the compliant field in the request it is automatically set to false and you cannot distinguish between compliant = false and the compliant field not set.

oxisto: But shouldn’t you get a `*bool` with optional?

api/orchestrator/orchestrator.proto
anatheka: Ok, that makes sense. But why is the additional_binding needed for filtered_cloud_service_id?

api/assessment/assessment.proto
anatheka: No, maybe I did something wrong :woman_shrugging:

oxisto: Seems to work:
```
optional bool filtered_compliant = 3;
```

generates 

```
FilteredCompliant *bool  `protobuf:"varint,3,opt,name=filtered_compliant,json=filteredCompliant,proto3,oneof" json:"filtered_compliant,omitempty"`
```

You may need to remove the additional binding (see below), otherwise I think protoc will complain about optional parameters in the query path.

api/orchestrator/orchestrator.proto
oxisto: I think its not needed but we liked the idea of associating it to the `/cloud_services` sub-resource path as well, but its not needed.

api/assessment/assessment.proto
anatheka: Ok, if you don't do it with additional bindings. But with the binding I've only got an bool. 

And why do we need the additional binding for the filtered_cloud_service_id? We also could use the query parameters or not?

api/orchestrator/orchestrator.proto
anatheka: Ok, that is confusing.

oxisto: I think we can remove all additional bindings and make all of the filtered_ properties optional then.

anatheka: Just done :)

service/orchestrator/assessment_results.go
```diff
@@ -34,41 +34,53 @@ import (
...
 	// unless one can access *all* the cloud services.
-	if !all && req.FilteredCloudServiceId != "" && !slices.Contains(allowed, req.FilteredCloudServiceId) {
+	if !all && util.Deref(req.FilteredCloudServiceId) != "" && !slices.Contains(allowed, util.Deref(req.FilteredCloudServiceId)) {
 		return nil, service.ErrPermissionDenied
 	}
 
 	for _, v := range values {
-		if req.FilteredCloudServiceId != "" && v.CloudServiceId != req.FilteredCloudServiceId {
+		// Check for filtered cloud service ID
+		if util.Deref(req.FilteredCloudServiceId) != "" && v.CloudServiceId != util.Deref(req.FilteredCloudServiceId) {
```
lebogg, line 63: The first check of the if clause means if `req.FilteredCloudServiceId` is empty (`""`), this will always break out the loop and, thus, no values are appended. 
But it should return all values then (the other way around).

Maybe add a test for this case

service/orchestrator/assessment_results.go
```diff
@@ -34,41 +34,53 @@ import (
...
 			continue
 		}
 
-		filtered = append(filtered, v)
+		// Check for filter compliant
+		if req.FilteredCompliant != nil && util.Deref(req.FilteredCompliant) != v.Compliant {
```
lebogg, line 68: The nil check is also done in `util.Deref`. Maybe it is redundant here?
I guess you want to prevent a check if the optional value is not set anyway but then we don't need the `Deref` here if you check it right before.

service/orchestrator/assessment_results.go
```diff
@@ -34,41 +34,53 @@ import (
...
+			continue
+		}
+
+		// Check for filter metric ID
+		if req.FilteredMetricId != nil && util.Deref(req.FilteredMetricId) != v.MetricId {
```
lebogg, line 73: The nil check is also done in `util.Deref`. Maybe it is redundant here?
I guess you want to prevent a check if the optional value is not set anyway but then we don't need the `Deref` here if you check it right before.

lebogg (changes requested):

service/orchestrator/assessment_results.go
anatheka: The values are appended if the FilteredCloudServiceId is empty and if the FilteredCloudServiceId is not empty the value must have the same CloudServiceID as the  FilteredCloudServiceId.  Otherwise the value is not appended. 

If the FilteredCloudServiceId is empty the value is appended because the if check fails. 

There are a lot of checks, which check do you still expect?

service/orchestrator/assessment_results.go
anatheka: There is no real nil check in Deref(), the pointer is just returned.

service/orchestrator/assessment_results.go
lebogg: Ok, my fault. I read it the wrong way :)

But like below it seems that we could get rid of the `util.Deref(req.FilteredCloudServiceId) != ""` check since we check if `req.FilteredCloudServiceId` and `v.CloudServiceId`  are the same anyway

Or we could do at least sth. like:
```
if filteredID := util.Deref(req.FilteredCloudServiceId); filteredID != "" && v.CloudServiceId != filteredID {
```

Or are there some edge cases, e.g., where a `cloudServiceID` is empty which it shouldn't be.

anatheka: No, we cannot get rid of the `util.Deref(req.FilteredCloudServiceId) != ""` check. FilteredCloudServiceId is an optional parameter and if we only check `req.FilteredCloudServiceId != v.CloudServiceId` it is always true if we do not set the optional parameter FilteredCloudServiceId and the assessment result is discarded.

Now I stored the value of the filteredCloudSeriveId in an extra variable: `filteredCloudServiceId := util.Deref(req.FilteredCloudServiceId)`. Is that better for you?

service/orchestrator/assessment_results.go
anatheka: See answer above.

anatheka:
I added the filter metricId to a list of metricIds, e.g., in the evaluation I need several metricIds and not only one metricId.

internal/testutil/prototest/prototest.proto
```diff
@@ -0,0 +1,44 @@
+// Copyright 2022 Fraunhofer AISEC
```
lebogg, line 1: Maybe get rid of that file and minimize the generated go file. But on the other side it is good if sth. changes with proto messages and their generations

internal/util/deref.go
```diff
@@ -34,3 +34,8 @@ func Deref[T any](p *T) T {
 
 	return result
 }
+
+// Ref references pointer values
+func Ref[T any](p T) *T {
+	return &p
+}
```
lebogg, line 41: Is this really needed? ^^

internal/testutil/prototest/prototest.proto
lebogg: But in general, I like the approach very much :)

anatheka: If I delete that file we have the same problem as before. In internal/util/struct_test.go we need a proto struct for the test. 

I see no problem in having this file and would let it as it is.

oxisto: If you basically just need "any" protobuf message for testing, you can actually use EmptyPb for example.

api/assessment/assessment.proto
```diff
@@ -70,7 +70,12 @@ service Assessment {
 };
 
 message ListAssessmentResultsRequest {
-  string filtered_cloud_service_id = 1;
+  // Optionally, a filter for the cloud service id.  
+  optional string filtered_cloud_service_id = 1;
```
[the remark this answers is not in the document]
anatheka: We do it also in the discovery with filtered_*. I will leave it like this and it can be changed in a separate PR.

internal/util/deref.go
anatheka: Yes, I need it as discussed in mattermost.

api/assessment/assessment.proto
```diff
@@ -70,7 +70,12 @@ service Assessment {
 };
 
 message ListAssessmentResultsRequest {
-  string filtered_cloud_service_id = 1;
+  // Optionally, a filter for the cloud service id.
```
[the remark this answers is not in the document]
anatheka: I didn't find any styleguide for it, but I changed it as suggested.

internal/testutil/prototest/prototest.proto
anatheka: You nee a protobuf message with field names, I don't think that an EmtyPb is enough!?!

service/orchestrator/assessment_results.go
```diff
@@ -27,48 +27,71 @@ package orchestrator
...
+		return nil, status.Errorf(codes.InvalidArgument, "%s", err)
+	}
+
+	filteredCloudServiceId := util.Deref(req.FilteredCloudServiceId)
+
 	// Check, if the current filter is valid according to the authorization strategy. Omitting the cloud service ID is
 	// only allowed, if one can access *all* the cloud services.
 	all, allowed = svc.authz.AllowedCloudServices(ctx)
-	if !all && req.FilteredCloudServiceId == "" {
+	if !all && filteredCloudServiceId == "" {
```
oxisto, line 62: Shouln't  we check for nil instead? And we can use the `GetXXX` functions of the protobuf message, this removes the need for the `Derek` call.

service/orchestrator/assessment_results.go
```diff
@@ -27,48 +27,71 @@ package orchestrator
...
 
 	for _, v := range values {
-		if req.FilteredCloudServiceId != "" && v.CloudServiceId != req.FilteredCloudServiceId {
+		// Check for filtered cloud service ID
+		if filteredCloudServiceId != "" && v.CloudServiceId != filteredCloudServiceId {
+			continue
+		}
+
+		// Check for filter compliant
+		if req.FilteredCompliant != nil && util.Deref(req.FilteredCompliant) != v.Compliant {
```
oxisto, line 79: Same as above, I would use req.GetFilteredCompliant() instead:
```suggestion
		if req.FilteredCompliant != nil && req.GetFilteredCompliant() != v.Compliant {
```

oxisto (commented):
Minor deref stuff

oxisto (approved):

PR #888, opened 2022-12-04
Switching to `buf` for protobuf generation

oxisto:
Instead of manually invoking `protoc`, we are switching to the `buf` library for protobuf generation. In the long run we might also checkout the buf registry and/or Buf Connect.


api/auth/auth.proto
```diff
@@ -26,7 +26,7 @@
  */
 syntax = "proto3";
 
-package clouditor;
+package clouditor.v1;
```
anatheka, line 29: Why not clouditor.auth.v1?

anatheka, line 29:
```suggestion
package clouditor.auth.v1;
```

anatheka (changes requested):
Only one minor thing  or was there a reason for the shortened path?

anatheka (approved):

PR #895, opened 2022-12-10
Add missing fields to metric configuration

anatheka:
The following fields are missing in the metric configuration and thus the validateion and storage of assessment result fails in the orchestrator: 
- updated_at
- metric_id
- cloud_service_id


service/assessment/assessment.go
```diff
@@ -375,29 +375,40 @@ func (svc *Service) handleEvidence(ev *evidence.Evidence, resourceId string) (er
 
 		log.Debugf("Evaluated evidence %v with metric '%v' as %v", ev.Id, metricID, data.Compliant)
 
-		targetValue := data.TargetValue
-
-		convertedTargetValue, err := convertTargetValue(targetValue)
+		// Get metric
+		metricConfiguration, err := svc.orchestratorClient.GetMetricConfiguration(context.Background(), &orchestrator.GetMetricConfigurationRequest{
```
oxisto, line 379: This should be unnecessary. The `data` field already contains the target value and operator that was actually used in the evaluation. The evaluation fetches the metric configuration from the orchestrator. 

The only thing that is missing from this is probably the cloud service ID (although this one is the same as in the evidence) and the "updated at". We could probably somehow loop the updated at also back to the `data` object.

service/orchestrator/metrics.go
```diff
@@ -121,6 +121,8 @@ func prepareMetric(m *assessment.Metric) (err error) {
 	}
 
 	config.IsDefault = true
+	config.UpdatedAt = timestamppb.Now()
```
oxisto, line 124: I think the idea was that a default metric configuration has no update at (nil) because in fact it was never "updated" (from the default).

service/assessment/assessment.go
oxisto: Here are the necessary lines to adjust: https://github.com/clouditor/clouditor/blob/0e86f5dff8ce965dfab9bb6d3c9b3ac818cdc9cb/policies/rego.go#L222-L224

service/assessment/assessment.go
```diff
@@ -390,14 +390,16 @@ func (svc *Service) handleEvidence(ev *evidence.Evidence, resourceId string) (er
 		result := &assessment.AssessmentResult{
 			Id:             uuid.NewString(),
 			Timestamp:      timestamppb.Now(),
-			CloudServiceId: ev.CloudServiceId,
+			CloudServiceId: ev.GetCloudServiceId(),
 			MetricId:       metricID,
 			MetricConfiguration: &assessment.MetricConfiguration{
-				TargetValue: convertedTargetValue,
-				Operator:    data.Operator,
+				TargetValue:    convertedTargetValue,
```
oxisto, line 396: UpdatedAt and is_default missing from `data`

policies/rego.go
```diff
@@ -220,8 +220,12 @@ func (re *regoEval) evalMap(baseDir string, serviceID, metricID string, m map[st
...
 
 		c := map[string]interface{}{
-			"target_value": config.TargetValue.AsInterface(),
-			"operator":     config.Operator,
+			"target_value":     config.TargetValue.AsInterface(),
+			"operator":         config.Operator,
+			"metric_id":        metricID,
+			"updated_at":       config.UpdatedAt,
+			"is_default":       config.IsDefault,
+			"cloud_service_id": serviceID,
```
oxisto, line 228: cloud service ID is redundant, already stored in the evidence.

policies/rego.go
```diff
@@ -220,8 +220,12 @@ func (re *regoEval) evalMap(baseDir string, serviceID, metricID string, m map[st
 		operators := fmt.Sprintf("%s/policies/operators.rego", baseDir)
 
 		c := map[string]interface{}{
-			"target_value": config.TargetValue.AsInterface(),
-			"operator":     config.Operator,
+			"target_value":     config.TargetValue.AsInterface(),
+			"operator":         config.Operator,
+			"metric_id":        metricID,
```
oxisto, line 225: also redundant?

service/assessment/assessment.go
anatheka: Data has only these fields, I think that was the reason why I used the GetMetricConfiguration()

`type Result struct {
	Applicable  bool
	Compliant   bool
	TargetValue interface{} `mapstructure:"target_value"`
	Operator    string
	MetricId    string
}`

oxisto: You can add the missing fields to the `Result` struct and you need to add the missing lines here to fill the result struct:
https://github.com/clouditor/clouditor/blob/338742c6d17779385e2829a917a39c87533c54d5/policies/rego.go#L254-L255

oxisto (approved):

PR #896, opened 2022-12-10
Refactor AssessEvidence response

anatheka:
Based on the nested methods AssessEvidence and AssessEvidences the response message is not correct for the AssessEvidence method. We have to create the response message only in the AssessEvidences method because we use a gRPC response stream there. If no gRPC response stream is used, only an error message is returned.


oxisto (approved):
Good idea. This always bothered me.

PR #902, opened 2022-12-12
Add more validation checks

anatheka:
We have to check if the structs are generated correctly. We do that by using the validate method in the test files for the following protobuf messages: 

- [x] evidence
- [x] assessment result
- [x] metric
- [x] metric configuration
- [x] metric implementation
- [x] certificate
- [x] state
- [x] user
- [x] cloud service
- [x] catalog
- [x] category
- [x] control
- [x] target of evaluation


anatheka:
We have defined variables for e.g., the cloud service id in several places. We should consolidate that.

oxisto (approved):

PR #922, opened 2023-01-12
Consolidation of test variables

anatheka:
This PR consolidates the test variables, e.g., CloudServiceId in one file. Futhermore, one file for creating objects is added (only the creation of user objects is added yet).


internal/testdata/testdatautils.go
```diff
@@ -0,0 +1,21 @@
+package testdata
+
+import "clouditor.io/clouditor/api/auth"
+
+func NewUser1() *auth.User {
```
oxisto, line 5: The api/user stuff is legacy and we should really remove this at some point (maybe now!)

anatheka: I think that is a task for you :wink:

oxisto: Done. can you double-check if that is ok? Otherwise we can merge it.

oxisto (approved):

PR #931, opened 2023-01-23
Add new Azure discoveries

anatheka:
This PR adds missing discovery: 
- [x] Microsoft Defender Information
- [x] Backup Object Storages
- [x] Backup Compute Disks
- [x] VM AutomaticUpdates
- [x] Application Gateway
- [x] L3 Firewall based on NSG


lebogg:
Shouldn't we hide new metrics/metric names ? :)

anatheka:
The panic error comes from "func (mockSender) Do(req *http.Request)" because no entry for "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/res1/providers/Microsoft.DataProtection/backupVaults/backupAccount1/backupInstances" exists in "func (m mockComputeSender) Do(req *http.Request) ". 

How to solve that? What does the Azure endpoint return?

anatheka:
@oxisto do you have an idea why the test fails here, but not locally?

oxisto:
> @oxisto do you have an idea why the test fails here, but not locally?

Running `go test -timeout 1m clouditor.io/clouditor/service/discovery/azure -json | tparse -follow` in a terminal also panics locally because of this:

```
time="2023-05-03T19:29:42+02:00" level=error msg="could not discover backup vaults: error getting next page: %!v(PANIC=Error method: runtime error: invalid memory address or nil pointer dereference)" component=azure-discovery
```

In Test_azureComputeDiscovery_List/With_resource_group

If I run it in vscode, it somehow passes and recovery the panic though.

oxisto:
> > @oxisto do you have an idea why the test fails here, but not locally?
> 
> Running `go test -timeout 1m clouditor.io/clouditor/service/discovery/azure -json | tparse -follow` in a terminal also panics locally because of this:
> 
> ```
> time="2023-05-03T19:29:42+02:00" level=error msg="could not discover backup vaults: error getting next page: %!v(PANIC=Error method: runtime error: invalid memory address or nil pointer dereference)" component=azure-discovery
> ```
> 
> In Test_azureComputeDiscovery_List/With_resource_group
> 
> If I run it in vscode, it somehow passes and recovery the panic though.

Seems to be caused by this here:

```
time="2023-05-03T19:34:56+02:00" level=error msg="Not handling mock for /subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.DataProtection/backupVaults yet" component=azure-discovery
```

oxisto:
> > > @oxisto do you have an idea why the test fails here, but not locally?
> > 
> > 
> > Running `go test -timeout 1m clouditor.io/clouditor/service/discovery/azure -json | tparse -follow` in a terminal also panics locally because of this:
> > ```
> > time="2023-05-03T19:29:42+02:00" level=error msg="could not discover backup vaults: error getting next page: %!v(PANIC=Error method: runtime error: invalid memory address or nil pointer dereference)" component=azure-discovery
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In Test_azureComputeDiscovery_List/With_resource_group
> > If I run it in vscode, it somehow passes and recovery the panic though.
> 
> Seems to be caused by this here:
> 
> ```
> time="2023-05-03T19:34:56+02:00" level=error msg="Not handling mock for /subscriptions/00000000-0000-0000-0000-000000000000/providers/Microsoft.DataProtection/backupVaults yet" component=azure-discovery
> ```

Actually it seems to crash in the "Discovery error" test case, because this intentionally sets a wrong "sender" (the network mock sender) instead of compute. This one doesn't have the mocks for backupVaults and that's why it crashed. Not sure exactly where the issue is because it is perfectly ok for the network mock sender NOT to have the backup vaults endpoint and it should just return a 404, not sure why it panics here.

oxisto:
Ok I found it why .... this is caused by the `ResponseError` struct inside the azure SDK, which assumes that certain structs are there to fill out an error message, which are not there in our mock. I will fix this :)

service/discovery/azure/storage.go
```diff
@@ -386,6 +416,46 @@ func accountName(id string) string {
 	return splitName[8]
 }
 
+// idUpToStorageAccount returns the resource ID cutting of the storage type information, e.g., '/subscriptions/XXXXXXXXXXXX-XXXX-XXXX-XXXXXXXXXXXX/resourceGroups/resourceGroupName/providers/Microsoft.Storage/storageAccounts/containerName'
+func idUpToStorageAccount(id string) string {
```
oxisto, line 420: what's the point of this function? this confuses me.

service/discovery/azure/storage.go
```diff
@@ -107,8 +138,14 @@ func (d *azureStorageDiscovery) discoverStorageAccounts() ([]voc.IsCloudResource
 		return nil, err
 	}
 
+	// Discover backup vaults
+	err := d.azureDiscovery.discoverBackupVaults()
```
oxisto, line 142: This is intentionally called in the storage AND in the compute discovery, correct? Do we have a problem that the discovered vaults are then created twice?

service/discovery/azure/storage.go
anatheka: The ID is in the format `/subscriptions/XXXXXXXXXXXX-XXXX-XXXX-XXXXXXXXXXXX/resourceGroups/resourceGroupName/providers/Microsoft.Storage/storageAccounts/containerName/blobServices/default/containers/testContainer` but we need the ID as follows `/subscriptions/XXXXXXXXXXXX-XXXX-XXXX-XXXXXXXXXXXX/resourceGroups/resourceGroupName/providers/Microsoft.Storage/storageAccounts/containerName`. 

The function cuts of everything behind "containerName".

oxisto: Hm ok in the function `handleObjectStorage` you could / or should just use `account.ID` instead of `container.ID` to get the correct one. But in the vault API it seems like you only get the container ID.

Then I would suggest to at least replace the last line with `return "/" + strings.Join(split[1:9], "/")` and rename the function to something like `accountIDFromContainerID` or something so that one knows that this returns the account resource ID from a container ID.

service/discovery/azure/storage.go
oxisto: Ok I guess that's ok-ish for now.

service/discovery/azure/storage.go
anatheka: You're absolutely right. To be honest, I have no idea what that leftover is. I removed it.

oxisto (approved):

PR #937, opened 2023-01-25
Fix method responses

anatheka:
This PR changes the responses for the following methods:
- [x] AssessEvidence(s)
- [x] StoreAssessmentResult(s)
- [x] StoreEvidence(s)

All methods are custom methods and require a response message according to the style convention. The unary RPCs can be empty and the bidirectional streaming RPCs need the status and its message for error handling.


service/orchestrator/orchestrator.go
```diff
@@ -218,7 +218,9 @@ func (s *Service) StoreAssessmentResult(_ context.Context, req *orchestrator.Sto
 
 	go s.informHook(req.Result, nil)
 
-	resp = &orchestrator.StoreAssessmentResultResponse{}
+	resp = &orchestrator.StoreAssessmentResultResponse{
```
oxisto, line 221: We could also change the return type of the function to `Empty`?

api/orchestrator/orchestrator.proto
```diff
@@ -470,7 +470,10 @@ message StoreAssessmentResultRequest {
   clouditor.assessment.v1.AssessmentResult result = 1
       [ (validate.rules).message.required = true ];
 }
-message StoreAssessmentResultResponse {
+
+message StoreAssessmentResultResponse {}
```
oxisto, line 474: It would be good to have a small comment here, why this is empty and why the ResultsResponse is not empt

api/assessment/assessment.proto
```diff
@@ -99,7 +99,9 @@ message AssessEvidenceRequest {
   clouditor.evidence.v1.Evidence evidence = 1
       [ (validate.rules).message.required = true ];
 }
-message AssessEvidenceResponse {
+message AssessEvidenceResponse {}
```
oxisto, line 102: same as below

service/evidence/evidence_store.go
```diff
@@ -119,9 +112,18 @@ func (s *Service) StoreEvidences(stream evidence.EvidenceStore_StoreEvidencesSer
 		evidenceRequest := &evidence.StoreEvidenceRequest{
 			Evidence: req.Evidence,
 		}
-		res, err = s.StoreEvidence(context.Background(), evidenceRequest)
+		_, err = s.StoreEvidence(context.Background(), evidenceRequest)
 		if err != nil {
 			log.Errorf("Error storing evidence: %v", err)
+			// Create response message. The AssessEvidence method does not need that message, so we have to create it here for the stream response.
```
anatheka, line 118:
```suggestion
			// Create response message. The StoreEvidence method does not need that message, so we have to create it here for the stream response.
```

service/assessment/assessment.go
```diff
@@ -253,16 +255,14 @@ func (svc *Service) AssessEvidence(_ context.Context, req *assessment.AssessEvid
 		return res, status.Errorf(codes.Internal, "%v", err)
```
oxisto, line 255: Shouldn't we return `nil, status.Errorf` here?

service/evidence/evidence_store.go
```diff
@@ -75,21 +77,12 @@ func (s *Service) StoreEvidence(_ context.Context, req *evidence.StoreEvidenceRe
 
 		go s.informHooks(nil, newError)
 
-		resp = &evidence.StoreEvidenceResponse{
-			Status:        false,
-			StatusMessage: newError.Error(),
-		}
-
 		return resp, status.Errorf(codes.InvalidArgument, "%v", newError)
```
oxisto, line 80: here as well

service/assessment/assessment.go
oxisto: Also ValidateRequest already returns a status.Error() so we can just `return nil, err` here. otherwise we double-wrap the error.

service/assessment/assessment.go
```diff
@@ -250,19 +250,17 @@ func (svc *Service) AssessEvidence(_ context.Context, req *assessment.AssessEvid
...
 		err = fmt.Errorf("error while handling evidence: %v", err)
 		log.Error(err)
-		return res, status.Errorf(codes.Internal, "%v", err)
+		return nil, status.Errorf(codes.Internal, "%v", err)
 	}
 
-	res = &assessment.AssessEvidenceResponse{}
-
-	return res, nil
+	return nil, nil
```
oxisto, line 256: Shouldn't we return the empty response here? This was ok I think. Or is this automatically converted into an "empty" AssessEvidenceResponse?

anatheka: That is a good question. I thought that since we don't need the AssessEvidenceResponse, we might as well set it to nil. But you are right, maybe it is better to return an empty response for future changes. Otherwise we have to change the tests again.

oxisto (approved):

PR #940, opened 2023-01-28
Store `AssessmentResult` in persistence layer

oxisto:
This PR stores the `AssessmentResult` objects in the orchestrator into the persistent storage. It also fixes remaining authorisation checks in the assessment storage / retrieval layer.

The assessment results in the assessment services are intentionally left in-memory. We might actually remove the "storage" of them altogether. They are basically only used as a debug feature in case you want to look up results that somehow do not end up in the orchestrator. Should we remove them? @anatheka


service/assessment/assessment.go
```diff
@@ -96,7 +96,9 @@ type Service struct {
 	// hookMutex is used for (un)locking result hook calls
 	hookMutex sync.RWMutex
 
-	// Currently, results are just stored as a map (=in-memory). In the future, we will use a DB.
+	// results will contain all the assessment results of the current run. They will NOT be persisted to the storage,
```
anatheka, line 99:
```suggestion
	// Results will contain all the assessment results of the current run. They will NOT be persisted to the storage,
```

anatheka:
> This PR stores the `AssessmentResult` objects in the orchestrator into the persistent storage. It also fixes remaining authorisation checks in the assessment storage / retrieval layer.
> 
> The assessment results in the assessment services are intentionally left in-memory. We might actually remove the "storage" of them altogether. They are basically only used as a debug feature in case you want to look up results that somehow do not end up in the orchestrator. Should we remove them? @anatheka

I think it is still a general problem, how to make sure that all assessment results end up in the DB. Maybe until we have solved this, we should leave the results stored in memory.

anatheka (approved):
LGTM after fixing the conflict.

- Shouldn't  we move the assessment_results.go to service/assessment/?

service/assessment/assessment.go
anatheka: What I forgot: I think you have to update the engine as well.

oxisto: results is unexported -> lower case

PR #945, opened 2023-01-31
Fix CreateTargetOfEvaluation

anatheka:
This PR adds a page size of 2000 to the ListControls() in CreateTargetOfEvaluation. Otherwise we get only 50 controls instead of all contols. 

@oxisto do you know of another option to get "all" the controls? If I understand it correctly we need for the ListAllPaginated() a client we do not have.

### DIFF
--- a/service/orchestrator/toe.go
+++ b/service/orchestrator/toe.go
@@ -59,7 +59,10 @@ func (svc *Service) CreateTargetOfEvaluation(ctx context.Context, req *orchestra
 
 	// Certain catalogs do not allow scoping, in this case we need to pre-populate all controls into the scope.
 	if c.AllInScope {
-		lcres, err = svc.ListControls(ctx, &orchestrator.ListControlsRequest{CatalogId: req.TargetOfEvaluation.CatalogId})
+		lcres, err = svc.ListControls(ctx, &orchestrator.ListControlsRequest{
+			CatalogId: req.TargetOfEvaluation.CatalogId,
+			PageSize:  2000,
+		})
 		if err != nil {
 			// The error is already a gRPC error, so we can just return it
 			return nil, err
--- a/service/orchestrator/toe.go
+++ b/service/orchestrator/toe.go
@@ -29,10 +29,12 @@ import (
 	"context"
 	"errors"
 
+	"clouditor.io/clouditor/api"
 	"clouditor.io/clouditor/api/orchestrator"
 	"clouditor.io/clouditor/persistence"
 	"clouditor.io/clouditor/persistence/gorm"
 	"clouditor.io/clouditor/service"
+	"google.golang.org/grpc"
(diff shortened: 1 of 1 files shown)


oxisto:
> This PR adds a page size of 2000 to the ListControls() in CreateTargetOfEvaluation. Otherwise we get only 50 controls instead of all contols.
> 
> @oxisto do you know of another option to get "all" the controls? If I understand it correctly we need for the ListAllPaginated() a client we do not have.

```go
controls, err = api.ListAllPaginated(&orchestrator.ListControlsRequest{}, svc.ListControls, func(res *orchestrator.ListControlsResponse) []*orchestrator.Control {
			return res.Controls
		})
```

oxisto:
> > This PR adds a page size of 2000 to the ListControls() in CreateTargetOfEvaluation. Otherwise we get only 50 controls instead of all contols.
> > @oxisto do you know of another option to get "all" the controls? If I understand it correctly we need for the ListAllPaginated() a client we do not have.
> 
> You can use `ListAllPaginated` on everything that implements the interface (either the client or the service itself). It's basically magic.
> 
> ```go
> controls, err = api.ListAllPaginated(&orchestrator.ListControlsRequest{}, svc.ListControls, func(res *orchestrator.ListControlsResponse) []*orchestrator.Control {
>    return res.Controls
> })
> ```

Ah... my mistake. It won't work since the client interface also has the additional gRPC options... Ok let me check if I can fix that somehow

oxisto:
> > > This PR adds a page size of 2000 to the ListControls() in CreateTargetOfEvaluation. Otherwise we get only 50 controls instead of all contols.
> > > @oxisto do you know of another option to get "all" the controls? If I understand it correctly we need for the ListAllPaginated() a client we do not have.
> > 
> > 
> > You can use `ListAllPaginated` on everything that implements the interface (either the client or the service itself). It's basically magic.
> > ```go
> > controls, err = api.ListAllPaginated(&orchestrator.ListControlsRequest{}, svc.ListControls, func(res *orchestrator.ListControlsResponse) []*orchestrator.Control {
> >    return res.Controls
> > })
> > ```
> 
> Ah... my mistake. It won't work since the client interface also has the additional gRPC options... Ok let me check if I can fix that somehow

Got it. It's very hacky but should work. Basically wrapping the server function in a lambda function that looks like the client function (and just ignores the grpc call options):
```go
controls, err = api.ListAllPaginated(&orchestrator.ListControlsRequest{}, func(ctx context.Context, req *orchestrator.ListControlsRequest, opts ...grpc.CallOption) (*orchestrator.ListControlsResponse, error) {
	return svc.ListControls(ctx, req)
}, func(res *orchestrator.ListControlsResponse) []*orchestrator.Control {
	return res.Controls
})
```

oxisto (approved):

PR #954, opened 2023-02-03
Add metadata endpoint

### LINKED ISSUE #946: Provide a API endpoint for various metadata of the running engine
This may include

- The version of Clouditor (tricky to get if we are not running as a dependency)
- List of plugins (future?)

anatheka:
This PR adds an metadata endpoint for various metadata information. 
Closes #946


anatheka:
How to include ldflags in vscode?

.github/workflows/build.yml
```diff
@@ -36,7 +38,7 @@ jobs:
       - name: Build
         run: |
           mkdir -p build
-          go build -o build ./...
+          go build -ldflags="-X 'main.version=$(git describe --tags --abbrev=0)'" -o build ./...
```
oxisto, line 41: We probably don't need to do that here but in the Dockerfile that is actually being built for a specific release.

api/orchestrator/orchestrator.proto
```diff
@@ -949,6 +963,17 @@ message RemoveCertificateRequest {
   string certificate_id = 1 [ (validate.rules).string.min_len = 1 ];
 }
 
+message Metadata {
+  string clouditor_version = 1;
+  string golang_version = 2;
+  repeated Dependency dependency = 3;
```
oxisto, line 969: dependencies?

cmd/engine/engine.go
```diff
@@ -124,6 +124,9 @@ var (
 	providers            []string
 
 	log *logrus.Entry
+
+	// clouditorVersion is used for the clouditorVersion, linked during the build process
+	clouditorVersion string
```
oxisto, line 129: would it make sense to check if this is empty and if yes, then use the git commit sha info from the runtime.BuildInfo?

service/orchestrator/orchestrator.go
```diff
@@ -140,13 +145,24 @@ func WithAuthorizationStrategyJWT(key string) ServiceOption {
 	}
 }
 
+// WithClouditorVersion is an option to set the Clouditors last release version.
```
oxisto, line 148:
```suggestion
// WithClouditorVersion is an option to set the Clouditor's last release version.
```

api/orchestrator/orchestrator.proto
```diff
@@ -39,6 +39,14 @@ option go_package = "clouditor.io/clouditor/api/orchestrator";
 
 // Manages the orchestration of components within the Clouditor architecture
 service Orchestrator {
+  // Get Metadata
+  rpc Metadata(MetadataRequest)
+  returns (MetadataResponse) {
+    option (google.api.http) = {
+      get: "/v1/orchestrator/metadata"
```
oxisto, line 46: maybe something related to "runtime" instead of metadata?

oxisto (commented):
Some initial thoughts

cmd/engine/engine.go
anatheka: I don't know. We can do that. But I think if we do not get the tag info, we probably won't get the commit hash either. Except when we forget to tag.

api/orchestrator/orchestrator.proto
anatheka: For me ok.

.github/workflows/build.yml
anatheka: That's right. But the dockerfile is not ready yet :)

cmd/engine/engine.go
oxisto: You always get the the commit hash in `vcs.revision` in https://pkg.go.dev/runtime/debug#BuildSetting

oxisto: Ok this is beyond stupid. If you issue `go build cmd/engine/engine.go` from the root, `vcs.revision` will be empty. If you issue `cd cmd/engine && go build` it correctly sets the revision. Somehow this is documented in https://pkg.go.dev/cmd/go@go1.18#hdr-Relative_import_paths

.github/workflows/build.yml
anatheka: I did this for future testing, but it's not necessary either.

cmd/engine/engine.go
oxisto: Some additional issues to watch: https://github.com/golang/go/issues/50603

oxisto: `go build ./cmd/engine` also works to produce the git hash, which apparently should be the "correct" way of doing it anyway. 

In this case we still need to inject the official version via `ldflags` and then merge the information at some point.

api/orchestrator/orchestrator.proto
anatheka: Ah, you meant plural instead of singular?

oxisto: Yes

cmd/engine/engine.go
anatheka: Getting the information from git is not that easy. Maybe we should just create a file with all the important information during the release creation and commits via Github Actions.

api/orchestrator/orchestrator.proto
```diff
@@ -39,6 +39,14 @@ option go_package = "clouditor.io/clouditor/api/orchestrator";
 
 // Manages the orchestration of components within the Clouditor architecture
 service Orchestrator {
+  // Get Runtime
+  rpc Runtime(RuntimeRequest)
```
oxisto, line 43: GetRuntime would be a fitting request I would say. The response could then be actually named `Runtime` without any wrapping (since we allow this for `Get` calls)

api/orchestrator/orchestrator.proto
```diff
@@ -949,6 +963,28 @@ message RemoveCertificateRequest {
   string certificate_id = 1 [ (validate.rules).string.min_len = 1 ];
 }
 
+message Runtime {
+  // release_version is the latest Clouditor release version for this commit
+  string release_version = 1;
```
oxisto, line 968: should this be set to optional, since this is only used for release versions? I guess in non-release versions this is empty?

api/orchestrator/orchestrator.proto
```diff
@@ -949,6 +963,28 @@ message RemoveCertificateRequest {
...
+  // vcs is the used version control system
+  string vcs = 2;
+
+  // commit_hash is the current Clouditor commit hash
+  string commit_hash = 3;
+
+  // golang_version is the used golang version
+  string golang_version = 4;
+
+  // dependency is a list of used runtime dependencies
```
oxisto, line 979:
```suggestion
  // dependencies is a list of used runtime dependencies
```

cmd/engine/engine.go
```diff
@@ -214,7 +217,9 @@ func doCmd(_ *cobra.Command, _ []string) (err error) {
  $$ |      $$ |$$ |  $$ |$$ |  $$ |$$ |  $$ |$$ |  $$ |$$\ $$ |  $$ |$$ |
  \$$$$$$\  $$ |\$$$$$   |\$$$$$   |\$$$$$$  |$$ |  \$$$   |\$$$$$   |$$ |
   \_______|\__| \______/  \______/  \_______|\__|   \____/  \______/ \__|
- `)
+ 
+  Version %s\n
+  `, version)
```
oxisto, line 222: can we somehow set a default value in case this is not linked (and not a release) to something like (devel)? Or even better construct a "fake" version out of the git hash?

oxisto:
Can you have a look if this looks good, @anatheka ?

api/runtime/runtime.go
```diff
@@ -0,0 +1,39 @@
...
+// \$$$$$$\  $$ |\$$$$$   |\$$$$$   |\$$$$$$  |$$ |  \$$$   |\$$$$$   |$$ |
+//  \_______|\__| \______/  \______/  \_______|\__|   \____/  \______/ \__|
+//
+// This file is part of Clouditor Community Edition.
+
+package runtime
+
+import "fmt"
+
+// VersionString returns a version, if this is a release version, or a pseudo-version using a git commit hash otherwise.
```
anatheka, line 30:
```suggestion
// VersionString returns a version. If it is a release version the release version returned, otherwise a pseudo version with a git commit hash or v0.0.0 if no commit hash is known.
```

service/runtime.go
```diff
@@ -0,0 +1,95 @@
...
+// populateRuntimeInfo populates the runtime info using the build info and other sources.
+func populateRuntimeInfo() {
+	if populated {
+		return
+	}
+
+	// Set build info
+	buildInfo, ok := debug.ReadBuildInfo()
+	if !ok {
+		log.Errorf("Could not read build info. Runtime information will not be complete")
```
anatheka, line 56:
```suggestion
		log.Errorf("Could not read build info. Runtime information will not be complete.")
```

anatheka (commented):
Looks good, just a minor comments

oxisto (approved):

PR #956, opened 2023-02-06
Add assurance level property for control

anatheka:
First draft for the assurance level.
For now, there is a workaround as long as the the read EUCS catalog has not the right naming with B, S, or H at the end, e.g., "OPS-01.1S".

Futhermore, the UI has to be updated, the related PR can be found here clouditor/clouditor-ui#317


service/orchestrator/toe.go
```diff
@@ -109,14 +168,23 @@ func (svc *Service) GetTargetOfEvaluation(_ context.Context, req *orchestrator.G
...
 	if err != nil {
 		return nil, err
 	}
 
+	// TODO: authz
+	if req.GetCloudServiceId() != "" {
+		conds = append(conds, "cloud_service_id = ?", req.CloudServiceId)
+	}
+
+	// TODO: filter according to catalog (and acually make it mutually exclusive)
```
oxisto, line 184: Can we achieve this somehow with the validation framework?

service/orchestrator/toe.go
```diff
@@ -206,8 +275,10 @@ func (svc *Service) ListControlsInScope(ctx context.Context, req *orchestrator.L
 		return nil, service.ErrPermissionDenied
 	}
 
+	conds = append(conds, "target_of_evaluation_cloud_service_id = ? AND target_of_evaluation_catalog_id = ?", req.CloudServiceId, req.CatalogId)
```
oxisto, line 278: Need to add authz here as well. Something like:
```
// Check, if this request has access to the cloud service according to our authorization strategy.
	if !s.authz.CheckAccess(ctx, service.AccessRead, req) {
	}
```

and then make sure that ListControlsInScopeRequest implements `GetCloudServiceId` in api/orchestrator/orchestrator.go

anatheka: This method already has the check. See line 274

anatheka: Do we want to implement all auth checks here or in another PR?

service/orchestrator/toe.go
anatheka: Do we have to check that really? Either the binding `/v1/orchestrator/cloud_services/{cloud_service_id}/toes` with the cloud_service_id or `/v1/orchestrator/catalogs/{catalog_id}/toes` with the catalog_id is used. I would check if one of both is set and define the conds accordingly.

oxisto: That's true, and should probably be sufficient. I just wanted to make sure that an `invalid-argument` error is created if someone is issuing the gRPC call directly with neither the cloud service nor the catalog.

oxisto: Maybe it even makes sense to also have something like `/v1/orchestrator/toes` which displays all toes (currently visible to the session), mainly for debugging reasons, etc.

service/orchestrator/toe.go
oxisto: > Do we want to implement all auth checks here or in another PR?

Might as well fix it here for those methods we touch. I think may comment might have been to another function. I am pretty certain that one of them was still missing the check.

anatheka: Ok, I will check that.

service/orchestrator/toe.go
anatheka: We have the `/v1/orchestrator/toes` already: 
 ```
// Lists all Targets of Evaluation
  rpc ListTargetsOfEvaluation(ListTargetsOfEvaluationRequest)
      returns (ListTargetsOfEvaluationResponse) {
    option (google.api.http) = {
      get : "/v1/orchestrator/toes"
      additional_bindings {
        get : "/v1/orchestrator/cloud_services/{cloud_service_id}/toes"
      }
      additional_bindings {
        get : "/v1/orchestrator/catalogs/{catalog_id}/toes"
      }
    };
  }
```

I am not sure what you exactly mean.

Comment (anatheka) [reply service/orchestrator/toe.go]: I've added the auth check to `ListTargetsOfEvaluation` and `CreateTargetOfEvaluationRequest`.

File: service/orchestrator/toe.go (hunk 2 above)
Comment (anatheka) [reply service/orchestrator/toe.go:288]: I've added the auth check to `ListTargetsOfEvaluation` and `CreateTargetOfEvaluationRequest`.

File: service/orchestrator/toe.go (hunk 1 above)
Comment (oxisto) [reply service/orchestrator/toe.go]: Ok, I was getting confused apparently

File: service/orchestrator/toe.go (hunk 2 above)
Comment (oxisto) [reply service/orchestrator/toe.go:237]: I took the liberty in adding the remaining authz checks for all functions in the `toe.go` while we are changing this file anyway. Looks good from my side. one last check from your side @anatheka ?

Comment (anatheka) [reply service/orchestrator/toe.go:237]: Perfect, from my side it also fits.

Comment (anatheka) [reply service/orchestrator/toe.go:237]: Maybe you can also have a look at the UI PR clouditor/clouditor-ui#317

File: api/orchestrator/orchestrator.proto [hunk 3]
```diff
@@ -645,13 +645,28 @@ message Catalog {
...
+    (validate.rules).repeated.items.message.required = true
   ];
 
   // Certain security catalogs do not allow to select the scope of the controls,
   // but all controls are automatically "in scope", however they can be set to a
   // DELEGATED status.
   bool all_in_scope = 5;
+
+  // Certain security catalogs allow the differentiation of their controls according to an assurance level. If the property assurance_level is set to true, the catalog is using the assurance level.
+  bool assurance_level = 6;
```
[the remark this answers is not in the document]
anatheka: I think we can also check if assurance_levels is empty.

oxisto (approved):

PR #969, opened 2023-02-15
More debugging info for Store*/Create*/Update*/Remove* methods

anatheka:


service/orchestrator/catalogs.go
```diff
@@ -208,6 +215,8 @@ func (svc *Service) loadCatalogs() (err error) {
 		log.Errorf("Error while saving catalog %v", err)
 	}
 
+	log.Debugf("Catalog loaded  with name '%s' and id '%s'.", catalogs[0].GetName(), catalogs[0].GetId())
```
anatheka, line 218:
```suggestion
	log.Debugf("Catalog loaded with name '%s' and id '%s'.", catalogs[0].GetName(), catalogs[0].GetId())
```

service/orchestrator/cloud_service.go
```diff
@@ -158,6 +161,9 @@ func (s *Service) UpdateCloudService(ctx context.Context, req *orchestrator.Upda
 	}
 
 	go s.informHooks(ctx, response, nil)
+
+	log.Debugf("Cloud Service updated with name %s' and id '%s'.", req.CloudService.GetName(), req.CloudService.GetId())
```
anatheka, line 165:
```suggestion
	log.Debugf("Cloud Service updated with name '%s' and id '%s'.", req.CloudService.GetName(), req.CloudService.GetId())
```

service/orchestrator/cloud_service.go
```diff
@@ -67,6 +67,9 @@ func (s *Service) RegisterCloudService(ctx context.Context, req *orchestrator.Re
 	}
 
 	go s.informHooks(ctx, res, nil)
+
+	log.Debugf("Cloud Service registered with name %s'.", req.CloudService.GetName())
```
anatheka, line 71:
```suggestion
	log.Debugf("Cloud Service registered with name '%s'.", req.CloudService.GetName())
```

service/orchestrator/catalogs.go
```diff
@@ -208,6 +215,8 @@ func (svc *Service) loadCatalogs() (err error) {
 		log.Errorf("Error while saving catalog %v", err)
 	}
 
+	log.Debugf("Catalog loaded with name '%s' and id '%s'.", catalogs[0].GetName(), catalogs[0].GetId())
```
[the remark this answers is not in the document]
anatheka: Sure, that would be nice. But I just wanted a quick fix :smile: 
I will have a look at that.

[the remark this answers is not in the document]
anatheka: @oxisto does that make sense?

logging/logging.go
```diff
@@ -0,0 +1,81 @@
...
+		message = fmt.Sprintf("%s %s", req.GetType(), operation)
+	}
+
+	if req.GetCloudServiceId() != "" && len(params) > 0 {
+		message = fmt.Sprintf("%s for Cloud Service '%s' and Catalog '%s'", message, req.GetCloudServiceId(), params[0])
+	} else if req.GetCloudServiceId() != "" {
+		message = fmt.Sprintf("%s for Cloud Service '%s'", message, req.GetCloudServiceId())
+	}
+
+	switch loglevel {
```
anatheka, line 73: Do we need the distinction or do we just want to use debug?

logging/logging.go
```diff
@@ -0,0 +1,81 @@
...
+	case LoglevelDebug:
+		log.Debugf("%s.", message)
+	case LoglevelInfo:
+		log.Infof("%s.", message)
+	case LoglevelError:
+		log.Errorf("%s.", message)
+	}
```
anatheka, line 80: Do we need the distinction or do we just want to use debug?

service/orchestrator/catalogs.go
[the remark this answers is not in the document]
oxisto: I will quickly hack together an "alternative", which uses protorefect and which we are already using in the streams part of the API. I will do a separate PR with this branch as a base and then we can see which approach makes most sense. In either case we should consolidate both functionalities.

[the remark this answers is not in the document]
oxisto: Ok I was not really happy about this solution either, I am talking about this function here: https://github.com/clouditor/clouditor/blob/c641879888128bfcce3cc85da3b2a9910b759d90/api/streams.go#L256-L262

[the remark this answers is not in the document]
anatheka: I have not even thought about the function. Does it not work properly?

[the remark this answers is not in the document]
oxisto: It works perfectly, it only uses protobuf reflection, which can be a little slow and should probably only used when debug logging is activated.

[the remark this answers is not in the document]
oxisto: My suggestion: let's just revert back to the hard-coded log messages for now and revisit this some time in the future.

logging/logging.go
oxisto: The distinction could be fine, however I would suggest switching on the logrus log levels (https://pkg.go.dev/github.com/sirupsen/logrus#Level) rather than introducing our own constants.

api/orchestrator/orchestrator.go
```diff
@@ -62,6 +63,12 @@ type CloudServiceRequest interface {
 	proto.Message
 }
 
+type LogRequest interface {
+	GetId() string
```
oxisto, line 67: Maybe "target ID" / payload ID oder something similar? Its not really the ID of the request but rather of its payload/the target.

api/orchestrator/orchestrator.go
```diff
@@ -62,6 +63,12 @@ type CloudServiceRequest interface {
...
+	GetType() string
+	GetCloudServiceId() string
```
oxisto, line 69: I would omit that here and check if the request fullfills `CloudServiceRequest` in the log message code.

logging/logging.go
```diff
@@ -0,0 +1,81 @@
...
+	LoglevelInfo  = "info"
+	LoglevelError = "error"
+	Create        = "created"
+	Remove        = "removed"
+	Update        = "updated"
+	Register      = "registered"
+	Load          = "loaded"
+)
+
+// LogMessage creates a logging message with the given parameters
```
oxisto, line 46: I would name this `LogRequest`

logging/logging.go
```diff
@@ -0,0 +1,81 @@
...
+//  $$$$$$$\ $$ | $$$$$$\  $$\   $$\  $$$$$$$ |$$\ $$$$$$\    $$$$$$\   $$$$$$\
+// $$  _____|$$ |$$  __$$\ $$ |  $$ |$$  __$$ |$$ |\_$$  _|  $$  __$$\ $$  __$$\
+// $$ /      $$ |$$ /  $$ |$$ |  $$ |$$ /  $$ |$$ |  $$ |    $$ /  $$ |$$ | \__|
+// $$ |      $$ |$$ |  $$ |$$ |  $$ |$$ |  $$ |$$ |  $$ |$$\ $$ |  $$ |$$ |
+// \$$$$$$\  $$ |\$$$$$   |\$$$$$   |\$$$$$$  |$$ |  \$$$   |\$$$$$   |$$ |
+//  \_______|\__| \______/  \______/  \_______|\__|   \____/  \______/ \__|
+//
+// This file is part of Clouditor Community Edition.
+
+package logging
```
oxisto, line 26: Please move this whole package to internal, this should really only be used by Clouditor itself

logging/logging.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"fmt"
+
+	"clouditor.io/clouditor/api/orchestrator"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	LoglevelDebug = "debug"
```
oxisto, line 36: See below, we should remove those and use logrus log level instead.

logging/logging.go
```diff
@@ -0,0 +1,81 @@
...
+		message string
+	)
+
+	if req.GetId() != "" {
+		message = fmt.Sprintf("%s %s with ID '%s'", req.GetType(), operation, req.GetId())
+	} else {
+		message = fmt.Sprintf("%s %s", req.GetType(), operation)
+	}
+
+	if req.GetCloudServiceId() != "" && len(params) > 0 {
```
oxisto, line 67: Instead of this string comparison you check if the request is a `CloudServiceRequest´:

```go
csreq, ok := req.(CloudServiceRequest)
if ok && len(params) > 0 {
 /* this with csreq.GetCloudServiceId() */
} else if ok {
 /* that with csreq.GetCloudServiceId() */
} 
```

logging/logging.go
```diff
@@ -0,0 +1,81 @@
...
+	"clouditor.io/clouditor/api/orchestrator"
+	"github.com/sirupsen/logrus"
+)
+
+const (
+	LoglevelDebug = "debug"
+	LoglevelInfo  = "info"
+	LoglevelError = "error"
+	Create        = "created"
+	Remove        = "removed"
```
oxisto, line 40: There are already constants for Read, Update, Delete, etc. in the `api` (or service?) package related to the authorisation strategy. We can/should re-use them instead of new ones I think.

In general, I would avoid string-type constants, you can just use Go enums.

logging/logging.go
```diff
@@ -0,0 +1,81 @@
...
+	)
+
+	if req.GetId() != "" {
+		message = fmt.Sprintf("%s %s with ID '%s'", req.GetType(), operation, req.GetId())
+	} else {
+		message = fmt.Sprintf("%s %s", req.GetType(), operation)
+	}
+
+	if req.GetCloudServiceId() != "" && len(params) > 0 {
+		message = fmt.Sprintf("%s for Cloud Service '%s' and Catalog '%s'", message, req.GetCloudServiceId(), params[0])
```
oxisto, line 68: Very specific to the catalog. Maybe it would make more sense in this case to just append `params` completely and move the `"and Catalog..."` string to the caller.

oxisto (changes requested):
Looks like a very clean approach in general. I do have some naming/constant using issues.

api/orchestrator/orchestrator.go
anatheka: That also bothered me all the time :)

logging/logging.go
anatheka: I used strings because I use them again in the log message `(fmt.Sprintf("%s %s with ID '%s'", req.GetType(), request, req.GetPayloadID())).` I wanted to avoid writing more ifs for the requestType. I am not sure how to use enums with corresponding strings.

anatheka: I used strings because I use them again in the log message `(fmt.Sprintf("%s %s with ID '%s'", req.GetType(), request, req.GetPayloadID())).` I wanted to avoid writing more ifs for the requestType. I am not sure how to use enums with the corresponding strings. But I will find a solution.

logging/logging.go
anatheka: I'll change that, but isn't LogRequest not misleading? The function creates a log message and not a request. It only gets a request.

api/orchestrator/orchestrator.go
```diff
@@ -63,10 +62,11 @@ type CloudServiceRequest interface {
 	proto.Message
 }
 
+type T any
```
oxisto, line 65: I actually wanted to make this a type parameter of `LogRequest` ;)

api/orchestrator/orchestrator.go
```diff
@@ -63,10 +62,11 @@ type CloudServiceRequest interface {
...
+
 type LogRequest interface {
```
oxisto, line 67: `PayloadRequest`?

service/orchestrator/toe.go
```diff
@@ -100,7 +102,8 @@ func (svc *Service) CreateTargetOfEvaluation(ctx context.Context, req *orchestra
 
 	res = req.TargetOfEvaluation
 
-	logging.LogMessage(log, logging.LoglevelInfo, logging.Create, req, req.TargetOfEvaluation.GetCatalogId())
+	params := []string{fmt.Sprintf("and Catalog '%s'", req.TargetOfEvaluation.GetCatalogId())}
```
oxisto, line 105: any specific reasons for not inlining this in the request? casting this into a []string and then using `...` seems unnecessary :)

anatheka: Yes, because it becomes hard to read, I find ;) But I'll change it.

api/orchestrator/orchestrator.go
anatheka: :+1:

api/streams.go
```diff
@@ -206,15 +205,25 @@ func (c *StreamChannelOf[StreamType, MsgType]) sendLoop(s *StreamsOf[StreamType,
...
-			} else {
-				s.log.Debugf("%T sent to %s (%s)", m, c.component, c.target)
-			}
+		// We want to log some additional information about this stream and its payload
+		preq, ok := any(m).(PayloadRequest)
+		if !ok {
+			return
+		}
+
+		// Retrieve the payload from the request. The request itself is usually
```
oxisto, line 214: Lines 214-226 are also done in `LogRequest`. Maybe we can even call LogRequest here?! with the `sent to xxx` part as params? @anatheka

anatheka: I did it as in your suggestion. However, I changed the logging message again from `*assessment.Metric updated with ID 'TransportEncryptionEnabled'.` to `*assessment.Metric with ID 'TransportEncryptionEnabled' updated.` so that it fits. 

Furhermore, I had to move api/request.go to api/request/request.go due to import cycle errors from api/streams.go and internal/logging/logging.go.

@oxisto please check if it fits.

api/orchestrator/orchestrator.go
```diff
@@ -62,6 +62,10 @@ type CloudServiceRequest interface {
 	proto.Message
 }
 
+type PayloadRequest interface {
```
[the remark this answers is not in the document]
anatheka: So you mean for all other *Requests in all services? Do we really need that?

api/streams.go
oxisto: I tried to solve the circular dependency in a different way, similar to how protobuf does it. The actual definition of the interface is now moved to an internal package (which gives us more freedom) and the public `api` package just contains a type alias.

oxisto (approved):

PR #979, opened 2023-02-22
Removed `ListAssessmentResults` from assessment

oxisto:
This finally removes the `ListAssessmentResults` from the assessment service, which caused confusion. This also moves the `ListAssessmentResultsRequest` and response to the `orchestrator` package.


api/orchestrator/orchestrator.proto
```diff
@@ -614,7 +613,7 @@ message AssessmentTool {
 
   // a list of metrics that this tool can assess, referred by their ids
   repeated string available_metrics = 4
-      [ (validate.rules).repeated.items.string.min_len = 1 ];
+      [ (validate.rules).repeated .items.string.min_len = 1 ];
```
anatheka, line 616:
```suggestion
      [ (validate.rules).repeated.items.string.min_len = 1 ];
```

anatheka: Why did you add the spaces behind .repeated?

service/assessment/assessment.go
```diff
@@ -250,7 +243,7 @@ func (svc *Service) AssessEvidence(_ context.Context, req *assessment.AssessEvid
 	}
 
 	// Assess evidence
-	err = svc.handleEvidence(req.Evidence, resourceId)
+	_, err = svc.handleEvidence(req.Evidence, resourceId)
```
anatheka, line 246: Why are you retourning the assessment results in handleEvidence() if you not need them?

anatheka (commented):
I'm not sure if it is a good idea to move the ListAssessmentResults to the orchestrator. This means we can no longer deploy the assessment component without the orchestrator. I am also not sure if we want to do that. 

Furthermore, if we leave it like this, shouldn't we also move the ListEvidences() and GetEvidence() to the Orchestrator?

service/assessment/assessment.go
oxisto: We had one test that tested whether the results were created successfully. It kind of makes sense to return it, if somebody wants to have a look at it.

api/orchestrator/orchestrator.proto
oxisto: This is my stupid protobuf formatter, sorry.

service/assessment/assessment.go
anatheka: You mean while debugging? That we can also do in the handle function, but ok.

anatheka (approved):

PR #980, opened 2023-02-22
Persist discovered resource in discovery service

oxisto:


service/discovery/discovery.go
```diff
@@ -186,6 +191,14 @@ func NewService(opts ...ServiceOption) *Service {
 		o(s)
 	}
 
+	// Default to an in-memory storage, if nothing was explicitly set
+	if s.storage == nil {
+		s.storage, err = inmemory.NewStorage()
+		if err != nil {
+			log.Errorf("Could not initialize the storage: %v", err)
```
anatheka, line 198:
```suggestion
			log.Errorf("Could not initialize in-memory storage: %v", err)
```

anatheka: Maybe another error message if no persistence storage is explicitly set?

anatheka (commented):
Looks good, only minor comments.

service/discovery/discovery.go
oxisto: I copy/pasted this snippet from the other services just to be consistent. I would suggest not to change that in this PR.

api/discovery/discovery.proto
```diff
@@ -72,6 +72,28 @@ message QueryRequest {
...
+// Resource is a wrapper around google.protobuf.Value that is needed for
+// persistence reasons.
+message Resource {
+  // Id contains a unique ID for each resource. This is specific for the cloud
+  // provider this resource was gathered for and can for example be a resource
+  // URL.
+  string id = 1 [ (validate.rules).string.uuid = true ];
+  // CloudServiceId is the UUID for the cloud service to which this resource
+  // belongs to.
+  string cloud_service_id = 2 [ (validate.rules).string.min_len = 1 ];
```
anatheka, line 88:
```suggestion
  string cloud_service_id = 2 [ (validate.rules).string.uuid = true ];
```

api/discovery/discovery.proto
```diff
@@ -72,6 +72,28 @@ message QueryRequest {
...
+  // ResourceType contains a comma seperated string of resource types according
+  // to our ontology.
+  string resource_type = 3 [ (validate.rules).string.min_len = 1 ];
```
anatheka, line 91: Is there a reason not to use a list here as well as in the assessment result and call it resource_types? 

```
// Resource types
  repeated string resource_types = 8 [
    (tagger.tags) = "gorm:\"serializer:json\"",
    (validate.rules).repeated .min_items = 1
  ];
```

anatheka (changes requested):

api/discovery/discovery.proto
oxisto: Sort of, because we do a filter on the resource type. I think this may also be possible with the JSON type but I am not sure if this is possible in the same way with Postgres and sqlite. I will do a quick check, if this works.

oxisto: I did a check and it is a complete mess to check this in a JSON. Therefore, I would keep it as string

anatheka (approved):

PR #982, opened 2023-02-24
Add missing information in metrics.json

lebogg:


lebogg:
Something wrent wrong here

PR #983, opened 2023-02-24
Add missing information in metrics list (metrics.json)

lebogg:


lebogg:
Is there a difference between TLSVersion and TransportEncryptionAlgorithm metrics?

anatheka:
> Is there a difference between TLSVersion and TransportEncryptionAlgorithm metrics?

I think the TransportEncryptionAlgorithm is for example TLS and the TLSVersion a specific version for TLS (e.g., 1.3). We could have also another TransportEncryptionAlgorithm with other versions.

But you are right, currently we are not differentiating between them. We only check "is it TLS?" and then we check if it is a specific TLS version. 

The same question I had a few days ago :smile:

service/orchestrator/metrics.json
```diff
@@ -1,128 +1,450 @@
...
-    "id": "MutualAuthentication"
+    "id": "MutualAuthentication",
+    "name": "Mutual Authentication",
+    "description": "This metric is used to assess if mutual authentication, including client certificate, has been enabled",
+    "category": "",
+    "scale": 1,
+    "range": {
+      "allowedValues": {
+        "values": [
+          "*"
```
anatheka, line 302: Shouldn't that be a bool?

service/orchestrator/metrics.json
```diff
@@ -1,128 +1,450 @@
...
+        ]
+      }
+    }
   },
   {
     "id": "TransportEncryptionAlgorithm",
     "name": "Transport Encryption: Protocol Version",
     "category": "CKM-02",
-    "description": "This metric describes, whether a up-to-date transport encryption protocol version is used",
+    "description": "This metric describes, if a up-to-date transport encryption protocol version is used",
```
anatheka, line 442:
```suggestion
    "description": "This metric describes, if a up-to-date transport encryption protocol is used",
```

service/orchestrator/metrics.json
```diff
@@ -1,128 +1,450 @@
...
+    "description": "This metric is used to assess the configured log retention (days)",
+    "category": "",
+    "scale": 3,
+    "range": {
+      "allowedValues": {
+        "values": [
+          true,
+          false
+        ]
+      }
```
anatheka, line 189:
```suggestion
      "minMax": {
        "min": 1,
        "max": 365
      }
```

service/orchestrator/metrics.json
lebogg: Yes, then it is more consistent. Bit it is almost a good example for a metric where our current notion of `scale` and `range` reaches its limits. Because it an more "complex" metric where we check multiple things, it is not so easy applicable. But in these case both checks are of boolean nature, so it is fine to change it.

service/orchestrator/metrics.json
```diff
@@ -1,128 +1,450 @@
...
+    }
   },
   {
-    "id": "AnomalyDetectionOutput"
+    "id": "AnomalyDetectionOutput",
+    "name": "Anomaly Detection: Output",
+    "description": "This metric is used to assess if the output is ???",
```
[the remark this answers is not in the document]
anatheka: But couldn't we describe it as "This metric is used to assess if the output location is valid." or "This metric is used to assess if the output is a valid location." or something similar?

service/orchestrator/README.md
```diff
@@ -0,0 +1,9 @@
+# `metrics.json`: Format of metrics (see metric.proto for more details)
+| Field | Description |
+| ----------- | ----------------- |
+| id   | The ID of metric. Must not contain spaces.  |
+| name     | A more readable name    |
+| description  | A short description |
+| category  | The control of the catalog |
```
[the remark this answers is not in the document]
oxisto: I think that this property is obsolete and can be removed

service/orchestrator/README.md
```diff
@@ -0,0 +1,9 @@
+# `metrics.json`: Format of metrics (see metric.proto for more details)
+| Field | Description |
```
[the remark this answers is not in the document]
oxisto: The idea is nice but wouldn't it make more sense to see how we can auto-generate this. I foresee a future where we change things in the metric and nobody is updating this file. I would remove it actually.

service/orchestrator/metrics.json
```diff
@@ -1,128 +1,450 @@
...
+        ]
+      }
     }
   },
   {
     "id": "AtRestEncryptionEnabled",
     "name": "Storage Encryption: Enabled",
     "category": "CKM-03",
-    "description": "This metric describes whether storage encryption is enabled",
+    "description": "This metric describes if storage encryption is enabled",
```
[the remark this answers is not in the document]
oxisto: if used with "describes", "whether" is the correct term here, otherwise you are describing something only if storage encryption is enabled and that does not make sense ;)

lebogg:
Why again, did we blocked this? @oxisto @anatheka

oxisto:
> Why again, did we block this? @oxisto @anatheka

We didn't block it, you stopped working on it :P

It is not working however in the current state because of the way that the information is read into the engine. Last time I started a real Clouditor engine with this PR everything was broken.

lebogg:
Ok, I will look into it :+1: 
I thought you were saying we first have to change sth. else (maybe in Rego-specific code) until we can move forward here.

service/orchestrator/metrics.json
```diff
@@ -1,128 +1,421 @@
...
+    "description": "This metric is used to assess if a up-to-date transport encryption protocol is used",
     "scale": 1,
     "range": {
       "order": {
         "values": [
-          "TLS 1.0",
-          "TLS 1.1",
-          "TLS 1.2",
-          "TLS 1.3"
+          "TLS"
```
oxisto, line 418: shouldn't this include the version numbers?

lebogg:
> Nope. Running this with a real Postgres DB still results in a lot of errors:
> 
> ```
> INFO[0000] Creating storage                              component=storage
> ERRO[0000] Error while saving metrics: ERROR: ON CONFLICT DO UPDATE requires inference specification or constraint name (SQLSTATE 42601)  component=orchestrator
> ...
> ```

- First of all, this error also occurs in main ;)
- I think it has something to do with the fact that we use in `loadMetrics` the `Save` DB-operation. Saving (=creating in this case) the metric succeeds but saving the implementation not. My guess is that the `where` clause doesn't find the corresponding metric implementation since it doesn't exist. Or maybe it has sth. to do with the association to the metric and saving/updating the implementation fails since there is no association yet. I changed both operations to `Create` and it works now.
- Another way maybe is to allow updates on conflicts for Postgres (when we assume the conflict is valid here): https://gorm.io/docs/create.html#Upsert-x2F-On-Conflict. This would also makes sense why it is working in the in memory case.
- I still have to test if the implementation is cascaded correctly to the metric

service/orchestrator/metrics.json
lebogg: TBH, I always stumble over our approach with the two metrics `TransportEncryptionAlgorithm` and `TLSVersion`. The former checks if we use an appropriate protocol, e.g. TLS but could also be another in theory. And the latter only argues about the TLS version. Or do you see each TLS version as its own EncryptionAlgorithm? If so, the `TLSVersion` metric is redundant .
I think either we do sth. like `TransportEncryptionAlgorithm` and `TransportEncryptionAlgorithmVersion`, or just `TLSXXX` metrics^^

anatheka: I feel the same way and would prefer to use only one metric like TLSVersion. :smile:

oxisto: Since there are no other suitable transport encryption algorithms besides TLS I think it’s not really practical to have a metric with just „TLS“. I would propose to merge them but with a TransportEncryption „prefix“

service/orchestrator/metrics.go
```diff
@@ -72,9 +72,9 @@ func (svc *Service) loadMetrics() (err error) {
 	// still need to be aware of the particular metric.
 	for _, m := range metrics {
 		// Somehow, we first need to save the metric, otherwise we are running into weird constraint issues.
```
anatheka, line 74:
```suggestion
		// Somehow, we first need to create the metric, otherwise we are running into weird constraint issues.
```

service/orchestrator/metrics.go
```diff
@@ -72,9 +72,9 @@ func (svc *Service) loadMetrics() (err error) {
...
-		err = svc.storage.Save(m, "id = ?", m.Id)
+		err = svc.storage.Create(m)
 		if err != nil {
-			log.Errorf("Error while saving metrics: %v", err)
+			log.Errorf("Error while saving metric configuration for '%s': %v", m.Id, err)
```
anatheka, line 77: I would let it as "Error while saving metrics" as we are storing the metric and not only the metric_configuration.

anatheka (approved):
Just a few minor things from the last review. And I updated the demo_catalog file for the updated metric names.

PR #988, opened 2023-02-27
Add parent controls to controls_in_scope at ToE creation

anatheka:
This PR 
- adds the controls (e.g., OPS-01) to the controls_in_scope, 
- updates the testdata mock control variables,
- adds a MockControlsInScope,
- cuts off the seconds from timestamp for less test errors  (Test_GetCertificate()) and
- fixes the ListControls(). The  case that only the catalog id is given was missing.


anatheka:
I'm not sure if that is the best solution for getting the parent controls. We can also use a hash map, but I think since there are not so many controls we can do it that way. 

Also, we could consider querying the controls for the corresponding assurance levels directly via the database. But for now, I would leave it as is.

service/orchestrator/toe.go
```diff
@@ -408,7 +409,19 @@ func getControls(controls []*orchestrator.Control, levels []string, level string
 		c = append(c, high...)
 	}
 
-	return c, nil
+	// Add parent controls to the sub-control included in the list. That results in duplicates that we can remove later.
```
oxisto, line 412: Stupid question: Do the "parent" controls have the "assurance level" set? Because if its empty, you could just add all the controls that do not have an assurance level and you would  only need to do it once and not loop the controls again.

oxisto: This would basically add the semantic that a control with an empty assurance level will be added regardless of the assurance level chosen in the ToE

anatheka: Maybe I don't understand it correctly, but then I have also to loop again and find the controls which do not have sub-controls in the list.

PR #997, opened 2023-03-07
Retrieve allow all authorisation decision from JWT

oxisto:


anatheka (changes requested):
LGTM, but I think a test for  `"func (a *AuthorizationStrategyJWT) AllowedCloudServices(ctx context.Context)` is missing.

anatheka (approved):
:+1:

PR #999, opened 2023-03-09
Add filter options to `ListEvidences`

### LINKED ISSUE #967: Add filtering for ListEvidences
The following filter options should be added
- CloudServiceID
- ToolID
- ~~Resource~~ -> is done by #1021 in the future when we added resource message to proto (#1016)

lebogg:
Some Guidance:
- [x] Making yourself familiar with CLI/UI/grpcurl (helps with testing)
- [x] Update .proto (api folder)
- [x] Generate Go files (with buf)
- [x] Modifiy `ListEvidences` in `evidence_store`
- [x]  Write Test-Case for ListEvidences()

`ListAssessmentResults` should be quite similar: https://github.com/clouditor/clouditor/blob/1ffa3bba642e6b5ae8221473cfd95f2562f6e7a5/service/orchestrator/assessment_results.go#L47

Resolves #967


luis-gasparschroeder:
- [ ] Write Test-Case for ListEvidences()

service/evidence/evidence_store.go
```diff
@@ -186,6 +187,10 @@ func (svc *Service) ListEvidences(ctx context.Context, req *evidence.ListEvidenc
 		allowed []string
 		conds   []any
 	)
+	// Validate Input
```
oxisto, line 190: This is already part of `service.ValidateRequest`

service/evidence/evidence_store.go
```diff
@@ -202,9 +207,34 @@ func (svc *Service) ListEvidences(ctx context.Context, req *evidence.ListEvidenc
...
+		}
+		if filter.GetToolId() != "" {
+			query = append(query, "tool_id = ?")
+			conds = append(conds, filter.GetToolId())
+		}
+	}
+
+	// In any case, we need to make sure that we only select assessment results of cloud services that we have access to
+	if !all {
+		// query = append(query, "cloud_service_id IN ?")
```
oxisto, line 227: Why is this line commented out?

internal/testutil/evidencetest/evidence.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"clouditor.io/clouditor/api/evidence"
+	"clouditor.io/clouditor/internal/testdata"
+	"clouditor.io/clouditor/internal/util"
+	"google.golang.org/protobuf/types/known/structpb"
+	"google.golang.org/protobuf/types/known/timestamppb"
+)
+
+var (
+	MockListEvidenceRequest1 = &evidence.ListEvidencesRequest{
```
oxisto, line 12: I would prefer keeping these in the table tests directly, otherwise its very hard to see what kind of values the tests are using.

service/evidence/evidence_store_test.go
```diff
@@ -293,34 +293,176 @@ func TestService_StoreEvidences(t *testing.T) {
...
+			},
+			wantErr: assert.NoError,
+			wantResp: func(tt assert.TestingT, i1 interface{}, i2 ...interface{}) bool {
+				res, ok := i1.(*evidence.ListEvidencesResponse)
+				assert.True(t, ok)
 
-	resp, err := s.ListEvidences(context.TODO(), &evidence.ListEvidencesRequest{})
-	assert.NoError(t, err)
-	assert.Equal(t, 2, len(resp.Evidences))
+				for _, r := range res.Evidences {
```
lebogg, line 338:
```suggestion
				// Loop through all received evidences and check whether tool and service ids are correct. 
				for _, r := range res.Evidences {
```

service/evidence/evidence_store_test.go
lebogg, line 1: I would add these tests:
- pagination error (reproduce the error you had in the beginning)
- tool id is not a UUID
- service id is not a UUID
For the last 2 tests you first have to define that they should be UUIDs in the proto files. Then the corresponding validate function will return error if it is not a valid UUID. See, e.g.:
https://github.com/clouditor/clouditor/blob/995e9390d73a635014fe8861c6f570a7dec640c7/api/orchestrator/orchestrator.proto#L601

api/evidence/evidence_store.proto
```diff
@@ -80,11 +80,19 @@ message StoreEvidencesResponse {
 }
 
 message ListEvidencesRequest {
+  // page_size: 0 = default, > 0 = set value
   int32 page_size = 1;
   string page_token = 2;
   string order_by = 3;
   bool asc = 4;
+  Filter filter = 5;
```
lebogg, line 88: I am not sure if we should make this optional @oxisto @anatheka ?

lebogg (commented):
The last filtering option mentioned in the issue is missing: "resource". @anatheka I guess you mean the resource id by that?

api/evidence/evidence_store.proto
anatheka: I'm not sure if we should set the Filter or the properties of the filter optional. Does one or the other have any implications?

internal/testutil/evidencetest/evidence.go
luis-gasparschroeder: The idea behind a separate file was the decoupling of test data. Might be the case that other tests need the data, this would lead to code duplication in my opinion.

service/evidence/evidence_store.go
luis-gasparschroeder: Forgetfulness

luis-gasparschroeder:
Your feedback is applied! 
Feel free checking out the changes taken👍🏻

service/evidence/evidence_store.go
```diff
@@ -186,12 +187,15 @@ func (svc *Service) ListEvidences(ctx context.Context, req *evidence.ListEvidenc
 		allowed []string
 		conds   []any
 	)
-
 	// Validate request
 	err = service.ValidateRequest(req)
 	if err != nil {
 		return nil, err
 	}
+	// Validate context
```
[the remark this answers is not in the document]
oxisto: Ok fair enough. The "problem" that I see is that we want to minimise the amount of "custom" things we do in a request. So we should probably also check the context as part of the `ValidateRequest` request method. But we should do this then for all requests at once and not as part of this PR for a single request.

On the other hand, a `nil` request is not really a problem per se. Only if the current `AuthorizationStrategy` that takes information out of the request does not do a proper check (which I now see, that the `AuthorizationStrategyJWT` does NOT, so this is a problem, but probably also for separate PR)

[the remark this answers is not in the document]
oxisto: See https://github.com/clouditor/clouditor/pull/1024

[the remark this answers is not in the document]
luis-gasparschroeder: If a nil request is not a problem, should I just remove the validate request if statement?

service/evidence/evidence_store.go
```diff
@@ -202,9 +206,34 @@ func (svc *Service) ListEvidences(ctx context.Context, req *evidence.ListEvidenc
...
+			query = append(query, "cloud_service_id = ?")
+			conds = append(conds, filter.GetCloudServiceId())
+		}
+		if filter.GetToolId() != "" {
+			query = append(query, "tool_id = ?")
+			conds = append(conds, filter.GetToolId())
+		}
+	}
+
+	// In any case, we need to make sure that we only select assessment results of cloud services that we have access to
```
lebogg, line 224:
```suggestion
	// In any case, we need to make sure that we only select evidences of cloud services that we have access to
```

service/evidence/evidence_store.go
```diff
@@ -202,9 +207,34 @@ func (svc *Service) ListEvidences(ctx context.Context, req *evidence.ListEvidenc
...
```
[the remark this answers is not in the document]
lebogg:
```suggestion
	// In any case, we need to make sure that we only select evidences of cloud services that we have access to
```

api/evidence/evidence_store.proto
lebogg: I would think to apply optional to Filter as well as for the fields in it. Because we can choose if we want to apply any filter or not (making `Filter` optional) and then we choose which exact filter we want to apply (making `Filter`s fieleds opional. I guess, at least, these fields have to be optional anyway).

@luis-gasparschroeder Please do not resolve conversations where we not discussed the matter conclusively :)

service/evidence/evidence_store.go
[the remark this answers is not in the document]
luis-gasparschroeder: Ok sure!

api/evidence/evidence_store.proto
luis-gasparschroeder: I made the Filter optional

internal/testdata/testdata.go
```diff
@@ -56,10 +56,18 @@ const (
...
+	MockAnotherEvidenceID        = "22222222-2222-2222-2222-222222222222"
+	MockAnotherEvidenceToolID    = "49d85e98-c3da-11ed-afa1-0242ac120002"
+	MockEvidencePageSize         = 0
+	MockAnotherEvidencePageSize  = 2
+	MockEvidencePageToken        = ""
+	MockAnotherEvidencePageToken = ""
+	MockEvidenceOrderBy          = ""
+	MockAnotherEvidenceOrderBy   = ""
+	MockEvidenceAsc              = true
+	MockAnotherEvidenceAsc       = false
```
lebogg, line 70: Remove unused constants

service/evidence/evidence_store.go
```diff
@@ -202,9 +202,34 @@ func (svc *Service) ListEvidences(ctx context.Context, req *evidence.ListEvidenc
 
 	res = new(evidence.ListEvidencesResponse)
 
+	var query []string
+
+	// Apply filter options
+	if req.GetFilter() != nil {
+		filter := req.GetFilter()
```
lebogg, line 209: There is a nice litte Go idiom to make this a one-liner: `if filter := req.GetFilter(); filter != nil {`

service/evidence/evidence_store.go
```diff
@@ -202,9 +202,34 @@ func (svc *Service) ListEvidences(ctx context.Context, req *evidence.ListEvidenc
...
+		if filter.GetCloudServiceId() != "" {
+			query = append(query, "cloud_service_id = ?")
+			conds = append(conds, filter.GetCloudServiceId())
+		}
```
lebogg, line 213: here you can apply it as well (to call `GetCloudServiceId` only once)

service/evidence/evidence_store.go
```diff
@@ -202,9 +202,34 @@
...
+		if filter.GetToolId() != "" {
```
lebogg, line 214: same here

service/evidence/evidence_store_test.go
```diff
@@ -293,34 +295,205 @@ func TestService_StoreEvidences(t *testing.T) {
...
+					PageToken: evidencetest.MockListEvidenceRequest2.PageToken,
+					OrderBy:   "Wrong Input",
+					Asc:       evidencetest.MockListEvidenceRequest2.Asc,
+				},
+			},
+			wantErr:  assert.Error,
+			wantResp: assert.Nil,
+		},
+	}
+	for _, currentTest := range tests {
```
lebogg, line 485: Use `tt` please to be consistent with other code in this project. If we decide to change it later, we will do a full refactoring then.

service/evidence/evidence_store.go
lebogg, line 1: I think there is a check missing for the case that the user has not the permission to fetch evidences for the requested cloud service id, sth. like this:
```
if !all && req.GetFilter().GetCloudServiceId() != "" && !slices.Contains(allowed, req.GetFilter().GetCloudServiceId()) {
	}
```

service/evidence/evidence_store.go
luis-gasparschroeder: Very handy!

service/evidence/evidence_store_test.go
```diff
@@ -293,34 +297,234 @@ func TestService_StoreEvidences(t *testing.T) {
...
+				req: &evidence.ListEvidencesRequest{
+					PageSize:  evidencetest.MockListEvidenceRequest1.PageSize,
+					PageToken: evidencetest.MockListEvidenceRequest1.PageToken,
+					OrderBy:   evidencetest.MockListEvidenceRequest1.OrderBy,
+					Asc:       evidencetest.MockListEvidenceRequest1.Asc,
+					Filter: &evidence.Filter{
+						CloudServiceId: evidencetest.MockListEvidenceRequest1.Filter.CloudServiceId,
+						ToolId:         evidencetest.MockListEvidenceRequest1.Filter.ToolId,
+					},
+				},
```
lebogg, line 335: Ah, I just recognize it: You can simply use `evidencetest.MockListEvidenceRequest1` as request (`req: evidencetest.MockListEvidenceRequest1`). Or do I miss something?

service/evidence/evidence_store_test.go
```diff
@@ -706,3 +901,22 @@ func TestService_GetEvidence(t *testing.T) {
 		})
 	}
 }
+
+func newAuthorizationStrategy(cloudServiceIDs []string) service.AuthorizationStrategy {
```
oxisto, line 905: Just a minor thing. This could actually be interesting for multiple tests so it might make sense to move this into into the `testutil` package. Otherwise I think we can finally merge this for now.

oxisto (approved):

service/evidence/evidence_store_test.go
luis-gasparschroeder: @lebogg

lebogg: Yes, I did it initially but circular dependencies ...

lebogg: Maybe you can outsource the `AuthorizationStrategy` interface to API or internal?

PR #1008, opened 2023-03-13
Add Resource Group Argument to Engine

### LINKED ISSUE #1007: Add Option to Start Engine with specific Resource Group Discovery
Currently, the engine starts the Discovery (when auto start is selected) without a configuration for scoping it to a resource group.
This should be possible via a cmd argument (=flag) or via a environment variable.

lebogg:
With this PR, the engine is able to start the discovery with a pre-configured resource group. And due to Viper configuration also via environment variable possible.

Resolves #1007


cmd/engine/engine.go
```diff
@@ -163,6 +166,7 @@ func init() {
 	engineCmd.Flags().Bool(CreateDefaultTarget, DefaultCreateDefaultTarget, "Creates a default target cloud service if it does not exist")
 	engineCmd.Flags().Bool(DiscoveryAutoStartFlag, DefaultDiscoveryAutoStart, "Automatically start the discovery when engine starts")
 	engineCmd.Flags().StringSliceP(DiscoveryProviderFlag, "p", []string{}, "Providers to discover, separated by comma")
+	engineCmd.Flags().String(DiscoveryResourceGroupFlag, DefaultDiscoveryResourceGroup, "Set the scope of the discovery to a resource group")
```
oxisto, line 169:
```suggestion
	engineCmd.Flags().String(DiscoveryResourceGroupFlag, DefaultDiscoveryResourceGroup, "Limit the scope of the discovery to a resource group (currently only used in the Azure discoverer")
```

oxisto (approved):

lebogg:
@oxisto Are you happy with the test design of the discovery? I had to add a nil check for `authConfig.JWKS` due to a panic in tests (tricky to find since it was hidden in a defer ^^)

cmd/engine/engine.go
```diff
@@ -244,17 +249,19 @@ func doCmd(_ *cobra.Command, _ []string) (err error) {
...
-		service_discovery.WithStorage(db),
-		service_discovery.WithOAuth2Authorizer(
-			// Configure the OAuth 2.0 client credentials for this service
-			&clientcredentials.Config{
-				ClientID:     viper.GetString(ServiceOAuth2ClientIDFlag),
-				ClientSecret: viper.GetString(ServiceOAuth2ClientSecretFlag),
-				TokenURL:     viper.GetString(ServiceOAuth2EndpointFlag),
-			}),
-	)
+	if discoveryService == nil {
```
oxisto, line 252: What is the point of this nil check?

cmd/engine/engine.go
```diff
@@ -394,10 +405,13 @@ func doCmd(_ *cobra.Command, _ []string) (err error) {
 	log.Infof("Starting gRPC endpoint on :%d", grpcPort)
 
 	// Automatically start the discovery, if we have this flag enabled
-	if viper.GetBool(DiscoveryAutoStartFlag) {
+	getBool := viper.GetBool(DiscoveryAutoStartFlag)
```
oxisto, line 408: what's the point of this bool?

cmd/engine/engine_test.go
```diff
@@ -88,3 +91,50 @@ func Test_doCmd(t *testing.T) {
 		})
 	}
 }
+
+func Test_doCMDWithDiscovery(t *testing.T) {
```
oxisto, line 95: Personally, I am not sure if there is much sense in testing this in "main". In the long run it would probably be better to extract functionality out of our main method into smaller functions and make them testable, but going through hoops to have some testing function just for the sake of having tests in main is counter-productive in my opinion

oxisto (commented):
Personally, I could live without the tests.

cmd/engine/engine.go
lebogg: That the mockDiscoveryService doesn't get over written.

cmd/engine/engine.go
lebogg: Oh, it was for debugging. I change it back

cmd/engine/engine_test.go
lebogg: Can I somehow hard-reset to 407c26af4be03503647d8020e0336546ba0f77de (Update cmd/engine/engine.go)?

PR #1026, opened 2023-03-23
Add get request for assessment results

luis-gasparschroeder:
The method GetAssessmentResult is implemented in clouditor/service/orchestrator. 
Accordingly, the orchestrator.proto file was adjusted in clouditor/api/orchestrator. 
Lastly, test cases were written.


go.mod
```diff
@@ -168,6 +168,7 @@ require (
 	golang.org/x/lint v0.0.0-20210508222113-6edffad5e616 // indirect
 	golang.org/x/mod v0.8.0 // indirect
 	golang.org/x/tools v0.6.0 // indirect
+	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.3.0 // indirect
```
lebogg, line 171: `go mod tidy` will remove this unused dependency

internal/testutil/orchestratortest/orchestrator.go
```diff
@@ -132,6 +132,15 @@ func NewAnotherCloudService() *orchestrator.CloudService {
 	}
 }
 
+var (
+	MockAssessmentResultRequest1 = &orchestrator.AssessmentResultRequest{
+		Id: testdata.MockAssessmentResult1ID,
+	}
+	MockAssessmentResultRequest2 = &orchestrator.AssessmentResultRequest{
```
lebogg, line 139: Unused

service/orchestrator/assessment_results.go
```diff
@@ -43,6 +44,27 @@ import (
 	"clouditor.io/clouditor/service"
 )
 
+// GetAssessmentResult gets one assessment result by id
+func (svc *Service) GetAssessmentResult(ctx context.Context, req *orchestrator.AssessmentResultRequest) (res *assessment.AssessmentResult, err error) {
```
lebogg, line 48: Replace `ctx` with underscore (blank identifier) since we don't use it here

service/orchestrator/assessment_results.go
```diff
@@ -43,6 +44,27 @@ import (
...
+
+	// Validate request
+	err = service.ValidateRequest(req)
+	if err != nil {
```
lebogg, line 52: You can use the one liner idiom here

service/orchestrator/assessment_results.go
```diff
@@ -43,6 +44,27 @@ import (
...
+		return nil, err
```
lebogg, line 53: You could just use `return` only

lebogg, line 53:
```suggestion
		return
```

service/orchestrator/assessment_results_test.go
```diff
@@ -51,6 +51,77 @@ import (
...
+		wantErr assert.ErrorAssertionFunc
+	}{
+		{
+			name: "request is missing",
+			fields: fields{
+				storage: testutil.NewInMemoryStorage(t),
+			},
+			args:    args{},
+			res:     nil,
+			wantErr: assert.Error,
```
lebogg, line 77: When you use `assert.Error`, you just know that SOME error occurred. But for the case of existing errors, you want to have checks that is really the error you are targeting (here it can be 3 different kinds of error).
Thus, it is better to create a function here and check the respective errors.
See example of last error case (internal) in the other comment

service/orchestrator/assessment_results_test.go
```diff
@@ -51,6 +51,77 @@ import (
...
+			fields: fields{
+				storage: testutil.NewInMemoryStorage(t, func(s persistence.Storage) {
+					assert.NoError(t, s.Create(&orchestratortest.MockAssessmentResult2))
+					assert.NoError(t, s.Create(&orchestratortest.MockAssessmentResult3))
+				}),
+			},
+			args:    args{req: orchestratortest.MockAssessmentResultRequest1},
+			req:     orchestratortest.MockAssessmentResultRequest1,
+			res:     nil,
+			wantErr: assert.Error,
```
lebogg, line 103: see above

service/orchestrator/assessment_results_test.go
lebogg, line 1: You can also check the the database error. In `testutil` there is a storage mock for this case. Then do something like this:
```
storage: &testutil.StorageWithError{GetErr: ErrSomeError}
```
...
```
wantErr: func(t assert.TestingT, err error, i ...interface{}) bool {
				assert.Equal(t, codes.Internal, status.Code(err))
				return assert.Contains(t, err.Error(), ErrSomeError.Error())
			}
```

lebogg (commented):
looks good but I have some recommendations :)

api/orchestrator/orchestrator.proto
```diff
@@ -778,6 +786,10 @@ message TargetOfEvaluation {
   ];
 }
 
+message AssessmentResultRequest {
+  string id = 1;
```
oxisto, line 790: missing validate here. ID should always be a uuid.

api/orchestrator/orchestrator.proto
```diff
@@ -778,6 +786,10 @@ message TargetOfEvaluation {
   ];
 }
 
+message AssessmentResultRequest {
```
oxisto, line 789: This should be named `GetAssessmentResultRequest`

api/orchestrator/orchestrator.proto
```diff
@@ -97,6 +97,14 @@ service Orchestrator {
   rpc StoreAssessmentResults(stream StoreAssessmentResultRequest)
       returns (stream StoreAssessmentResultsResponse);
 
+  // Get an assessment result by ID
```
oxisto, line 100:
```suggestion
  // Retrieves a particular assessment result by its ID
```

internal/testutil/orchestratortest/orchestrator.go
```diff
@@ -132,6 +132,12 @@ func NewAnotherCloudService() *orchestrator.CloudService {
 	}
 }
 
+var (
```
oxisto, line 135: can be joined with the `var` block below

service/orchestrator/assessment_results.go
```diff
@@ -26,6 +26,7 @@
 package orchestrator
 
 import (
+	"clouditor.io/clouditor/persistence"
```
oxisto, line 29: import order

service/orchestrator/assessment_results.go
```diff
@@ -43,6 +44,26 @@ import (
...
 )
 
+// GetAssessmentResult gets one assessment result by id
+func (svc *Service) GetAssessmentResult(_ context.Context, req *orchestrator.AssessmentResultRequest) (res *assessment.AssessmentResult, err error) {
+
+	// Validate request
+	if service.ValidateRequest(req) != nil {
+		return
+	}
+
```
oxisto, line 54: Missing authorization

service/orchestrator/assessment_results_test.go
```diff
@@ -29,6 +29,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"google.golang.org/grpc/codes"
```
oxisto, line 32: import order formatting

internal/testutil/orchestratortest/orchestrator.go
luis-gasparschroeder: MockAssessmentResultRequest1 is of type orchestrator.GetAssessmentResultRequest, whereby MockAssessmentResult1 is of type assessment.AssessmentResult.

Should variables of different types be joined in Go?

oxisto: That’s not a problem

service/orchestrator/assessment_results.go
```diff
@@ -40,9 +40,35 @@ import (
...
+// GetAssessmentResult gets one assessment result by id
+func (svc *Service) GetAssessmentResult(ctx context.Context, req *orchestrator.GetAssessmentResultRequest) (res *assessment.AssessmentResult, err error) {
+
+	// Validate request
+	if service.ValidateRequest(req) != nil {
+		return
+	}
+
+	// Check, if this request has access to the cloud service according to our authorization strategy.
+	if !svc.authz.CheckAccess(ctx, service.AccessRead, req) {
```
oxisto, line 56: This will probably not work, because the request does not contain the cloud service ID, only the ID for a specific assessment result. 

We probably need to add an additional query, to check if `cloud_service_id` of the result is within `allowed`. There should be similar code already for list-assessment results.

oxisto: Alternatively, we restrict the quest is only retrieving a result based on id + cloud service id. But then we still need to update the query in line 62 and the HTTP API annotations in the protocol file.

api/orchestrator/orchestrator.proto
```diff
@@ -778,6 +786,13 @@ message TargetOfEvaluation {
   ];
 }
 
+message GetAssessmentResultRequest {
+  string id = 1
+      [ (validate.rules).string.uuid = true ];
+  string cloud_service_id = 2
```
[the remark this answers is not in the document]
luis-gasparschroeder: svc.authz.CheckAccess requires a cloud_service_id.

In case of no cloud_serivce_id, the following error occurs:
`Cannot use 'req' (type *orchestrator.GetAssessmentResultRequest) as the type api.CloudServiceRequest Type does not implement 'api.CloudServiceRequest' as some methods are missing: GetCloudServiceId() string`

What's the problem of having the cloud_service_id?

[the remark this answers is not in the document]
luis-gasparschroeder: The CheckAccess method requires a cloud_service_id. 
As far as I understood this comment (https://github.com/clouditor/clouditor/pull/1026#discussion_r1147925621) correctly, authorization should be added to this get-request. 

Other get-requests whose request parameter type has no cloud_service_id, do not have authorization checks
(i.e. GetMetricImplementation())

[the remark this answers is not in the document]
lebogg: In the case of `GetMetricImplementation`, the metric implementation is independent from a cloud service. But if there are other requests whose result is dependent on a cloud service, feel free to create an issue :)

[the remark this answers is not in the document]
oxisto: > The CheckAccess method requires a cloud_service_id. As far as I understood this comment ([#1026 (comment)](https://github.com/clouditor/clouditor/pull/1026#discussion_r1147925621)) correctly, authorization should be added to this get-request.
> 
> Other get-requests whose request parameter type has no cloud_service_id, do not have authorization checks (i.e. GetMetricImplementation())

Yes, but the problem is, 
a) nobody is filling the `cloud_service_id`, because the http pattern for request only contains the id (`get : "/v1/orchestrator/assessment_results/{id}"`). So the cloud service ID is always empty and the check would probably always fail.
b) you are only including the ID in the query to the database. So even if first check would succeed, you could retrieve any result by their ID, even if it's associated to another cloud service.

I discussed possible solutions in the other comment below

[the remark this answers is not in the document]
luis-gasparschroeder: Roger that

service/orchestrator/assessment_results.go
luis-gasparschroeder: GetAssessmentResultRequest only has one field -> id. 
GetAssessmentResult() checks if cloud_service_id is allowed.
Test cases are updated and expanded appropriately.

lebogg (approved):
LGTM

PR #1027, opened 2023-03-23
Add Common Issue section to ReadMe

luis-gasparschroeder:
With this PR the ReadMe includes a section with common issues I ran into

PR #1040, opened 2023-03-29
Exclude engine/main from codecov

lebogg:
To avoid failing codecov checks when modifying engine.go

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,6 +1,7 @@
 ignore:
   - "**/*.pb.go" # ignore protoc generated files
   - "**/*.pb.*.go" # ignore protoc-gen-* generated files
+  - "cmd/engine/engine.go" # ignore main
 coverage:
   range: "70...85"
   status:


anatheka:
Do we really want that? What errors do we have with the engine?

lebogg:
> What errors do we have with the engine?

No error, it is just cumbersome to test the main because to do so you have to change the code with e.g. introducing interfaces. But when you not test it, the codecov fails. Therefore, I would exclude it. If you want to test something anyway, nobody stops you from doing it. It is just about codecov check. To circumvent it, we probably should outsource more code out of main.

See also discussion at https://github.com/clouditor/clouditor/pull/1008#pullrequestreview-1340073160

oxisto:
Why?

anatheka:
Hmm, I would let the engine in codecov and add the missing test, but that are only my two cents.

oxisto:
> Even #1027 fails although only the README is changed. And when I loop through the PR list for potential reviews and see the red X I usually don't look into it because it seems not to be ready for review

The problem is not in the engine.go. It is because the branch is not up-to-date with main and therefore codecov gets mixed up when calculating patch difference.

You can see it here: 
Screenshot 2023-03-29 at 21 45 47

The base commit it is referring to is https://github.com/clouditor/clouditor/commit/995e9390d73a635014fe8861c6f570a7dec640c7, the current main is further along. By comparing it to an old base, it detects a diff change in `engine.go` (which basically was because we ignored it in https://github.com/clouditor/clouditor/pull/1008 and this gets inherited

Once you update #1027 to the current main, the patch difference will be calculated correctly again and there will be no test failure.

lebogg:
> Hmm, I would let the engine in codecov and add the missing test, but that are only my two cents.

You can't test everything in engine.go. That's the point. But since you are both against, I will close it. Thx for your input, I will live on :D

PR #1041, opened 2023-03-29
Fix GetAssessmentResult authz check

anatheka:
Only the "allowed" cloud_services were checked. The check for "all" was missing.


lebogg (approved):
Good catch. Maybe add a small test for the new case?

oxisto:
> Good catch. Maybe add a small test for the new case?

Would be a good idea. I just extended the auth strategy mock to include the "all" flag in https://github.com/clouditor/clouditor/pull/1043. This makes it easier to test it.

lebogg:
Not blocked anymore :+1:

service/orchestrator/assessment_results_test.go
```diff
@@ -83,20 +82,23 @@ func TestService_GetAssessmentResult(t *testing.T) {
...
 				authz: servicetest.NewAuthorizationStrategy(false, []string{orchestratortest.MockAssessmentResult1.CloudServiceId}),
 			},
 			args: args{
-				req: orchestratortest.MockAssessmentResultRequest1,
+				req: orchestratortest.MockAssessmentResultRequest2,
+				ctx: testutil.TestContextOnlyService1,
+			},
+			res: nil,
+			wantErr: func(t assert.TestingT, err error, i ...interface{}) bool {
+				return assert.Contains(t, err.Error(), "access denied")
```
lebogg, line 100:
```suggestion
				return assert.ErrorContains(t, err, service.ErrPermissionDenied.Error())
```

lebogg: Would be more safe if, e.g., the message of `ErrPermissionDenied` would be changed in the future

lebogg (approved):
LGTM.

PR #1043, opened 2023-03-30
Added a test function to create a mocked `AuthorizationStrategy`

### LINKED ISSUE #1038: Add mock for `AuthorizationStrategy` to test utils
Just a minor thing. This could actually be interesting for multiple tests so it might make sense to move this into into the `testutil` package. Otherwise I think we can finally merge this for now.

_Originally posted by @oxisto in https://github.com/clouditor/clouditor/pull/999#discussion_r1147913167_

oxisto:
This fixes #1038. It also tries to restructure the testfolders a little bit.


internal/testutil/servicetest/authz.go
```diff
@@ -0,0 +1,40 @@
+package servicetest
+
+import (
+	"context"
+
+	"clouditor.io/clouditor/internal/api"
+	"clouditor.io/clouditor/service"
```
lebogg, line 7: Shouldn't we try to completely avoid dependencies from the service packages? If yes, we could move the `RequestType` to API?

internal/testutil/servicetest/authz.go
```diff
@@ -0,0 +1,40 @@
...
+		cloudServiceIDs: cloudServiceIDs,
+	}
+}
+
+type AuthorizationStrategyMock struct {
+	all             bool
+	cloudServiceIDs []string
+}
+
+func (a *AuthorizationStrategyMock) CheckAccess(ctx context.Context, _ service.RequestType, req api.CloudServiceRequest) bool {
```
lebogg, line 23: different receiver names: here `a` and later `s`. Maybe merge ot `as`

internal/testutil/servicetest/authz.go
oxisto: Hmm yeah this dependency might cause problems. I am not sure about moving `RequestType` to API, since it's something really only related to the service itself and not used by any of the api / "client" packages. If you really want to remove the dependency, I could move it to `internal/service` and add an alias in `service`, similar to what we do for stuff in `internal/api` / `api`, e.g., `api.PayloadRequest`?

oxisto: In the end, I moved the type to the api package, it really seems to be cleaner.

oxisto: ah! but authorisation strategy is still in the `service` package and definitely belongs there ... So I cannot remove this dependency.

oxisto: I would keep this dependency for now. I think somewhere this year we probably need a more or less complete overhaul of the package structure, because we never really had a good structure in mind and just kept putting things there where they fit and it seems to cause us problems in some of the tests.

lebogg (approved):
LGTM

PR #1054, opened 2023-04-05
Added Discovery status to show Azure and AWS discovery differences

70mm1:


service/discovery/README.md
```diff
@@ -0,0 +1,182 @@
+# Discovery Status
+
+✅: Discovered
+❌: Not Discovered
+🚫: Not available
```
anatheka, line 5: If you want each in a new line you have to add a linkebreak, otherwise they are in one line.

service/discovery/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+## Compute
+<details>
+<summary>Compute</summary>
```
anatheka, line 9: Is it possible to that with the headline? So we have only one Compute and not one above the other.

anatheka (commented):
That looks really good.

anatheka (approved):

PR #1062, opened 2023-04-08
Various test improvements

oxisto:
Further cleanup of variables outside the `testdata` package. Making use of varargs in `NewAuthorizationService`. Consistent use of `NewAuthorizationService` mock.

### DIFF
--- a/internal/testutil/auth.go
+++ b/internal/testutil/auth.go
@@ -6,6 +6,7 @@ import (
 	"net"
 	"net/http"
 
+	"clouditor.io/clouditor/internal/testdata"
 	"github.com/golang-jwt/jwt/v4"
 	oauth2 "github.com/oxisto/oauth2go"
 	"github.com/oxisto/oauth2go/login"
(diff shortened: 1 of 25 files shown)


service/orchestrator/orchestrator.go
```diff
@@ -143,6 +143,12 @@ func WithAuthorizationStrategyJWT(key string, allowAllKey string) ServiceOption
 	}
 }
 
+func WithAuthorizationStrategy(authz service.AuthorizationStrategy) ServiceOption {
```
anatheka, line 146: A test for this function is missing.

internal/testutil/auth.go
```diff
@@ -1,93 +1,22 @@
...
 
-	"github.com/golang-jwt/jwt/v4"
 	oauth2 "github.com/oxisto/oauth2go"
 	"github.com/oxisto/oauth2go/login"
 	"golang.org/x/oauth2/clientcredentials"
-	"google.golang.org/grpc/metadata"
 )
 
 const (
 	TestAuthUser         = "clouditor"
```
anatheka, line 14: Maybe move the consts also to testdata?

anatheka (commented):
Only two minor things.

internal/testutil/auth.go
oxisto: Funnily enough, it was already defined there, I removed the constants from this file.

service/orchestrator/orchestrator.go
oxisto: Added test

anatheka (approved):
LGTM

PR #1075, opened 2023-04-20
Added Identity Discovery to list all Azure Users

70mm1:
TODO
- Check https://learn.microsoft.com/en-us/rest/api/resources/tenants/list?tabs=HTTP


anatheka:
"signInActivity" is only available with Azure AD Premium P1/P2 license [link](https://learn.microsoft.com/en-us/graph/api/resources/user?view=graph-rest-1.0). 

Azure AD Premium P1 5,60€/User [Link](https://www.microsoft.com/de-de/security/business/identity-access/azure-active-directory-pricing)

go.mod
```diff
@@ -157,15 +157,30 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-require github.com/envoyproxy/protoc-gen-validate v0.10.1
+require (
+	github.com/Azure/azure-sdk-for-go v68.0.0+incompatible
```
oxisto, line 161: This seems to re-add the "old" Azure SDK, which is deprecated and we want to avoid. Is there a way to fetch it using the "new" SDK?

anatheka: Not at the moment.

anatheka:
Check https://learn.microsoft.com/en-us/rest/api/resources/tenants/list?tabs=HTTP

lebogg:
Closed because not possible at the moment with our "resources"

PR #1076, opened 2023-04-20
Add authorization to AssessEvidence endpoint

oxisto:


service/assessment/assessment.go
```diff
@@ -242,6 +255,11 @@ func (svc *Service) AssessEvidence(_ context.Context, req *assessment.AssessEvid
 		return nil, status.Errorf(codes.InvalidArgument, "%v", err)
 	}
 
+	// Check if cloud_service_id in the service is within allowed or one can access *all* the cloud services
+	if !svc.authz.CheckAccess(ctx, service.AccessUpdate, req) {
```
lebogg, line 259: Is the context forwarded properly when called in `AssessEvidences` (`_, err = svc.AssessEvidence(context.Background(), assessEvidencesReq)`)?

service/assessment/assessment.go
```diff
@@ -242,6 +255,11 @@ func (svc *Service) AssessEvidence(_ context.Context, req *assessment.AssessEvid
...
+		return nil, service.ErrPermissionDenied
```
lebogg, line 260: Consider adding a test for this case

service/assessment/assessment.go
oxisto: It was not :) good catch. I will forward `stream.Context()` to the call here.

lebogg (approved):

PR #1077, opened 2023-04-21
Add authorization check to certificate endpoints

### LINKED ISSUE #1056: Add authorization to certificate endpoints
**Is your feature request related to a problem? Please describe.**
In the certification endpoints, we don't take into account if the user is authorized.

**Describe the solution you'd like**
Add authorization to all endpoints regarding certification, e.g.  `ListCertificates` and  `GetCertificate`

Should be very straightforward like in the other List requests that got updated previously.

lebogg:
Resolves #1056 

-  Add authorization checks
- Add test cases for introduced authorization checks
- Refactor tests, e.g. make all tests table tests
- Move certification code to separate certificate*.go files
- Add new endpoint: `ListPublicCertificates`


service/orchestrator/certificates.go
```diff
@@ -0,0 +1,152 @@
...
+
+// UpdateCertificate implements method for updating an existing certificate
+func (svc *Service) UpdateCertificate(_ context.Context, req *orchestrator.UpdateCertificateRequest) (response *orchestrator.Certificate, err error) {
+	// Validate request
+	err = service.ValidateRequest(req)
+	if err != nil {
+		return nil, err
+	}
+
+	count, err := svc.storage.Count(req.Certificate, req.Certificate.Id)
```
oxisto, line 113: Missing authz?

service/orchestrator/certificates.go
```diff
@@ -0,0 +1,152 @@
...
+	res = new(orchestrator.ListCertificatesResponse)
+
+	res.Certificates, res.NextPageToken, err = service.PaginateStorage[*orchestrator.Certificate](req, svc.storage,
+		service.DefaultPaginationOpts)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "could not paginate results: %v", err)
+	}
+
+	// Only list certificates user is authorized to see
+	var certificateSlice []*orchestrator.Certificate
```
oxisto, line 92: Instead of doing that, you should append an additional query parameter, that checks if the cloud service ID of the certificate is in a list of allowed services. See other List methods for examples.

service/orchestrator/certificates.go
```diff
@@ -0,0 +1,152 @@
...
+
+// RemoveCertificate implements method for removing a certificate
+func (svc *Service) RemoveCertificate(_ context.Context, req *orchestrator.RemoveCertificateRequest) (response *emptypb.Empty, err error) {
+	// Validate request
+	err = service.ValidateRequest(req)
+	if err != nil {
+		return nil, err
+	}
+
+	err = svc.storage.Delete(&orchestrator.Certificate{}, "Id = ?", req.CertificateId)
```
oxisto, line 142: Missing authz

service/orchestrator/certificates.go
lebogg: Ahh. You mean, first get all allowed cloud services and insert this into a query?

oxisto: Yes

internal/testutil/servicetest/orchestratortest/orchestrator.go
```diff
@@ -11,8 +11,21 @@ import (
...
+	}
+}
+func WithMockServiceID(cloudServiceID string) NewCertificateOption {
+	return func(certificate *orchestrator.Certificate) {
+		certificate.CloudServiceId = cloudServiceID
+	}
+}
+
+// NewCertificate creates a mock certificate.
+func NewCertificate(opts ...NewCertificateOption) *orchestrator.Certificate {
```
lebogg, line 28: - [ ] I will probably replace the `NewCertificate` function with 2 mock certificate variables (directly)

service/orchestrator/certificates.go
```diff
@@ -0,0 +1,172 @@
...
+	"clouditor.io/clouditor/service"
+
+	"github.com/sirupsen/logrus"
+	"golang.org/x/exp/slices"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+const certificationNotFoundErrorMessage = "certificate not found"
```
oxisto, line 19: Shouldn't this be `ErrCertificateNotFound` or something?

service/orchestrator/certificates.go
```diff
@@ -0,0 +1,172 @@
+package orchestrator
```
oxisto, line 1: Unfortunately, it's now hard to see what your changes were because of the changes + extraction, so it might be that some other points come up during this review, which were not your "fault" ;)

service/orchestrator/certificates.go
```diff
@@ -0,0 +1,172 @@
...
+
+const certificationNotFoundErrorMessage = "certificate not found"
+
+// CreateCertificate implements method for creating a new certificate
+func (svc *Service) CreateCertificate(ctx context.Context, req *orchestrator.CreateCertificateRequest) (
+	res *orchestrator.Certificate, err error) {
+
+	// Validate request
+	if err = service.ValidateRequest(req); err != nil {
+		err = status.Error(codes.InvalidArgument, err.Error())
```
oxisto, line 27: The `service.ValidateRequest` method already returns a gRPC error so you must directly return the error here.

service/orchestrator/certificates.go
```diff
@@ -0,0 +1,172 @@
...
+	return
+}
+
+// GetCertificate implements method for getting a certificate, e.g. to show its state in the UI
+func (svc *Service) GetCertificate(ctx context.Context, req *orchestrator.GetCertificateRequest) (
+	res *orchestrator.Certificate, err error) {
+
+	// Validate request
+	if err = service.ValidateRequest(req); err != nil {
+		err = status.Error(codes.InvalidArgument, err.Error())
```
oxisto, line 57: See above, no wrapping

service/orchestrator/certificates.go
```diff
@@ -0,0 +1,172 @@
...
+	res = new(orchestrator.Certificate)
+	err = svc.storage.Get(res, "Id = ?", req.CertificateId)
+	if errors.Is(err, persistence.ErrRecordNotFound) {
+		return nil, status.Errorf(codes.NotFound, certificationNotFoundErrorMessage)
+	} else if err != nil {
+		return nil, status.Errorf(codes.Internal, "database error: %v", err)
+	}
+
+	// Check if client is allowed to access the corresponding cloud service (targeted in the certificate)
+	areAllAllowed, allowedCloudServices := svc.authz.AllowedCloudServices(ctx)
```
oxisto, line 70: we used the variables `all` and `allowed` in other places, would be good to be consistent.

And actually we should append the cloud service check in the query (maybe we should also have a helper function for this), instead of first querying the DB and then looking at it.

service/orchestrator/certificates.go
```diff
@@ -0,0 +1,172 @@
...
+func (svc *Service) ListCertificates(ctx context.Context, req *orchestrator.ListCertificatesRequest) (res *orchestrator.ListCertificatesResponse, err error) {
+	// Validate request
+	err = service.ValidateRequest(req)
+	if err != nil {
+		return nil, err
+	}
+
+	// We only list certificates the user is authorized to see (w.r.t. the cloud service)
+	var conds []any
+	areAllAllowed, cloudServices := svc.authz.AllowedCloudServices(ctx)
```
oxisto, line 89: all, allowed

service/orchestrator/certificates.go
```diff
@@ -0,0 +1,172 @@
...
+// RemoveCertificate implements method for removing a certificate
+func (svc *Service) RemoveCertificate(ctx context.Context, req *orchestrator.RemoveCertificateRequest) (response *emptypb.Empty, err error) {
+	// Validate request
+	err = service.ValidateRequest(req)
+	if err != nil {
+		return nil, err
+	}
+
+	// Only remove certificate if user is authorized for the corresponding cloud service
+	areAllAllowed, allowedCloudService := svc.authz.AllowedCloudServices(ctx)
```
oxisto, line 150: all, allowed

service/orchestrator/certificates.go
```diff
@@ -0,0 +1,172 @@
...
+	// Todo(lebogg to all): Is this too much? A user having admin flag off but no cloud services attached would have full access otherwise. Would mean we have to change other endpoints with auth as well, I guess.
+	// 1st case: The user is neither admin nor is authorized for any cloud service because it's, e.g., a public user.
+	if !areAllAllowed && len(allowedCloudService) == 0 {
+		err = status.Error(codes.PermissionDenied, service.ErrPermissionDenied.Error())
```
oxisto, line 154: `ErrPermissionDenied` is already a gRPC error/status and can/should be returned directly.

service/orchestrator/certificates.go
```diff
@@ -0,0 +1,172 @@
...
+func (svc *Service) RemoveCertificate(ctx context.Context, req *orchestrator.RemoveCertificateRequest) (response *emptypb.Empty, err error) {
+	// Validate request
+	err = service.ValidateRequest(req)
+	if err != nil {
+		return nil, err
+	}
+
+	// Only remove certificate if user is authorized for the corresponding cloud service
+	areAllAllowed, allowedCloudService := svc.authz.AllowedCloudServices(ctx)
+	// Todo(lebogg to all): Is this too much? A user having admin flag off but no cloud services attached would have full access otherwise. Would mean we have to change other endpoints with auth as well, I guess.
```
oxisto, line 151: I am confused. Why should this be? If one has "all" off and the list of cloud services is empty, then the query will be `and cloud_service_id IN ?`, {}; this will always be empty so no certificates should be returned?

service/orchestrator/certificates.go
```diff
@@ -0,0 +1,172 @@
...
+	// 1st case: The user is neither admin nor is authorized for any cloud service because it's, e.g., a public user.
+	if !areAllAllowed && len(allowedCloudService) == 0 {
+		err = status.Error(codes.PermissionDenied, service.ErrPermissionDenied.Error())
+		return
+	} else if !areAllAllowed { // 2nd case: User is authorized for some cloud services (at least one)
+		err = svc.storage.Delete(&orchestrator.Certificate{},
```
oxisto, line 157: A little bit of inconsistent formatting. I would suggest `id` instead of `Id` (field) and `AND` instead of `and` (keyword).

service/orchestrator/certificates.go
lebogg: It was a string which i wrapped into gRPC errors and it is not an error itself. But after thinking about it, I now use a gRPC error directly.
I'm just not sure if such api error messages (=grpc errors) are better kept in the api package in general. Since it only used here, I leave it here.

service/orchestrator/certificates.go
lebogg: 1. Yes, i know and I don't like them :D For me they are inexpressive or even counter-intuitive: `allowed` suggests it is a boolean more than `all`. And i would prefer prefixing bools with is/has/... except they are common like `ok`
2. We could do it but then I guess we can't differentiate between NotFound and PermissionDenied anymore? Maybe there is some DB query trick I'm not aware of.

service/orchestrator/certificates.go
```diff
@@ -0,0 +1,166 @@
...
+
+	// Validate request
+	if err = service.ValidateRequest(req); err != nil {
+		return nil, err
+	}
+
+	// We only list certificates the user is authorized to see (w.r.t. the cloud service)
+	var conds []any
+	areAllAllowed, cloudServices := svc.authz.AllowedCloudServices(ctx)
+	if !areAllAllowed {
```
anatheka, line 90: Here is the PermissionDenied check missing.

server/grpc.go
```diff
@@ -102,6 +103,15 @@ func WithReflection() StartGRPCServerOption {
 	}
 }
 
+//TODO!!!
```
oxisto, line 106: Remove

service/orchestrator/certificates.go
```diff
@@ -0,0 +1,189 @@
...
+	res = new(orchestrator.Certificate)
+	err = svc.storage.Get(res, "Id = ?", req.CertificateId)
+	if errors.Is(err, persistence.ErrRecordNotFound) {
+		return nil, ErrCertificationNotFound
+	} else if err != nil {
+		return nil, status.Errorf(codes.Internal, "database error: %v", err)
+	}
+
+	// Check if client is allowed to access the corresponding cloud service (targeted in the certificate)
+	all, allowed := svc.authz.AllowedCloudServices(ctx)
```
oxisto, line 69: It is a little bit inconsistent that for `Get` we first retrieve the object and then to an authorisation check and for `List` and `Remove` we append the cloud service ID into the query (which I would actually prefer).

I know that in other Get-methods we also do it like this but maybe we should also change it there.

lebogg: The problem is that we then cannot return a permission denied error since we cannot differentiate between the response being empty because it was not found OR because the user actually was unauthorized

service/orchestrator/certificates.go
lebogg: We have to agree on how to do these things. These are the two options I see:
1) Return PermissionDeniedError: This means we would first have to retrieve the certificate, then look at its cloud service ID. If we are allowed, we then call the remove function. Otherwise we return PermissionDeniedError
2) Return NotFoundError: We build it into the DB query (`cloud_service_id IN ?`) which means we get a NotFoundError if the user is not auhorized. It could also means that the certificate is just not there.

@oxisto

anatheka: I'm for the first option. Otherwise, it is not possible to know why the certificate isn't removed. I think the same applies to the ListCertificates method.

service/orchestrator/certificates.go
lebogg: Will be done in future PR

service/orchestrator/certificates.go
oxisto: Well from a security (and privacy) point of view it should be indistinguishable whether the certificate is not there or whether you don’t have access, otherwise you are leaking the existence of a certificate with a certain ID.

@immqu I think we even have a check for this in the privacy property graph, correct?

anatheka: Maybe our access check is not that good. I think the ListCertificates endpoint shouldn't leak any information about certificates. There should be only two cases:
1. You have access and you get the List of certificates, maybe it is empty.
2. You do not have permission and do not receive information about any certificates or whether they exist.

I just worry that we will not recognize permission denied problems, e.g. if there is no cloud service id in the context. Suggestion: We check if areAllAllowed/all == false and the cloudServices/allowed list == empty. If yes, we return a permissionDeniedError, if not we return the list of (empty) certificates.

lebogg: > Well from a security (and privacy) point of view it should be indistinguishable ...
I agree to that

I think in the case of ListXXX methods it is pretty clear: You just get the list of things what you are allowed to see, i.e. in the case of ListCertificate: The list of certificates which are belonging to the cloud services you are allowed to see. Can be empty (there is no certificate yet that you are authorized to see w.r.t. cloud service) , can be all (e.g. admin flag), can be subset (e.g. you are only allowed to see certificates for cloud service 1 and 2 but not the rest)

anatheka: I added an explanation to the ListCertificates method.

anatheka (approved):
FInally :)

PR #1105, opened 2023-05-09
Get context from stream for authorization

### LINKED ISSUE #1104: Use/Forward context of streams for authorization
In `StoreAssessmentResults`, I guess, we should use `stream.Context()` to get the `ctx` from the stream and to forward it properly to the non-stream function `StoreAssessmentResult`
https://github.com/clouditor/clouditor/blob/fa694f08a73683efbe3df0efdc0c766037587844/service/orchestrator/assessment_results.go#L218
And maybe in other places we should check as well. I know it was done in `AssessEvidences`  in a recent PR #1076

lebogg:
Fixes #1104 
Fix it in the following streaming RPCs:
- `StoreAssessmentResults`
- `StoreEvidences`


anatheka (approved):
:+1:

lebogg (commented):
LGTM

PR #1119, opened 2023-05-22
Include raw discovery traffic in raw property of an evidence

anatheka:
This PR adds the resource object from the Cloud Provider as a json string to the evidence raw field.


service/discovery/azure/network_test.go
```diff
@@ -350,6 +353,7 @@ func Test_azureNetworkDiscovery_List(t *testing.T) {
 								},
 								Type:   voc.LoadBalancerType,
 								Labels: map[string]string{},
+								Raw:    "{\"*armnetwork.LoadBalancer\":[{\"id\":\"/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/res1/providers/Microsoft.Network/loadBalancers/lb3\",\"location\":\"eastus\",\"name\":\"lb3\",\"properties\":{\"frontendIPConfigurations\":[{\"id\":\"/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/res1/providers/Microsoft.Network/loadBalancers/lb1/frontendIPConfigurations/b9cb3645-25d0-4288-910a-020563f63b1c\",\"name\":\"b9cb3645-25d0-4288-910a-020563f63b1c\",\"properties\":{\"publicIPAddress\":{\"id\":\"/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/res1/providers/Microsoft.Network/publicIPAddresses/test-b9cb3645-25d0-4288-910a-020563f63b1d\"}}}],\"loadBalancingRules\":[{\"properties\":{\"frontendPort\":1234}},{\"properties\":{\"frontendPort\":5678}}]}}]}",
```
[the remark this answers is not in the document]
anatheka: Thanks for the tip, but I will not change that.

service/discovery/aws/compute.go
```diff
@@ -158,6 +160,13 @@ func (d *computeDiscovery) discoverVolumes() ([]*voc.BlockStorage, error) {
 			atRest.Algorithm = "AES-256"
 		}
 
+		// Convert object responses from Azure to string
+		rawInfo = voc.AddRawInfo(rawInfo, &res.Volumes[i])
+		raw, err := voc.ToStringInterface(rawInfo)
```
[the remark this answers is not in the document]
anatheka: Good catch, I fixed it.

api/discovery/discovery.go
```diff
@@ -51,7 +51,7 @@ type Authorizer interface {
 }
 
 // NewResource creates a new voc resource.
-func NewResource(d Discoverer, ID voc.ResourceID, name string, creationTime *time.Time, location voc.GeoLocation, labels map[string]string, typ []string) *voc.Resource {
+func NewResource(d Discoverer, ID voc.ResourceID, name string, creationTime *time.Time, location voc.GeoLocation, labels map[string]string, typ []string, raw string) *voc.Resource {
```
oxisto, line 54: Suggestion: Instead of directly specifying the raw string here, could we add a varargs any to the function like "rawObjects" and then use your voc.StringInterface function thingie to convert all the raw objects to a string and assign it here.

This would remove a lot of boiler-plate code out of the individual discovery functions. They would simply need to append the raw objects they want to consider when calling `NewResource`.

api/discovery/discovery.go
```diff
@@ -52,7 +57,12 @@ type Authorizer interface {
 }
 
 // NewResource creates a new voc resource.
-func NewResource(d Discoverer, ID voc.ResourceID, name string, creationTime *time.Time, location voc.GeoLocation, labels map[string]string, typ []string) *voc.Resource {
+func NewResource(d Discoverer, ID voc.ResourceID, name string, creationTime *time.Time, location voc.GeoLocation, labels map[string]string, typ []string, raw map[string][]interface{}) *voc.Resource {
```
oxisto, line 60: Almost there... just wondering... Why is "raw" a map and not just a list? If it would be a list, we could just use varargs and use it like `NewResource(..., myAzureResource, myOtherResource)` and avoid the `raw := make map` stuff.

anatheka: I thought it is better if we have a map with the object type as a key string. Otherwise it is difficult to understand (depending on the cloud) which entry contains the raw data for which object type.

oxisto (approved):

PR #1124, opened 2023-06-01
Update UpdateMetricConfiguration()

anatheka:
- Delete 2 unnecessary assignments


anatheka:
@oxisto @lebogg after an UpdateMetricConfiguration() the is_default field is no longer available but the timestamp field is when you call GetMetricConfiguration(). Is that the way it's supposed to be?

oxisto:
> @oxisto @lebogg after an UpdateMetricConfiguration() the is_default field is no longer available but the timestamp field is when you call GetMetricConfiguration() via grpcurl. Is that the way it's supposed to be?
> 
> With curl all fields are available. Or should we ignore that?

What do you mean with "is available"? It could be that grpcurl hides boolean fields with the "false" value, this is basically just a display issue. The important thing is that isDefault is really set to false.

anatheka:
> > @oxisto @lebogg after an UpdateMetricConfiguration() the is_default field is no longer available but the timestamp field is when you call GetMetricConfiguration() via grpcurl. Is that the way it's supposed to be?
> > With curl all fields are available. Or should we ignore that?
> 
> What do you mean with "is available"? It could be that grpcurl hides boolean fields with the "false" value, this is basically just a display issue. The important thing is that isDefault is really set to false.

Ah, I found it. The JSON tag includes an omitempty qualifier, so it is not shown. Thanks!

service/orchestrator/metrics_test.go
```diff
@@ -526,18 +600,100 @@ func TestService_GetMetric(t *testing.T) {
...
+									structpb.NewBoolValue(false),
+									structpb.NewBoolValue(true),
+								}}}},
+					})
+				}),
+			},
+			args: args{
+				req: &orchestrator.ListMetricsRequest{},
+			},
+			wantRes: &orchestrator.ListMetricsResponse{
```
anatheka, line 654: @oxisto @lebogg why is this test failing? I have no idea what the reason is.

anatheka: FIxed.

service/orchestrator/metrics_test.go
```diff
@@ -47,9 +42,40 @@ import (
 	"clouditor.io/clouditor/persistence"
 	"clouditor.io/clouditor/persistence/gorm"
 	"clouditor.io/clouditor/service"
+	"github.com/stretchr/testify/assert"
```
oxisto, line 45: minor: missing line separator between clouditor and the rest of the imports

service/orchestrator/metrics_test.go
```diff
@@ -547,8 +706,19 @@ func TestService_GetMetricImplementation(t *testing.T) {
...
 		args    args
 		wantRes *assessment.MetricImplementation
-		wantErr bool
+		wantErr assert.ErrorAssertionFunc
 	}{
+		{
+			name:    "Invalid input",
+			wantRes: nil,
+			wantErr: func(tt assert.TestingT, err error, i ...interface{}) bool {
+				gotStatus, ok := status.FromError(err)
```
oxisto, line 716: minor: Since this seems to repeat a lot of times, would it make sense to build a function that we could directly use as as `assert.ErrorAssertionFunc? Something like:
```go
func wantStatusCode(code codes.Code) assert.ErrorAssertionFunc {
	return func(t assert.TestingT, err error, i ...interface{}) bool {
		gotStatus, ok := status.FromError(err)
		if !ok {
			return false
		}
		return assert.Equal(t, gotStatus.Code(), code)
	}
}
```

this can then be used directly:
```go
...
wantErr: wantStatusCode(codes.InvalidArgument)
```

oxisto (approved):

PR #1125, opened 2023-06-02
Add tool_id to assessment_result and filter by tool_id

anatheka:
This PR adds 
- [x] a field tool_id  to the assessment_result and
- [x] adds the filter tool_id to ListAssessmentResults.


api/assessment/assessment.proto
```diff
@@ -134,6 +134,9 @@ message AssessmentResult {
 
   // The cloud service which this assessment result belongs to
   string cloud_service_id = 10 [ (validate.rules).string.uuid = true ];
+
+  // Reference to the tool which provided the assessment result
+  string tool_id = 11 [ (validate.rules).string.min_len = 1 ];
```
oxisto, line 139: Does making this mandatory create problems in MEDINA deployments or do we already specify it there anyway?

oxisto (approved):

api/assessment/assessment.proto
anatheka: It will create problems... We could remove the validation till Medina is over.

oxisto: Or make it an optional field but leave the validation already in place.

PR #1132, opened 2023-06-12
Fix test in service/assessment for the case Clouditor is running

anatheka:
This PR fixes a test, which fails if the Clouditor is already running.


service/assessment/assessment_test.go
```diff
@@ -945,7 +945,8 @@ func TestService_HandleEvidence(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			s := NewService()
+			// We need anohter evidence store address, otherwise the test fails if Clouditor is already running.
```
oxisto, line 948: As discussed offline, we should change the test to directly init the RPC connection in the individual tests instead of the `hasRPCConnection` variable. See `TestService_AssessEvidence` for an example.

oxisto (approved):

PR #1144, opened 2023-06-27
Extend discovery based on Medina issues

anatheka:
This PR updates/adds the following discovered resources properties:
- Update function runtime/version 
- Fix OsLoggingEnabled
- Add ActivityLoggingEnabled

### DIFF
--- a/service/discovery/azure/compute.go
+++ b/service/discovery/azure/compute.go
@@ -124,19 +124,19 @@ func (d *azureComputeDiscovery) List() (list []voc.IsCloudResource, err error) {
 	list = append(list, virtualMachines...)
 
 	// Discover functions and web apps
-	function, err := d.discoverSites()
+	resources, err := d.discoverFunctionsWebApps()
 	if err != nil {
 		return nil, fmt.Errorf("could not discover functions: %w", err)
 	}
-	if function != nil {
-		list = append(list, function...)
-	}
+	// if resources != nil {
+	list = append(list, resources...)
+	// }
 
 	return
 }
 
-// Discover function
-func (d *azureComputeDiscovery) discoverSites() ([]voc.IsCloudResource, error) {
+// Discover functions and web apps
+func (d *azureComputeDiscovery) discoverFunctionsWebApps() ([]voc.IsCloudResource, error) {
 	var list []voc.IsCloudResource
 
 	// initialize functions client
@@ -146,8 +146,8 @@ func (d *azureComputeDiscovery) discoverSites() ([]voc.IsCloudResource, error) {
 
 	// List functions
 	err := listPager(d.azureDiscovery,
-		d.clients.webAppsClient.NewListPager,
-		d.clients.webAppsClient.NewListByResourceGroupPager,
+		d.clients.sitesClient.NewListPager,
+		d.clients.sitesClient.NewListByResourceGroupPager,
 		func(res armappservice.WebAppsClientListResponse) []*armappservice.Site {
 			return res.Value
 		},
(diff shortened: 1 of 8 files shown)


service/discovery/azure/compute.go
```diff
@@ -272,6 +342,18 @@ func (d *azureComputeDiscovery) handleVirtualMachines(vm *armcompute.VirtualMach
...
 	}
 
+	// Check extensions
+	for _, extension := range vm.Resources {
+		// Azure Monitor Agent (AMA) collects monitoring data from the guest operating system of Azure and hybrid virtual machines and delivers it to Azure Monitor for use (https://learn.microsoft.com/en-us/azure/azure-monitor/agents/agents-overview). The extension names are
+		// * OMSAgentForLinux for Linux VMs and (legacy agent)
+		// * MicrosoftMonitoringAgent for Windows VMs (legacy agent)
+		// * AzureMonitoringWindowsAgent (new agent)
+		// * AzureMonitoringLinuxAgent (new agent)
+		if strings.Contains(*extension.ID, "OmsAgentForLinux") || strings.Contains(*extension.ID, "MicrosoftMonitoringAgent") || strings.Contains(*extension.ID, "AzureMonitoringWindowsAgent") || strings.Contains(*extension.ID, "AzureMonitoringLinuxAgent") {
```
lebogg, line 352: I guess there are no constants in the azure sdk defining these strings?

service/discovery/azure/compute.go
```diff
@@ -303,16 +385,23 @@ func (d *azureComputeDiscovery) handleVirtualMachines(vm *armcompute.VirtualMach
...
 					SecurityFeature: &voc.SecurityFeature{},
 				},
 				MonitoringLogDataEnabled: monitoringLogDataEnabled,
 				SecurityAlertsEnabled:    monitoringLogDataEnabled,
 			},
 		},
+		ActivityLogging: &voc.ActivityLogging{
+			Logging: &voc.Logging{
+				Enabled:         true,               // is always enabled
+				RetentionPeriod: 90,                 // always 90 days
```
lebogg, line 401: Maybe consider using a constant for retentionperiod

service/discovery/azure/compute.go
```diff
@@ -123,41 +123,71 @@ func (d *azureComputeDiscovery) List() (list []voc.IsCloudResource, err error) {
...
+			var r voc.IsCompute
+
+			// Check kind of site (see https://github.com/Azure/app-service-linux-docs/blob/master/Things_You_Should_Know/kind_property.md)
+			switch *site.Kind {
+			case "app": // Windows Web App
+			// TODO(all): TBD
+			case "app,linux": // Linux Web app
+			// TODO(all): TBD
+			case "app,linux,container": // Linux Container Web app
+			// TODO(all): TBD
```
lebogg, line 167: Add test cases for these, maybe to check that `r` is nil/empty

service/discovery/azure/compute.go
```diff
@@ -123,41 +123,71 @@ func (d *azureComputeDiscovery) List() (list []voc.IsCloudResource, err error) {
 	}
 	list = append(list, virtualMachines...)
 
-	// Discover functions
-	function, err := d.discoverFunctions()
```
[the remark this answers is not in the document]
anatheka: I'm not sure. At the end we discover functions and web apps and sites is only a step in between.

service/discovery/azure/compute.go
anatheka: No.

service/discovery/azure/compute.go
anatheka: That is a log of overhead because I have to add a function for each case in the Do method. I think that is out of proportion and ok if we only check 2 cases.

service/discovery/azure/compute.go
```diff
@@ -123,41 +123,71 @@ func (d *azureComputeDiscovery) List() (list []voc.IsCloudResource, err error) {
...
+			// TODO(all): TBD
+			case "functionapp": // Function Code App
+				r = d.handleFunction(site)
+			case "functionapp,linux": // Linux Consumption Function app
+				r = d.handleFunction(site)
+			case "functionapp,linux,container,kubernetes": // Function Container App on ARC
+			// TODO(all): TBD
+			case "functionapp,linux,kubernetes": // Function Code App on ARC
+				// TODO(all): TBD
+			}
```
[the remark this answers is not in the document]
anatheka: I added a default case with a debug message.  If we are in the default case it is not unnecessarily an error, so I think the log message is enough.

service/discovery/azure/compute.go
```diff
@@ -123,41 +123,71 @@ func (d *azureComputeDiscovery) List() (list []voc.IsCloudResource, err error) {
...
-	function, err := d.discoverFunctions()
+	// Discover functions and web apps
+	function, err := d.discoverSites()
 	if err != nil {
 		return nil, fmt.Errorf("could not discover functions: %w", err)
 	}
-	list = append(list, function...)
+	if function != nil {
+		list = append(list, function...)
+	}
```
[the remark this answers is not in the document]
anatheka: This is a relic. :)

lebogg (approved):

PR #1148, opened 2023-07-03
Add metadata to cloud service/catalog

anatheka:
This PR adds metadata for cloud services and catalogs, mainly to be consumed by the UI.

It also adds some improvements to catalog loading:

- The respective tests should now load from a "test" catalog instead of the demo one, this makes the tests more robust when additional files are present in the `catalogs` folder

- Loading a catalog now sets the parent information, which removes a lot of redundant parentXXX lines from the JSON files


api/orchestrator/orchestrator.proto
```diff
@@ -657,17 +657,18 @@ message CloudService {
 
   message Metadata {
     // a map of key/value pairs, e.g., env:prod
-   map<string, string> labels = 1;
+   map<string, string> labels = 1 ;
    
    // an icon for the cloud service used by the UI
    string icon = 2;
```
oxisto, line 663: I would probably make all of them optional and avoid defining a "default color" in the backend. This is something we can set as a default in the frontend.

api/orchestrator/orchestrator.proto
```diff
@@ -655,16 +655,20 @@ message CloudService {
...
-    map<string, string> tag   = 1;
+  message Metadata {
+    // a map of key/value pairs, e.g., env:prod
+   map<string, string> labels = 1 ;
+   
+   // an icon for the cloud service used by the UI
+   string icon = 2;
+  
+   // a color for the cloud service used by the UI
+   string color = 3;
```
oxisto, line 666: The color was part of the catalog, not the cloud service. We can probably re-use the same metadata message for catalog and cloud service?

api/orchestrator/orchestrator.proto
```diff
@@ -655,16 +655,20 @@ message CloudService {
     (validate.rules).repeated .items.message.required = true
   ];
 
-  // As long as repeated maps are not supported, we have to do this workaround. 
-  message Tag {
-    map<string, string> tag   = 1;
+  message Metadata {
```
oxisto, line 658: A "created_at" property would also be good, so that we know when the cloud service was created. Not sure if this would be part of the metadata property or directly a property of the service.

anatheka:
In the proto file we have defined that there must be at least 3 assurance levels. Do we want to change this? Or adjust the catalog?

anatheka:
> In the proto file we have defined that there must be at least 3 assurance levels. Do we want to change this? Or adjust the catalog?

We can delete the minimum, since the code can now handle any number of assurance levels.

oxisto (approved):
LGTM from my side

@anatheka I did some additional cleanup in the tests and loading of catalogs:
* The respective tests should now load from a "test" catalog instead of the demo one, I think you already had this initial idea and I rejected it, but now I get why it made sense. I only added a very small test catalog though
* Loading a catalog now sets the parent information, which removes a lot of redundant parentXXX lines from the JSON files

PR #1149, opened 2023-07-04
Add statistics endpoint

anatheka:


api/orchestrator/orchestrator.proto
```diff
@@ -184,6 +184,14 @@ service Orchestrator {
     };
   }
 
+  // Retrieves target cloud service statistics
+  rpc CloudServiceStatistics(CloudServiceStatisticsRequest)
```
oxisto, line 188: rename to `GetCloudServiceStatistics`?

api/orchestrator/orchestrator.proto
```diff
@@ -567,6 +575,34 @@ message ListCloudServicesResponse {
...
+  optional int64 number_of_assessment_results = 2;
+  
+  // number of evidences per cloud services 
+  optional int64 number_of_evidences = 3;
+
+  // number of selected catalogs per cloud services 
+  optional int64 number_of_selected_catalogs = 4;
+
+  // number of available cloud services 
+  optional int64 number_of_cloud_services = 5;
```
oxisto, line 596: I think the number of metrics, catalogs and cloud service is not necessarily needed as a statistic because we are fetching them in the global layout of the page anyway because we need them in so many places.

If we want them, we might want to move them to a separate "global" statistics message?

Also I think we can get rid of the optionals. In the worst case, the number is just 0.

oxisto (commented):
Thanks! Some small discussions points

api/orchestrator/orchestrator.proto
anatheka: I thought we can use the statistics endpoint for global and cloud service specific statistics. If a cloud service ID is specified, the global and specific statistics are returned, otherwise only the global statistics.

api/orchestrator/orchestrator.proto
anatheka: See answer below. Maybe we should use GetStatistics() and move it to orchestrator.go?

api/orchestrator/orchestrator.proto
oxisto: > I thought we can use the statistics endpoint for global and cloud service specific statistics. If a cloud service ID is specified, the global and specific statistics are returned, otherwise only the global statistics.

Seems error-prone :) I would prefer two different end-points

oxisto (approved):

PR #1151, opened 2023-07-06
Stopping gRPC server when hitting ctrl+c

### LINKED ISSUE #1141: Cannot stop clouditor engine in terminal
We should be able to stop/kill a running clouditor engine on the terminal with ctrl+c, but somehow it is either ignored or the engine hangs. Probably a deadlock in waiting for a channel.

oxisto:
The gRPC server was still waiting for some connections when hitting ctrl+c to shutdown. This was because we tried to gracefully shutdown, which did not work because streams were not closed. Now streams are closed properly, but we also need to use Stop() because somehow some connections are still active.

Closes #1141

### DIFF
--- a/api/streams.go
+++ b/api/streams.go
@@ -141,6 +141,13 @@ func (s *StreamsOf[StreamType, MsgType]) GetStream(target string, component stri
 	return c, nil
 }
 
+// CloseAll closes all streams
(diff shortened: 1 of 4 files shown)


anatheka (approved):
LGTM

PR #1161, opened 2023-07-10
Re-design of the evaluation service

### LINKED ISSUE #1154: Design proposal for manual and delegated evaluation results
The current way we deal with delegated and out of scope controls for a target of evaluation has several problems. Currently, we have a list of `controls_in_scope` for each target of evaluation, we somehow need to fill. 

- From a UI point of view, it is unnecessarily complicated to have a view to select and deselect possible controls from a catalog
- From a real-world perspective, not many catalogs even allow scoping or at least only allow to delegate or make only a small list of controls out of scope
- From a technical perspective, we are running into sync problems, if catalogs are ever updated and we would then need to update the `controls_in_scope`.

I therefore propose to remove the `controls_in_scope` property and instead consider all controls (of a given assurance level) in the evaluation service.

To make controls out of scope, delegate them or make them manually compliant, we can then allow to manually create an `EvaluationResult` (via API, UI). We can then introduce additional enums into the compliance state, namely:

- `EVALUATION_STATUS_DELEGATED` for controls that are delegated
- `EVALUATION_STATUS_OUT_OF_SCOPE` for controls to be considered out of scope (if the catalog allows this)
- `EVALUATION_STATUS_COMPLIANT_MANUALLY` for controls that are set to be compliant manually (we could also just use `EVALUATION_STATUS_COMPLIANT`)

oxisto:
This PR does a further cleanup of the evaluation service by removing the `controls_in_scope` property from the ToE.

Using errgroup instead of second layer of scheduler

Added status for manually non compliant and compliant

This implements the design discussion in #1154

Closes #1154


.vscode/launch.json
```diff
@@ -37,7 +37,10 @@
             "args": [
                 "--db-in-memory",
                 "--dashboard-url=http://localhost:5173",
-                "--api-cors-allowed-origins=http://localhost:5173"
+                "--api-cors-allowed-origins=http://localhost:5173",
+                "--discovery-auto-start",
+                "--discovery-provider=azure",
+                "--discovery-resource-group=DemoClouditorHappy"
```
anatheka, line 43: This should be removed.

api/orchestrator/orchestrator.proto
```diff
@@ -1079,6 +942,14 @@ message ListControlsRequest {
   string catalog_id = 1;
   string category_name = 2;
 
+  message Filter {
+    // Optional. Lists only controls with the specified assurance levels.
+    repeated string assurance_levels = 1
+        [ (validate.rules).repeated .items.string.min_len = 1 ];
```
anatheka, line 948:
```suggestion
        [ (validate.rules).repeated.items.string.min_len = 1 ];
```

clouditor.code-workspace
```diff
@@ -54,6 +55,7 @@
 			"sirupsen",
 			"stretchr",
 			"structpb",
+			"Subcontrol",
```
anatheka, line 58: What is Subcontrol?

internal/testutil/servicetest/orchestratortest/orchestrator.go
```diff
@@ -105,28 +105,6 @@ func NewTargetOfEvaluation(assuranceLevel string) *orchestrator.TargetOfEvaluati
 		toe.AssuranceLevel = &assuranceLevel
 	}
 
-	// Our test catalog does not allow scoping, so we need to emulate what we do in CreateTargetOfEvaluation
```
anatheka, line 108: I think the function "NewTargetOfEvaluationWithoutControlsInScope" and all the other MockControlsInScope* can be deleted.

service/orchestrator/catalogs.go
```diff
@@ -188,16 +192,27 @@ func (srv *Service) ListControls(_ context.Context, req *orchestrator.ListContro
...
 	}
+	if req.CatalogId != "" {
+		query = append(query, "category_catalog_id = ?")
+		args = append(args, req.CatalogId)
+	}
+
+	// Apply additional filter
+	if req.Filter != nil {
+		if len(req.Filter.AssuranceLevels) > 0 {
+			query = append(query, "(assurance_level IN ? OR assurance_level IS NULL)")
```
anatheka, line 206: Why do we need here the part "OR assurance_level is NULL"? I think here it is not possible that the assuranceLevel is null.

service/evaluation/evaluation.go
```diff
@@ -427,10 +418,28 @@ func (svc *Service) ListEvaluationResults(ctx context.Context, req *evaluation.L
...
 }
 
+// CreateEvaluationResult is a method implementation of the assessment interface
+func (svc *Service) CreateEvaluationResult(ctx context.Context, req *evaluation.CreateEvaluationResultRequest) (res *evaluation.EvaluationResult, err error) {
+	// Validate request
+	err = service.ValidateRequest(req)
+	if err != nil {
+		return nil, err
+	}
+
```
anatheka, line 428: Missing CheckAccess().

service/evaluation/evaluation.go
```diff
@@ -130,7 +132,7 @@ func NewService(opts ...service.Option[Service]) *Service {
 	svc := Service{
 		orchestrator:    api.NewRPCConnection(DefaultOrchestratorAddress, orchestrator.NewOrchestratorClient),
 		scheduler:       make(map[string]*gocron.Scheduler),
-		wg:              make(map[string]*sync.WaitGroup),
+		wg:              make(map[string]*semaphore.Weighted),
```
anatheka, line 135: Delete.

clouditor.code-workspace
oxisto: a part of `evaluateSubcontrol` :D

api/orchestrator/orchestrator.proto
oxisto: My auto-formatting always does this :( The problem is that this will return once I save the file again.

anatheka: OK, then I will ignore it in future.

service/orchestrator/catalogs.go
oxisto: I think the logic was that if no assurance level is specified (e.g. on the top level), the control is valid for all assurance levels.

service/evaluation/evaluation.go
```diff
@@ -444,80 +434,200 @@ func (svc *Service) addJobToScheduler(c *orchestrator.Control, toe *orchestrator
...
+		}
+
+		if c.IsRelevantFor(toe, catalog) {
+			relevant = append(relevant, c)
+		}
+
+		// If we ignore the control, we can skip it
+		if slices.Contains(ignored, c.Id) {
+			continue
+		}
```
anatheka, line 521: Don't the two functions have to be swapped

service/evaluation/evaluation.go
```diff
@@ -444,80 +434,200 @@ func (svc *Service) addJobToScheduler(c *orchestrator.Control, toe *orchestrator
...
+	var (
+		status   = evaluation.EvaluationStatus_EVALUATION_STATUS_PENDING
+		result   *evaluation.EvaluationResult
+		results  []*evaluation.EvaluationResult
+		relevant []*orchestrator.Control
+		ignored  []string
+		g        *errgroup.Group
+	)
+
+	// Gather a list of sub control IDs that we have manual results for and thus we are ignoring
```
anatheka, line 569: Is it possible to outsource this to a new function? It is duplicated code with few exceptions which is also used in evaluateCatalog().

service/evaluation/evaluation.go
```diff
@@ -526,38 +636,37 @@ func (svc *Service) evaluateControl(toe *orchestrator.TargetOfEvaluation, catego
 	}
 
 	// Create evaluation result
-	evalResult = &evaluation.EvaluationResult{
+	result = &evaluation.EvaluationResult{
 		Id:                         uuid.NewString(),
 		Timestamp:                  timestamppb.Now(),
-		ControlCategoryName:        categoryName,
-		ControlId:                  controlId,
+		ControlCategoryName:        control.CategoryName,
```
anatheka, line 642: Why don't you use the GetX() functions here?

service/evaluation/evaluation.go
```diff
@@ -526,38 +636,37 @@ func (svc *Service) evaluateControl(toe *orchestrator.TargetOfEvaluation, catego
...
 		ControlCatalogId:           toe.GetCatalogId(),
 		Status:                     status,
 		FailingAssessmentResultIds: nonCompliantAssessmentResults,
 	}
 
-	err = svc.storage.Create(evalResult)
+	err = svc.storage.Create(result)
 	if err != nil {
-		log.Errorf("error storing evaluation result for control ID '%s' in database: %v", controlId, err)
+		log.Errorf("error storing evaluation result for control ID '%s' in database: %v", control.Id, err)
```
anatheka, line 652: Extend with the CloudServiceID for a better debugging.

service/evaluation/evaluation.go
```diff
@@ -592,7 +696,7 @@ func (svc *Service) evaluateSubcontrol(toe *orchestrator.TargetOfEvaluation, con
 			// We let the scheduler running if we do not get the assessment results from the orchestrator, maybe it is
 			// only a temporary network problem
 			log.Errorf("could not get assessment results for Cloud Service ID '%s' and MetricIds '%s' from Orchestrator: %v", toe.GetCloudServiceId(), getMetricIds(metrics), err)
-		} else if len(assessmentResults) == 0 {
+		} else if len(assessments) == 0 {
 			// We let the scheduler running if we do not get the assessment results from the orchestrator, maybe it is
 			// only a temporary network problem
```
anatheka, line 701: I think in this case no assessment results are available an no network errors are present.

service/evaluation/evaluation.go
```diff
@@ -644,15 +738,12 @@ func (svc *Service) evaluateSubcontrol(toe *orchestrator.TargetOfEvaluation, con
 	err = svc.storage.Create(eval)
 	if err != nil {
 		log.Errorf("error storing evaluation result for control ID '%s' in database: %v", control.Id, err)
+		return nil, err
 	}
 
-	log.Debugf("Evaluation result stored for ControlID '%s' and Cloud Service ID '%s' with ID '%s'.", control.Id, toe.GetCloudServiceId(), eval.Id)
+	log.Infof("Evaluation result for %s (in cloud service %s) was %s", control.Id, toe.GetCloudServiceId(), eval.Status.String())
```
anatheka, line 744:
```suggestion
	log.Infof("Evaluation result for control ID %s (in cloud service %s) was %s", control.GetId(), toe.GetCloudServiceId(), eval.Status.String())
```

api/orchestrator/orchestrator.go
```diff
@@ -180,7 +157,24 @@ func (req *RemoveTargetOfEvaluationRequest) GetPayload() proto.Message {
...
+		return true
+	}
+
+	// If the control does not explicitly specify an assurance level, we are also ok
+	if c.AssuranceLevel == nil || toe.AssuranceLevel == nil {
+		return true
+	}
+
+	// Otherwise, we need to retrieve the possible assurance levels (in order) from the catalogs and compare the
+	// indicies
```
anatheka, line 175:
```suggestion
	// indices
```

anatheka (commented):
First review, i need to look at it again, it's so much :)
LGTM, only some minor things.

api/orchestrator/orchestrator.go
```diff
@@ -180,7 +157,24 @@ func (req *RemoveTargetOfEvaluationRequest) GetPayload() proto.Message {
...
+	idxControl := slices.Index(catalog.AssuranceLevels, *c.AssuranceLevel)
+	idxToe := slices.Index(catalog.AssuranceLevels, *toe.AssuranceLevel)
+
+	return idxControl <= idxToe
```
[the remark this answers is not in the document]
oxisto: Because a "higher" assurance level comprises the controls of a lower one

service/evaluation/evaluation.go
```diff
@@ -185,94 +184,56 @@ func (svc *Service) StartEvaluation(ctx context.Context, req *evaluation.StartEv
 	}
 
 	// Get all Controls from Orchestrator for the evaluation
-	err = svc.cacheControls(req.GetCatalogId())
+	err = svc.cacheControls(req.CatalogId)
```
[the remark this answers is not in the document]
oxisto: there is no need for the extra getter is there? we can just directly access the field

service/evaluation/evaluation.go
```diff
@@ -185,94 +184,56 @@ func (svc *Service) StartEvaluation(ctx context.Context, req *evaluation.StartEv
...
 
-	// Check if scheduler for ToE has started already
-	if svc.scheduler != nil && svc.scheduler[schedulerTag] != nil && svc.scheduler[schedulerTag].IsRunning() {
+	// Check, if a previous job exists and/or is running
+	jobs, err = svc.scheduler.FindJobsByTag(req.CloudServiceId, req.CatalogId)
+	if err != nil && !errors.Is(err, gocron.ErrJobNotFoundWithTag) {
+		err = fmt.Errorf("error while retrieving existing scheduler job: %w", err)
+		log.Error(err)
+		return nil, status.Errorf(codes.Internal, "%s", err)
+	} else if len(jobs) > 0 {
```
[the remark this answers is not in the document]
oxisto: it is an error, because if we cannot retrieve the scheduler jobs, we cannot determine whether it is already running or not.

service/evaluation/evaluation.go
oxisto: as a general rule I would say that its always faster to directly access the property instead of the getter? Accessing via the getter is only convinient when accessing optional fields which could potentially be nil.

service/evaluation/evaluation.go
```diff
@@ -427,12 +387,42 @@ func (svc *Service) ListEvaluationResults(ctx context.Context, req *evaluation.L
...
+		return nil, status.Errorf(codes.InvalidArgument, "validity must be set")
+	}
+
+	// Check, if this request has access to the cloud service according to our authorization strategy.
+	if !svc.authz.CheckAccess(ctx, service.AccessCreate, req) {
+		return nil, service.ErrPermissionDenied
+	}
+
+	res = req.Result
+	res.Id = uuid.NewString()
```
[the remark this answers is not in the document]
oxisto: Not quite sure what you mean here; In case of evidences and assessment results we need to generate the UUID externally because both are created externally and refer to each other.

service/evaluation/evaluation.go
oxisto: I will try to do that in a future iteration

PR #1165, opened 2023-07-11
Added missing context in Orchestrator hook

oxisto:
Added missing context in Orchestrator hook

### DIFF
--- a/service/assessment/benchmark_test.go
+++ b/service/assessment/benchmark_test.go
@@ -63,7 +63,7 @@ func createEvidences(n int, m int, b *testing.B) int {
 
 	svc := NewService(WithOrchestratorAddress(addr), WithEvidenceStoreAddress(addr))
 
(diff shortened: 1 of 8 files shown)


service/discovery/azure/compute.go
```diff
@@ -232,25 +232,25 @@ func (d *azureComputeDiscovery) handleFunction(function *armappservice.Site) voc
 		}
 
 		// Check all runtime versions to get the used runtime language and runtime version
-		if config.Properties.JavaVersion != nil {
+		if *config.Properties.JavaVersion != "" {
```
oxisto, line 235: Does that make sense?! Won't this crash if the property is not set? shouldn't we use our util.Deref here at least?

oxisto: Also, I wouldn't change too many things in this PR that have nothing to do with the original "problem".

PR #1175, opened 2023-07-24
QuickFix for UpdateMetricImplementation

anatheka:
**Only merge in Medina branch**

In Medina it is not possible to update a metric_implementation if it not already exists. The following error is returned: 
```
ERROR:
  Code: Unknown
  Message: could not save metric implementation: ERROR: there is no unique or exclusion constraint matching the ON CONFLICT specification (SQLSTATE 42P10)
```
Locally it works with db_in_memory as well with a postgres DB.

As a quick fix I updated the Save() method in persistence/gorm/gorm.go. It uses now Save() and Create() instead of only Save(). If Save() returns an error it is tried to use Create().

### DIFF
--- a/persistence/gorm/gorm.go
+++ b/persistence/gorm/gorm.go
@@ -298,6 +298,7 @@ func (s *storage) Count(r any, conds ...any) (count int64, err error) {
 }
 
 func (s *storage) Save(r any, conds ...any) error {
+	// Save the record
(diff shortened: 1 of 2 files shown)


anatheka:
This PR is not needed anymore, the error was due to erroneous DB constraints in the used DB.

PR #1176, opened 2023-07-26
Some rego updates

anatheka:
- [x] Add resource_type check to L3FirewallEnabled
- [x] Add resource_type check to WebApplicationFirewallEnabled
- [x] Add operators compare function isIn for array of actual_values and string of target_value


policies/bundles/L3FirewallEnabled/metric.rego
```diff
@@ -13,5 +13,6 @@ applicable {
 }
 
 compliant {
+	compare("isIn",  "NetworkInterface", input.type)
```
lebogg, line 16: When I think about it, I'm for putting the `input.type` check to the applicable rule. Otherwise, the metric is applied to L3WebFirewall as well what we want to avoid here in the first place.

policies/bundles/WebApplicationFirewallEnabled/metric.rego
```diff
@@ -13,5 +13,6 @@ applicable {
 }
 
 compliant {
+	compare("isIn",  "LoadBalancer", input.type)
```
lebogg, line 16: Same here

lebogg (commented):
The two metrics are addressing the input differently (see line 9, respectively):

- `enabled := input.accessRestriction.enabled` vs
- `enabled := input.webApplicationFirewall.enabled`

One of them should therefore fail because the hierarchy is not right. I guess it is the second one and, thus, both should be `enabled := input.accessRestriction.enabled` or even `enabled := input.accessRestriction.Firewall.<L3Firewall/WebApplicationFirewall>.enabled`? Maybe I miss something. Did you test it in our testbed?

anatheka:
> The two metrics are addressing the input differently (see line 9, respectively):
> 
>     * `enabled := input.accessRestriction.enabled` vs
> 
>     * `enabled := input.webApplicationFirewall.enabled`
> 
> 
> One of them should therefore fail because the hierarchy is not right. I guess it is the second one and, thus, both should be `enabled := input.accessRestriction.enabled` or even `enabled := input.accessRestriction.Firewall.<L3Firewall/WebApplicationFirewall>.enabled`? Maybe I miss something. Did you test it in our testbed?

I've fixed it and in the pasta I tested it several times for the L3Firewall but not for the WebApplicationFirewall :woman_shrugging:

policies/bundles/WebApplicationFirewallEnabled/metric.rego
```diff
@@ -6,10 +6,11 @@ default applicable = false
 
 default compliant = false
 
-enabled := input.webApplicationFirewall.enabled
+enabled := input.accessRestriction.enabled
```
lebogg, line 9: Is this actually working this way? When I look at webprotoger, `accessRestriction` is parent of `firewall` which is parent of `webApplicationFirewall`, i.e. `webApplicationFirewall` inherits everything from its parents but not vice versa, right?

lebogg: Same for metric l3firewall

policies/bundles/L3FirewallEnabled/metric.rego
anatheka: Done

lebogg (approved):
LGTM now

PR #1196, opened 2023-09-11
Fix UpdateCertificate db query

anatheka:
Extend count call with id-part.


lebogg (approved):
Interesting that it isn't/wasn't apparent in the tests, right? The happy path should have failed then?

anatheka:
> Interesting that it isn't/wasn't apparent in the tests, right? The happy path should have failed then?

Yes, is it possible that this is due to the inmemory?

PR #1198, opened 2023-09-13
More DB Fixes

anatheka:
This PR fixes an issue with the cloud service selection in ListCertificates and adds a test with multiple cloud services in ListEvidences.


service/evidence/evidence_store.go
```diff
@@ -219,8 +223,12 @@ func (svc *Service) ListEvidences(ctx context.Context, req *evidence.ListEvidenc
 
 	// In any case, we need to make sure that we only select evidences of cloud services that we have access to
 	if !all {
+
 		query = append(query, "cloud_service_id IN ?")
-		conds = append(conds, allowed)
+
+		// Add all cloud_service_ids to the slice
+		ids = util.StringToAnySlice(allowed)
```
oxisto, line 230: are you sure about this? I thought we had several places where we are using it like before and I thought it was working fine?

It is also used like this in https://gorm.io/docs/query.html#String-Conditions, with just a regular array.

anatheka: I guess it's never really worked before. It's just not noticed because we never had a list of cloud_service_ids before. 
The problem is the following: Before we got a list that looked like this '(("11111111-1111-1111-1111-111111111111","22222222-2222-2222-2222-222222222222"))' , but we need a list that looks like that '(("11111111-1111-1111-1111-111111111111"),("22222222-2222-2222-2222-222222222222"))'. Maybe it is also possible to change that in the List() function, I don't know.

anatheka: I added a new test with 2 cloud_service_ids. You can try it out by changing the lines 228-231 back.

oxisto: I changed the lines back to
```
query = append(query, "cloud_service_id IN ?")
conds = append(conds, allowed)
```

and your test succeeds without a problem 🤷

service/orchestrator/certificates.go
```diff
@@ -85,12 +86,18 @@ func (svc *Service) ListCertificates(ctx context.Context, req *orchestrator.List
...
+	)
+
 	all, allowed := svc.authz.AllowedCloudServices(ctx)
 	if !all {
-		conds = append([]any{"cloud_service_id IN ?"}, []any{allowed})
+		query = append(query, "cloud_service_id IN ?")
+		conds = append(conds, allowed)
 	}
 
+	conds = append([]any{strings.Join(query, " AND ")}, conds...)
```
oxisto, line 100: If we use the same paradigm of having a query and a conds object, which then later gets combined, it works. We should probably think of having this pattern extracted into a function or similar

oxisto (approved):

service/orchestrator/certificates.go
lebogg: Here, the `" AND "` part never gets executed/joined because we only have a single query (`"cloud_service_id IN ?"`)?! As a general pattern, yes, I could understand it; but here it is just adding something that's not used. So what do we actually achieve here?

oxisto: We achieve that it works ;)

lebogg: aha

oxisto: I extracted it into a function

PR #1228, opened 2023-10-26
Update demo catalog with Codyze metrics

anatheka:


catalogs/demo_catalog.json
```diff
@@ -34,6 +34,16 @@
                             },
                             {
                                 "id": "CommSec-01.2",
+                                "description": "The CSP shall use a state-of-the-art TLS version.",
+                                "metrics": [
+                                    {
+                                        "id": "TLSVersion"
```
oxisto, line 40: There is no TLSVersion metric (anymore), We only have TransportEncryptionProtocol

anatheka: I know, but Codyze still uses TLSVersion as a metric, so I just added that. 
But regardless of that, it is still not well solved in Clouditor We have the name transport_encryption_protocol and check for the protocol including version number.

oxisto: Since MEDINA is now finished, we can probably fork the medina-codyze repo to GitHub and move to our recent set of metrics. And maybe then finally solve this TLS mess once and for all before EMERALD starts.

service/orchestrator/metrics.json
```diff
@@ -390,6 +446,45 @@
       }
     }
   },
+  {
+    "id": "TlsCipherSuites",
```
oxisto, line 450: We should probably rename them to `TransportEncryptionXXX` in the future to harmonize them with the others, but I guess for now its ok.

oxisto (approved):

lebogg (approved):

PR #1230, opened 2023-10-29
Initial attempt of a resource graph

oxisto:
In order to build a resource graph, several adjustments in other components were needed as well:

- Update of the ontology to get the `parent` attribute
- Modelling a `ResourceGroup` as part of a `Resource`
- Adding the optional `Parent` attribute to all discoverers; currently only filled by Azure
- Providing an example of filling the parent attribute in Azure by discovering resource groups and subscriptions
- Adding new API endpoints to the discoverer, they are currently marked as `EXPERIMENTAL` and are part of the `v1experimental` namespace, because their API is definitely not stable.


api/discovery/discovery.go
```diff
@@ -73,6 +101,36 @@
...
 	}
 }
+
+func (r *Resource) ToVocResource() (voc.IsCloudResource, error) {
+	var (
+		b   []byte
+		err error
+	)
+
+	typ := strings.Split(r.ResourceType, ",")[0]
```
lebogg, line 115: Thus, we have to keep in mind that 1st) we have to use the right order of types in r.ResourceType (e.g., VirtualMachine before Compute) and 2nd) they have to accurately match the name of the structs in voc

api/discovery/discovery.proto
```diff
@@ -84,7 +84,7 @@ message Resource {
   // Id contains a unique ID for each resource. This is specific for the cloud
   // provider this resource was gathered for and can for example be a resource
   // URL.
-  string id = 1 [ (validate.rules).string.uuid = true ];
+  string id = 1 [ (validate.rules).string.min_len = 1 ];
```
lebogg, line 87: Interesting, how did it work before ^^ Resource IDs in Azure (starting with "/subscriptions") definitely weren't UUIDs

api/discovery/experimental.proto
```diff
@@ -0,0 +1,86 @@
...
+  //
+  // Note: THIS API IS EXPERIMENTAL AND SUBJECT TO CHANGE
+  rpc ListGraphEdges(ListGraphEdgesRequest) returns (ListGraphEdgesResponse) {
+    option (google.api.http) = {
+      get : "/v1experimental/discovery/graph/edges"
+    };
+  }
+}
+
+message GraphEdge {
```
lebogg, line 66: Should be located at the end of the file

service/discovery/azure/azure.go
```diff
@@ -471,6 +477,14 @@ func resourceGroupName(id string) string {
 	return strings.Split(id, "/")[4]
 }
 
+func resourceGroupID(ID *string) voc.ResourceID {
+	// split according to "/"
+	s := strings.Split(util.Deref(ID), "/")
+	id := strings.Join(s[:5], "/")
```
lebogg, line 483: Is this separation guaranteed? Maybe there are resources w/o resource group?

lebogg: Maybe a test would clarify it

service/discovery/azure/resource_group.go
```diff
@@ -0,0 +1,127 @@
...
+			},
+			labels(rg.Tags),
+			voc.ResourceID(*d.sub.SubscriptionID),
+			[]string{"ResourceGroup", "Resource"},
+		),
+	}
+}
+
+// handleSubscription returns a [voc.Account] out of an existing [armsubscription.Subscription].
+func (d *azureResourceGroupDiscovery) handleSubscription(rg *armsubscription.Subscription) *voc.Account {
```
lebogg, line 110: `rg` parameter isn't needed and  used

lebogg: And something like 's' would be more fitting

service/discovery/discovery.go
```diff
@@ -457,3 +448,21 @@ func (svc *Service) ListResources(ctx context.Context, req *discovery.ListResour
 func (svc *Service) GetCloudServiceId() string {
 	return svc.csID
 }
+
+func toDiscoveryResource(resource voc.IsCloudResource) (r *discovery.Resource, v *structpb.Value, err error) {
```
lebogg, line 452: Add 1 test

go.mod
```diff
@@ -187,6 +187,8 @@ require (
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
+require github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources v1.0.0
```
lebogg, line 190: To azure block

api/discovery/discovery.go
oxisto: That should be guaranteed by owl2java. My idea would also to auto-generate this function at some point.

api/discovery/discovery.proto
oxisto: It probably worked because we never added "Resource" objects from the outside, only the ones inside an evidence and those are different and not validated.

lebogg: Right :D

lebogg (approved):
tip top (to me)

PR #1237, opened 2023-11-01
Added ontology objects for application dependencies

oxisto:
Currently they are hardcoded, we need to update owl2java at some point again. It will then displayed nicely in the UI as well. I added a section to the README with an example.

Screenshot 2023-11-01 at 10 25 38


voc/related.go
```diff
@@ -41,9 +41,10 @@ func (v VirtualMachine) Related() []string {
 		list = append(list, string(b))
 	}
 
-	if v.BootLogging != nil {
-		list = append(list, v.BootLogging.Related()...)
-	}
+	// Boot/os logging in Azure points to a URL, not a Resource ID, this is broken!
```
lebogg, line 44: I guess this change caused some tests to fail (see build error)

voc/library.go
```diff
@@ -0,0 +1,36 @@
...
+
+// Library is an entity in our Cloud ontology. It encapsulates the (source) code of a library, similar to an application.
+// TODO(oxisto): Add this to the ontology and auto-generate it
+type Library struct {
+	*Resource
+	Functionalities     []*Functionality `json:"functionalities"`
+	ProgrammingLanguage string           `json:"programmingLanguage"`
+	TranslationUnits    []ResourceID     `json:"translationUnits"`
+	Dependencies        []ResourceID     `json:"dependencies"`
+}
```
lebogg, line 36: I would prefer some small comments to clarify what they mean, e.g. for `Functionalities` and `TranslationUnits`. But, on the other hand, they would probably be overwritten by the auto generation anyway.

lebogg (commented):
Good but fix tests

voc/related.go
oxisto: Ah good point. Should I remove it from the Azure discoverer then? Basically the discoverer is doing it incorrectly.

voc/library.go
oxisto: Yes. I think we had a beta branch of the auto-generator that generates also comments, but development there has stalled for a while.

voc/library.go
```diff
@@ -0,0 +1,36 @@
...
+// \$$$$$$\  $$ |\$$$$$   |\$$$$$   |\$$$$$$  |$$ |  \$$$   |\$$$$$   |$$ |
+//  \_______|\__| \______/  \______/  \_______|\__|   \____/  \______/ \__|
+//
+// This file is part of Clouditor Community Edition.
+
+package voc
+
+// Library is an entity in our Cloud ontology. It encapsulates the (source) code of a library, similar to an application.
+// TODO(oxisto): Add this to the ontology and auto-generate it
+type Library struct {
```
lebogg, line 30: Do we want to have sth. like `LibraryType = [Library(, Application), Resource]` here as well?

voc/translation_unit_declaration.go
```diff
@@ -0,0 +1,37 @@
...
+// This file is part of Clouditor Community Edition.
+
+package voc
+
+var TranslationUnitDeclarationType = []string{"TranslationUnitDeclaration", "Resource"}
+
+// TranslationUnitDeclaration is an entity in our Cloud ontology. It refers to https://fraunhofer-aisec.github.io/cpg/CPG/specs/graph/#translationunitdeclaration
+type TranslationUnitDeclaration struct {
+	*Resource
+	Name string `json:"name"`
```
lebogg, line 35: We have `Name` in the anonymous filed `Resource` already - with the same json tag. Will this be a problem?

api/discovery/experimental.proto
```diff
@@ -45,7 +45,8 @@ service ExperimentalDiscovery {
   // the resource graph.
   //
   // Note: THIS API IS EXPERIMENTAL AND SUBJECT TO CHANGE
-  rpc UpdateResource(UpdateResourceRequest) returns (UpdateResourceResponse) {
+  rpc UpdateResource(UpdateResourceRequest)
+      returns (clouditor.discovery.v1.Resource) {
```
lebogg, line 49: Yes :+1:

voc/related.go
```diff
@@ -117,3 +117,19 @@ func (s Storage) Related() []string {
...
+	}
+
+	for _, tu := range a.TranslationUnits {
+		list = append(list, string(tu))
+	}
+
+	list = append(list, a.Resource.Related()...)
+
+	return list
+}
```
lebogg, line 135: Isn't Compute a related field as well?

voc/translation_unit_declaration.go
oxisto: That is true, its redundant

PR #1257, opened 2023-11-23
Extend discovery

lebogg:
This PR will extend the Azure discovery by adding:
- Key Vault resource
- Zone Redundancy security features
- Web App function
- Table Storage w/ Backup

Note: In the end, we will check what we will transfer to main and what not.


lebogg:
Only temporary work, parts of it will may be added in the future

PR #1258, opened 2023-11-23
Azure Discovery: Add WebApp

anatheka:
What does the property 'enabled' mean in HTTPEndpoint.TransportEncryption. Do we want to add here the HTTP vs HTTPS usage?


service/discovery/azure/compute.go
```diff
@@ -279,6 +277,51 @@
...
 
+func (d *azureComputeDiscovery) handleWebApp(webApp *armappservice.Site) voc.IsCompute {
+	var tlsVersion = ""
+
+	// If a mandatory field is empty, the whole function is empty
+	if webApp == nil {
+		return nil
+	}
+
+	if *webApp.Properties.SiteConfig.MinTLSVersion == armappservice.SupportedTLSVersionsOne2 {
```
lebogg, line 288: `util.Deref`

service/discovery/azure/compute.go
```diff
@@ -279,6 +277,51 @@ func (d *azureComputeDiscovery) handleFunction(function *armappservice.Site) voc
...
+		tlsVersion = constants.TLS1_2
+	} else if *webApp.Properties.SiteConfig.MinTLSVersion == armappservice.SupportedTLSVersionsOne1 {
```
lebogg, line 290: `util.Deref`

service/discovery/azure/compute.go
```diff
@@ -279,6 +277,51 @@ func (d *azureComputeDiscovery) handleFunction(function *armappservice.Site) voc
...
+		tlsVersion = constants.TLS1_1
+	} else if *webApp.Properties.SiteConfig.MinTLSVersion == armappservice.SupportedTLSVersionsOne0 {
```
lebogg, line 292: `util.Deref`

service/discovery/azure/compute.go
```diff
@@ -279,6 +277,51 @@ func (d *azureComputeDiscovery) handleFunction(function *armappservice.Site) voc
...
+				// config,
+			),
+			NetworkInterfaces: []voc.ResourceID{voc.ResourceID(*webApp.Properties.VirtualNetworkSubnetID)}, // Add the Virtual Network Subnet ID
+		},
+		HttpEndpoint: &voc.HttpEndpoint{
+			TransportEncryption: &voc.TransportEncryption{
+				Enforced:   util.Deref(webApp.Properties.HTTPSOnly),
+				TlsVersion: tlsVersion,
+				Algorithm:  constants.TLS,
+				// Enabled: ,
```
lebogg, line 319: I would say if local variable `tlsVersion` is not empty than it is enabled?

service/discovery/azure/compute.go
```diff
@@ -279,6 +277,51 @@ func (d *azureComputeDiscovery) handleFunction(function *armappservice.Site) voc
...
+				webApp,
+				// config,
+			),
+			NetworkInterfaces: []voc.ResourceID{voc.ResourceID(*webApp.Properties.VirtualNetworkSubnetID)}, // Add the Virtual Network Subnet ID
+		},
+		HttpEndpoint: &voc.HttpEndpoint{
+			TransportEncryption: &voc.TransportEncryption{
+				Enforced:   util.Deref(webApp.Properties.HTTPSOnly),
+				TlsVersion: tlsVersion,
+				Algorithm:  constants.TLS,
```
lebogg, line 318: I think, here we want the TLS Cipher (e.g. AES or ECDHE)

lebogg (changes requested):

service/discovery/azure/compute.go
```diff
@@ -279,6 +277,64 @@ func (d *azureComputeDiscovery) handleFunction(function *armappservice.Site) voc
...
+		tlsVersion = constants.TLS1_1
+	case armappservice.SupportedTLSVersionsOne0:
+		tlsVersion = constants.TLS1_0
+	}
+
+	if tlsVersion != "" {
+		enc = &voc.TransportEncryption{
+			Enforced:   util.Deref(webApp.Properties.HTTPSOnly),
+			TlsVersion: tlsVersion,
+			Algorithm:  AES256,
```
lebogg, line 304: Doesn't have to be AES256, see also:
See https://techcommunity.microsoft.com/t5/apps-on-azure-blog/min-tls-cipher-suite-preview-now-available-on-azure-portal-and/ba-p/3804134

I think what we actually want here is a list of possible cipher suites the resource with transport encryption (here web app) allows?

go.mod
```diff
@@ -194,4 +194,6 @@ require (
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
+require github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/appconfiguration/armappconfiguration v1.1.1 // indirect
```
oxisto, line 197: fix please :)

service/discovery/azure/compute.go
oxisto: It should be there now with release 2.3.0. See https://github.com/Azure/azure-sdk-for-go/blob/0d3efbd9bb2603eed4e91e71379d11ad43ead15a/sdk/resourcemanager/appservice/armappservice/CHANGELOG.md?plain=1#L18

I think we are using a very old version of this module (1.x vs. 2.x)

anatheka: I added the cipher suits to WebApps and functions.

lebogg (approved):
LGTM

PR #1259, opened 2023-11-23
Azure Discovery: Add CosmosDB

anatheka:


service/discovery/azure/storage.go
```diff
@@ -113,9 +114,107 @@
...
+		return nil, err
+	}
+
+	// Check if KeyVaultURI is set
+	// By default the Cosmos DB account is encrypted by Azure managed keys. Optionally, it is possible to add a second encryption layer with customer key encryption. (see https://learn.microsoft.com/en-us/azure/cosmos-db/how-to-setup-customer-managed-keys?tabs=azure-portal)
+	if account.Properties.KeyVaultKeyURI != nil {
+		enc = &voc.CustomerKeyEncryption{
+			AtRestEncryption: &voc.AtRestEncryption{
+				Enabled: true,
+				// Algorithm: algorithm, //TODO(anatheka): How do we get the algorithm? Are we available to do it by the related resources?
```
lebogg, line 183: Yes, it seems so. The other option would be to do an additional redundant call to keyvault/keys client (redundant because we do it in KeyVaults already). So this should be a perfect fit for related resources.

For now it is enough to see if it is enabled I think.

service/discovery/azure/storage.go
```diff
@@ -113,9 +114,107 @@ func (d *azureStorageDiscovery) List() (list []voc.IsCloudResource, err error) {
...
+			},
+			KeyUrl: *account.Properties.KeyVaultKeyURI,
```
lebogg, line 185: I think it is safer to use `util.Deref` here

service/discovery/azure/storage.go
```diff
@@ -113,9 +114,107 @@ func (d *azureStorageDiscovery) List() (list []voc.IsCloudResource, err error) {
...
+			},
+		}
+	}
+
+	// Create Cosmos DB database account voc object
+	dbStorage := &voc.DatabaseStorage{
+		Storage: &voc.Storage{
+			Resource: discovery.NewResource(d,
+				voc.ResourceID(*account.ID),
+				*account.Name,
```
lebogg, line 201: `util.Deref`

go.mod
```diff
@@ -193,4 +193,6 @@ require (
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
+require github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/cosmos/armcosmos v1.0.0 // indirect
```
lebogg, line 196: put into azure block

lebogg:
And don't forget to update the `types` in api/discovery/discovery.go:70

lebogg (approved):
LGTM

PR #1264, opened 2023-11-27
Add discovery interval option

anatheka:
Adds an option with which the discovery interval can be set. Otherwise, the discovery interval is set to 5 minutes.


oxisto (approved):

service/discovery/discovery.go
```diff
@@ -167,6 +163,13 @@ func WithStorage(storage persistence.Storage) ServiceOption {
 	}
 }
 
+// WithDiscoveryInterval is an option to set the discovery interval. If not set, the discovery is set to 5 minutes.
+func WithDiscoveryInterval(interval int) ServiceOption {
```
oxisto, line 167: Can we use `time.Duration` instead?

oxisto (approved):

PR #1265, opened 2023-11-27
Add RemoveMetric() and metric property deprecated

anatheka:
This PR introduces a deprecated field of type timestamp for metrics and adds a method RemoveMetric.

### DIFF
--- a/api/orchestrator/orchestrator.go
+++ b/api/orchestrator/orchestrator.go
@@ -30,6 +30,7 @@ import (
 	"slices"
 
 	"clouditor.io/clouditor/api/assessment"
+
 	"google.golang.org/protobuf/proto"
 )
 
--- a/service/orchestrator/metrics.go
+++ b/service/orchestrator/metrics.go
@@ -47,7 +47,7 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-// ErrCertificationNotFound indicates the certification was not found
+// ErrMetricNotFound indicates the certification was not found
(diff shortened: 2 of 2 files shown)


oxisto (commented):
Looks good: A more general point: Do we want to exclude the metric from the `ListMetrics` calls?

oxisto:
> > Looks good: A more general point: Do we want to exclude the metric from the `ListMetrics` calls?
> 
> That is a good question. Maybe we can use a filter option for all metrics including the deprecated and otherwise without the deprecated metrics?

sounds good

oxisto (approved):

api/orchestrator/orchestrator.go
```diff
@@ -29,6 +29,7 @@ import (
 	"context"
 	"slices"
 
+	"clouditor.io/clouditor/api/assessment"
 	"google.golang.org/protobuf/proto"
```
oxisto, line 33: nit: missing new line before external dependencies

service/orchestrator/metrics.go
```diff
@@ -43,9 +43,13 @@ import (
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/emptypb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+// ErrCertificationNotFound indicates the certification was not found
```
oxisto, line 50: copy/paste error

PR #1273, opened 2023-12-01
Update VM  bootLoggingOutput

anatheka:
**Blocked** before we finish this PR, we do more important tasks. 

Problem: 'vm.Properties.DiagnosticsProfile.BootDiagnostics' has only the StorageURI (e.g., "https://YYYY.blob.core.windows.net/") instead of the resource ID (e.g., '/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/XXXX/providers/Microsoft.Storage/storageAccounts/YYYY') as the storage location. It is not possible to put together the storage resource ID by using the VM resource group, because they can be in different resource groups.

Solution: We have to search the storage resource ID in our discovered resources (e.g, via a data query or an internal storage list of discovered resources) by searching for the storage name. Storage names are unique storage accounts provide unique namespaces.

This PR updates the VM property LoggingService.


anatheka:
As far as I can see, we are not able to get the resource ID for the VM bootdiagnostics output folder. Even if we get the resource ID for the storage account we do not know the used storage account container resource ID. Currently, we have the following possibilities: 
1. We do not fill the LoggingService for BootLogging 
2. We use the storageURI instead of the resource ID, but I don't know what the added value is at the moment
3. We try to get the correct resource ID by using the related resources (as soon as these are implemented)

lebogg:
Theoretically not blocked anymore

anatheka:
It is not being worked on at this time and is therefore being closed. An issue regarding this has been created #1771

PR #1274, opened 2023-12-01
Fix webApp, function and cosmosDB

anatheka:


service/discovery/azure/compute.go
```diff
@@ -158,12 +158,18 @@ func (d *azureComputeDiscovery) discoverFunctionsWebApps() ([]voc.IsCloudResourc
 		func(site *armappservice.Site) error {
 			var r voc.IsCompute
 
+			// Get configuration
+			config, err := d.clients.webAppClient.GetConfiguration(context.Background(), *site.Properties.ResourceGroup, *site.Name, &armappservice.WebAppsClientGetConfigurationOptions{})
+			if err != nil {
+				log.Errorf("error getting site config: %v", err)
```
lebogg, line 164: You can quickly press the resolve button: Is the following working appropriately if we get error and config is nil/empty... (probably done with tests)

service/discovery/azure/compute.go
```diff
@@ -207,33 +213,26 @@ func (d *azureComputeDiscovery) discoverFunctionsWebApps() ([]voc.IsCloudResourc
 	return list, nil
 }
 
-func (d *azureComputeDiscovery) handleFunction(function *armappservice.Site) voc.IsCompute {
+func (d *azureComputeDiscovery) handleFunction(site *armappservice.Site, config armappservice.WebAppsClientGetConfigurationResponse) voc.IsCompute {
```
lebogg, line 216: I personally liked more the `fucntion` naming but site is ok, ofc

service/discovery/azure/compute.go
```diff
@@ -207,33 +213,26 @@ func (d *azureComputeDiscovery) discoverFunctionsWebApps() ([]voc.IsCloudResourc
...
-		err                 error
-		publicNetworkAccess = false
+		ni              []voc.ResourceID
+		publicAccess    bool
+		resourceLogging *voc.ResourceLogging
+		runtimeLanguage string
+		runtimeVersion  string
 	)
 
-	// If a mandatory field is empty, the whole function is empty
```
lebogg, line 219: a comment here would not be so wrong

service/discovery/azure/compute.go
```diff
@@ -207,33 +213,26 @@ func (d *azureComputeDiscovery) discoverFunctionsWebApps() ([]voc.IsCloudResourc
...
-	if *function.Kind == "functionapp,linux" { // Linux function
-		runtimeLanguage, runtimeVersion = runtimeInfo(*function.Properties.SiteConfig.LinuxFxVersion)
-	} else if *function.Kind == "functionapp" { // Windows function, we need to get also the config information
-		// Get site config
-		config, err = d.clients.sitesClient.GetConfiguration(context.Background(), *function.Properties.ResourceGroup, *function.Name, &armappservice.WebAppsClientGetConfigurationOptions{})
-		if err != nil {
-			log.Errorf("error getting site config: %v", err)
-		}
+	// Get properties for the function
+	ni, publicAccess, resourceLogging = d.getSiteProperties(site)
```
lebogg, line 231: I like. Alternatively, one could do single getter functions and use them directly withing `return &voc.Fucntion` part

service/discovery/azure/compute.go
```diff
@@ -207,33 +213,26 @@ func (d *azureComputeDiscovery) discoverFunctionsWebApps() ([]voc.IsCloudResourc
...
 
+	if *site.Kind == "functionapp,linux" { // Linux function
```
lebogg, line 233: Suggestion: Maybe extract thi big if and else to a separate function to keep the code in this method cleaner. Personal preference though

service/discovery/azure/compute.go
```diff
@@ -262,119 +261,71 @@ func (d *azureComputeDiscovery) handleFunction(function *armappservice.Site) voc
 	return &voc.Function{
 		Compute: &voc.Compute{
 			Resource: discovery.NewResource(d,
-				voc.ResourceID(util.Deref(function.ID)),
-				util.Deref(function.Name),
+				voc.ResourceID(util.Deref(site.ID)),
```
lebogg, line 264: See above. Here, `function.ID`  would be more clear to me

service/discovery/azure/compute.go
```diff
@@ -262,119 +261,71 @@ func (d *azureComputeDiscovery) handleFunction(function *armappservice.Site) voc
...
-		PublicNetworkAccess: publicNetworkAccess,
+		RuntimeLanguage: runtimeLanguage,
+		RuntimeVersion:  runtimeVersion,
+		PublicAccess:    publicAccess,
+		Redundancy:      getRedundancy(site),
 	}
 }
 
-func (d *azureComputeDiscovery) handleWebApp(webApp *armappservice.Site) voc.IsCompute {
+func (d *azureComputeDiscovery) handleWebApp(site *armappservice.Site, config armappservice.WebAppsClientGetConfigurationResponse) voc.IsCompute {
```
lebogg, line 290: Points above about functions count here as well, I guess

service/discovery/azure/compute.go
```diff
@@ -262,119 +261,71 @@ func (d *azureComputeDiscovery) handleFunction(function *armappservice.Site) voc
...
 	return &voc.WebApp{
 		Compute: &voc.Compute{
 			Resource: discovery.NewResource(d,
-				voc.ResourceID(util.Deref(webApp.ID)),
-				util.Deref(webApp.Name),
-				// No creation time available
-				nil, // Only the last modified time is available
+				voc.ResourceID(util.Deref(site.ID)),
+				util.Deref(site.Name),
+				nil, // TODO(all): Only the last modified time is available, do we want that information?
```
lebogg, line 310: No, would be misleading I would say

service/discovery/azure/compute.go
```diff
@@ -724,9 +618,179 @@ func (d *azureComputeDiscovery) keyURL(diskEncryptionSetID string) (string, *arm
...
+	ni = []voc.ResourceID{}
+	resourceLogging = &voc.ResourceLogging{}
+
+	// If a mandatory field is empty, the whole function is empty
+	if site == nil {
+		return
+	}
+
+	// Get virtual network subnet ID
+	if site.Properties.VirtualNetworkSubnetID != nil {
```
lebogg, line 633: Properties shouldn't be ever nil. Otherwise, it could panic here. But I guess, Properties is a mandatory field of site

lebogg: Couldn't we use/build a util.deref function that takes into account the different levels and check if they are nil? I guess not or it is very hard

service/discovery/azure/compute.go
```diff
@@ -724,9 +618,179 @@ func (d *azureComputeDiscovery) keyURL(diskEncryptionSetID string) (string, *arm
...
+		ni = []voc.ResourceID{voc.ResourceID(util.Deref(site.Properties.VirtualNetworkSubnetID))}
+	}
+
+	// Check if resource is public available
+	if util.Deref(site.Properties.PublicNetworkAccess) == "Enabled" {
```
lebogg, line 638: I am not sure if it is the case here as well. But mostly, the Azure SDK defines Types and constants with underlying type string that holds this (magic) string.

service/discovery/azure/compute.go
```diff
@@ -724,9 +618,179 @@ func (d *azureComputeDiscovery) keyURL(diskEncryptionSetID string) (string, *arm
...
+	case armappservice.SupportedTLSVersionsOne0:
+		tlsVersion = constants.TLS1_0
+	}
+
+	// Create transportEncryption voc object
+	if tlsVersion != "" {
+		enc = &voc.TransportEncryption{
+			Enforced:   util.Deref(siteProperties.HTTPSOnly),
+			TlsVersion: tlsVersion,
+			Algorithm:  string(util.Deref(config.Properties.MinTLSCipherSuite)),
```
lebogg, line 708: Maybe add comment that it just got added to the SDK recently and field is probably not filled. To avoid confusion of future contributor/maintainer

service/discovery/azure/compute.go
```diff
@@ -724,9 +618,179 @@ func (d *azureComputeDiscovery) keyURL(diskEncryptionSetID string) (string, *arm
...
+			Algorithm: string(util.Deref(config.Properties.MinTLSCipherSuite)),
+		}
+	}
+
+	return
+}
+
+// runtimeInfo returns the runtime language and version
+func runtimeInfo(runtime string) (runtimeLanguage string, runtimeVersion string) {
+	if runtime == "" || !strings.Contains(runtime, "|") {
```
lebogg, line 724: Maybe add a lil bit of explanation why "|" is so important here.

service/discovery/azure/compute.go
```diff
@@ -724,9 +618,179 @@ func (d *azureComputeDiscovery) keyURL(diskEncryptionSetID string) (string, *arm
...
+		return "", ""
+	}
+	split := strings.Split(runtime, "|")
+	runtimeLanguage = split[0]
+	runtimeVersion = split[1]
+
+	return
+}
+
+// automaticUpdates returns automaticUpdatesEnabled and automaticUpdatesInterval for a given VM.
```
lebogg, line 734: I guess nothing changed here. Just re-ordering

voc/database_storage.go
```diff
@@ -32,4 +32,5 @@ var DatabaseStorageType = []string{"DatabaseStorage", "Storage", "Resource"}
 // DatabaseStorage is an entity in our Cloud ontology. describes the actual database or a table in a database
 type DatabaseStorage struct {
 	*Storage
+	PublicAccess bool `json:"publicAccess"`
```
lebogg, line 35: For me it is ok. But we should discuss it internally how we want to do that. E.g., maybe we should an http endpoint here as well.

voc/function.go
```diff
@@ -31,8 +31,9 @@ var FunctionType = []string{"Function", "Compute", "Resource"}
 
 type Function struct {
 	*Compute
-	HttpEndpoint        *HttpEndpoint `json:"httpEndpoint"`
-	RuntimeVersion      string        `json:"runtimeVersion"`
-	RuntimeLanguage     string        `json:"runtimeLanguage"`
-	PublicNetworkAccess bool          `json:"publicNetworkAccess"`
+	HttpEndpoint    *HttpEndpoint `json:"httpEndpoint"`
```
lebogg, line 34: Speaking of http endpoint: It should include a public access field, shouldn't it?

voc/function.go
```diff
@@ -31,8 +31,9 @@ var FunctionType = []string{"Function", "Compute", "Resource"}
...
+	RuntimeVersion  string        `json:"runtimeVersion"`
+	RuntimeLanguage string        `json:"runtimeLanguage"`
+	PublicAccess    bool          `json:"publicNetworkAccess"`
```
lebogg, line 37:
```suggestion
	PublicAccess    bool          `json:"publicAccess"`
```

voc/web_app.go
```diff
@@ -31,6 +31,7 @@ var WebAppType = []string{"WebApp", "Compute", "Resource"}
 
 type WebApp struct {
 	*Compute
-	HttpEndpoint        *HttpEndpoint `json:"httpEndpoint"`
-	PublicNetworkAccess bool          `json:"publicNetworkAccess"`
+	HttpEndpoint *HttpEndpoint `json:"httpEndpoint"`
+	PublicAccess bool          `json:"publicNetworkAccess"`
```
lebogg, line 35:
```suggestion
	PublicAccess bool          `json:"publicAccess"`
```

lebogg (commented):
Only small things

service/discovery/azure/compute.go
anatheka: Ups..  I've cleaned up too much :)

service/discovery/azure/compute.go
anatheka: I don't know exactly how to handle this either. And I also thought about a extended Deref-function. Maybe we could convert your comment to an issue.

service/discovery/azure/compute.go
anatheka: That is right, but in that case I couldn't find the corresponding constant.

service/discovery/azure/compute.go
anatheka: I don't think we need to explain that, it's a normal string from Azure and they use the pipe and if we don't check that, we get a string split error just like any other string splits.

voc/function.go
anatheka: I think yes, but I think that is something for our developer meeting.

service/discovery/azure/compute.go
anatheka: Yes, I think it works properly. In handleFunction and handleWebApp we check if the config is empty.

service/discovery/azure/storage.go
```diff
@@ -256,69 +325,98 @@ func (d *azureStorageDiscovery) discoverSqlServers() ([]voc.IsCloudResource, err
...
-			err = fmt.Errorf("%s: %v", ErrGettingNextPage, err)
-			return nil, err
+			log.Errorf("%s: %v", ErrGettingNextPage, err)
+			return list, anomalyDetectionList
 		}
 
 		for _, value := range pageResponse.Value {
-			// Getting anomaly detection status
-			anomalyDetectionEnabeld, err := d.anomalyDetectionEnabled(server, value)
+			// Create anomaly detection property
```
anatheka, line 404: @oxisto @lebogg regarding the TODO of oxisto do we want the anomalyDetection as part of the databaseService or databaseStorage? I have done it as a list of anomalyDetections in the databaseService, but I'm not sure if that is a good idea. What do you think?
 
```
// TODO(oxisto): This is not 100 % accurate. According to our ontology definition, the SQL server would be
// the database service and individual databases would be DatabaseStorage objects. However, the problem is
// that azure defines anomaly detection on a per-database level and we currently have anomaly detection as
// part of the service
```

oxisto: Sounds interesting, the only problem that I see is that we cannot differentiate between the different anomaly detection objects, can we?

anatheka: I've added an additional field 'resource' for the database resource ID, that could be a possibility. 
grafik

oxisto: We changed it to `Scope` to avoid confusion.

voc/database_storage.go
anatheka: That is changed back. See the comment of @oxisto.

PR #1280, opened 2023-12-06
Switch to Azure client factory

anatheka:


service/discovery/azure/compute.go
```diff
@@ -140,15 +130,15 @@ func (d *azureComputeDiscovery) List() (list []voc.IsCloudResource, err error) {
...
+	// initialize appservice factory client
+	if err := d.initClientAppserviceFactory(); err != nil {
 		return nil, err
 	}
 
 	// List functions
 	err := listPager(d.azureDiscovery,
-		d.clients.sitesClient.NewListPager,
-		d.clients.sitesClient.NewListByResourceGroupPager,
+		d.clients.appserviceFactory.NewWebAppsClient().NewListPager,
```
oxisto, line 140: This is creating multiple clients here in the same function and re-creates them every 5 minutes when the function is called. Do we know if this is a "good" idea, or the intentional way to do that? Are there any "real world" examples using the factory clients?

oxisto: At least in the examples they seem to re-use the clients between functions, so maybe we should only use the client factory to creat the clients but still store a reference in our clients object?

anatheka:
We are closing this PR, as the client factory currently offers no advantage.

PR #1281, opened 2023-12-06
Fix scheduler interval and tests

anatheka:


service/discovery/discovery.go
```diff
@@ -296,15 +296,16 @@ func (svc *Service) Start(ctx context.Context, req *discovery.StartDiscoveryRequ
 	}
 
 	for _, v := range discoverer {
-		log.Infof("Scheduling {%s} to execute every {%d} minutes...", v.Name(), svc.discoveryInterval)
+		log.Infof("Scheduling {%s} to execute every {%f} minutes...", v.Name(), svc.discoveryInterval.Minutes())
```
oxisto, line 299: I think you can just use `%s` directly with the interval, this should pretty print it in something like `5m` or `10h5m`, depending on the interval

service/discovery/discovery_test.go
```diff
@@ -43,8 +43,10 @@ import (
 	"clouditor.io/clouditor/internal/testutil/clitest"
 	"clouditor.io/clouditor/internal/testutil/servicetest"
 	"clouditor.io/clouditor/internal/util"
+	"clouditor.io/clouditor/persistence"
 	"clouditor.io/clouditor/service"
 	"clouditor.io/clouditor/voc"
+	"github.com/go-co-op/gocron"
```
oxisto, line 49: nit: missing newline between clouditor imports and rest

oxisto (approved):
some minor things, otherwise looks good.

PR #1282, opened 2023-12-07
Add tests to Azure resource group discovery

anatheka:
This PR adds a test file for the Azure resource group discovery.
@oxisto please check the failing tests. Are you sure, that you want only the subscription ID as parent ID and not the resource ID?


oxisto:
> This PR adds a test file for the Azure resource group discovery. @oxisto please check the failing tests. Are you sure, that you want only the subscription ID as parent ID and not the resource ID?

It should map to the "ID" field of the subscription, which is created as an `Account` object. I think an Azure subscription does not have a proper "resource ID" so in the end it is up to us whether we want to include the "/subscription" prefix or not, but it needs to be consistent between the parent ID of a a resource group and the "ID" of a subscription/account.

oxisto:
> The subscription has a proper resource ID and additionally the subscription ID. The question is, if the parent should contain the resource ID ("/subscriptions/00000000-0000-0000-0000-000000000000") or the subscription ID ("00000000-0000-0000-0000-000000000000"). For all other resources we use the resources ID and I think it is more consistent if we also use the resource id for the sub/account. But I don't know if that causes problems with the graph or something.

Then let's use the resource ID for it as well. Then we need to adjust also the line above.

oxisto (approved):

PR #1306, opened 2024-01-16
Re-organizing Azure discovery file structure

oxisto:
In the past we had major problems because of the complexity and the amount of code in the Azure discoverer in a single file. This proposes a new scheme for each discoverer subgroup that follows the following schema:
* `_discover.go` contains the `discoverXXX` functions, which list all resources in the azure native API
* `_handle.go` contains the `handleXXX` functions, which translate the azure API objects into ontology objects
* `_properties.go` contains all utility functions that translate specific, more complex properties, into properties used in the ontology
* `_test.go` contains all the tests. We could possibily also split this up into the three groups above

Common functions, especially the properties should go into the `azure_properties.go` to avoid duplication.


lebogg (approved):
LGTM

service/discovery/azure/storage_properties.go
```diff
@@ -0,0 +1,108 @@
...
+
+	"clouditor.io/clouditor/internal/constants"
+	"clouditor.io/clouditor/internal/util"
+	"clouditor.io/clouditor/voc"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/sql/armsql"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/storage/armstorage"
+)
+
+// checkTlsVersion returns Clouditor's TLS version constants for the given TLS version
+func checkTlsVersion(version string) string {
```
anatheka, line 41: We have to check the tls version in several places. We do it also in getTransportEncryption in compute_properties.go. We should consolidate this. But perhaps in a separate PR.

anatheka (approved):
Great job.

PR #1309, opened 2024-01-18
Consolidate Azure TLS property check

anatheka:


oxisto (approved):
We need to modify this according to our ontology changes, but for now this is good.

PR #1320, opened 2024-02-08
Update discovery to new ontology proto

anatheka:
This PR updates the discovery to the new ontology
- [x] K8S discovery
- [x] K8S tests
- [x] Azure discovery
- [x] Azure tests
- [x] AWS discovery
- [x] AWS tests
- [x] Re-enable test `TestService_Start` since it needs Azure and K8S discovery

This is breaking compatibility with the existing storage layer, since we adapted many changes in the ontology and finally also harmonized the way ontology structs are persisted. Previously, we used a `google.protobuf.Struct` without any formal definition of its content. Now, a `google.protobuf.Any` object is used, which can be used to wrap any valid protobuf message. Since all ontology objects are now protobuf messages, we can embed the ontology resources this way. We do an additional check, if the embedded proto message is really a message belonging to our ontology using a new interface called `ontology.IsResource`. This also removes any weird type registries we had previously when we tried to load a  `google.protobuf.Struct` back into a (now non-existing) `voc.IsCloudResource` for the discovery API. Instead, we can just use the existing unmarshal functionality of `Any`.

Remaining ontology issues:

- [x] A lot of fields that are supposed to be a duration are just an int64. We could maybe use google.protobuf.Duration instead?
- [x] Similar, some dates are just int64, maybe a google.protobuf.Timestamp would be better
- [x] The `LoggingService` of `Logging` was an array before, now its just a single ID
- [x] There are still problems with nested security features, we need to decide how they are represented in the JSON, so Rego works
- [x] Remove cloud service ID from ontology? Its already there in the evidence and the database wrapper struct
- [x] Why is parent named `parent_resource_id`?
- [x] Was automaticUpdates renamed to automaticSecurityUpdates?
- [x] Anomaly detection of database service should be list
- [x] Oslogging naming broken
- [x] Should we make the IDs a `*string` (`optional string`) in proto? Since they almost all seem optional. This would work better with the discoverers
- [x] Some properties have a duration but are named `DaysSince...`. We should rename them to something like `DurationSince...`
- [ ] `Function` and `WebApp` are missing HttpEndpoint, PublicAccess, Redundancy. Once this is done, we need to re-enable this in the Azure discoverer
- [x] Add an `s` to repeated properties, e.g., `Backups` or `HttpEndpoints` in `LoadBalancer`.


anatheka:
>  A lot of fields that are supposed to be a duration are just an int64. We could maybe use google.protobuf.Duration instead?
I thought that is not possible, see https://github.com/clouditor/clouditor/blob/f31ad2174d6ec5d6994e43d3befcc9dba4c516ec/api/assessment/metric.proto#L66

oxisto:
> > A lot of fields that are supposed to be a duration are just an int64. We could maybe use google.protobuf.Duration instead?
> 
> I thought that is not possible, see
> 
> https://github.com/clouditor/clouditor/blob/f31ad2174d6ec5d6994e43d3befcc9dba4c516ec/api/assessment/metric.proto#L66

Good point, however the relevant issue was closed (https://github.com/google/gnostic/issues/351) so we might need to revisit it

anatheka:
> The LoggingService of Logging was an array before, now its just a single ID

The LoggingService has the property "prop:to" and I thought it should just contain the id to the loggingService. Before it was an array of ResourceIDs. We have the property 4 times: 
- Resource HttpRequest: to HttpEnpoint
- Resource Logging: to LoggingService
- Resource DatabaseOperation: to  DatabaseService:
- Resource ObjectStorageRequest: to ObjectStorage

anatheka:
> Why is parent named parent_resource_id?

This is an error. I wanted to name it like "parent_<resource_name>_id". But I will change it to "parent_id".

oxisto:
> > Why is parent named parent_resource_id?
> 
> I named it in the schema "parent_<resource_name>_id". The BaseResource has "parent Resource" and that results in "string parent_resource_id". For example the resource DatabaseStorage has two parent properties. The one from BaseResource and "parent Storage". So it results to
> 
> * "string parent_resource_id" and
> * "string parent_storage_id".
> 
> Otherwise we have a conflict. But maybe the "parent Storage" in DatabaseStorage isn't correct, so we can delete it and use only "string parent_id" for the BaseResource.

The one from parent Storage is obsolete and should be deleted -> we should only have parent_id

service/assessment/assessment.go
```diff
@@ -289,11 +282,33 @@ func (svc *Service) AssessEvidences(stream assessment.Assessment_AssessEvidences
...
+		types    []string
+		m        proto.Message
+		resource ontology.IsResource
+	)
+
+	// First, try to extract the resource out of the evidence and validate it
+	m, err = ev.Resource.UnmarshalNew()
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "could not unmarshal resource proto message: %v", err)
+	}
```
anatheka, line 299: Should we move that to api.Validate()?

oxisto: It is a special case only for the wrapped resource. Adding it to the generic validate might be confusing.

oxisto: Additionally, handling it here has the advantage that we only need to unwrap the resource message once here. It is additionally unwrap in the Rego code, but I am trying to get rid of that as well.

internal/testutil/prototest/proto.go
```diff
@@ -0,0 +1,41 @@
...
+
+// NewAny creates a new [*anypb.Any] from a [proto.Message] with an assert that no error has been thrown.
+func NewAny(t *testing.T, m proto.Message) *anypb.Any {
+	a, err := anypb.New(m)
+	assert.NoError(t, err)
+
+	return a
+}
+
+// NewAny creates a new [*anypb.Any] from a [proto.Message] with a panic that no error has been thrown.
```
anatheka, line 21: "with a panic that no error has been thrown." ? Copy/paste error?

policies/rego.go
```diff
@@ -88,21 +89,24 @@ func NewRegoEval(opts ...RegoEvalOption) PolicyEval {
...
 	baseDir = "."
-	m = evidence.Resource.GetStructValue().AsMap()
 
-	types, err = evidence.ResourceTypes()
+	m, err = ontology.ResourceMap(r)
 	if err != nil {
-		return nil, fmt.Errorf("could not extract resource types from evidence: %w", err)
+		return nil, err
 	}
+	fmt.Println(m)
```
anatheka, line 107: Do we want that output?

policies/rego_test.go
```diff
@@ -444,3 +419,13 @@ func Test_regoEval_evalMap(t *testing.T) {
 		})
 	}
 }
+
+// TODO: Needed until IsCloudResource embeds proto.Message
```
anatheka, line 423: I think this can be deleted now?

anatheka (commented):
Looks very good. Only some minor comments.

service/discovery/azure/storage_discover.go
```diff
@@ -85,16 +85,15 @@ func (d *azureDiscovery) discoverCosmosDB() ([]voc.IsCloudResource, error) {
...
-		list []voc.IsCloudResource
+		list []ontology.IsResource
 		err  error
 	)
 
 	// initialize Mongo DB resources client
 	if err = d.initMongoDResourcesBClient(); err != nil {
-		log.Errorf("error initializing Mongo DB resource client: %v", err)
-		return list
+		return nil
```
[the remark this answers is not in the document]
anatheka: I found the answer :)

policies/bundles/CustomerKeyEncryption/metric.rego
```diff
@@ -4,13 +4,13 @@ default applicable = false
 
 default compliant = false
 
-import input.atRestEncryption as enc
+import input.atRestEncryption.customerKeyEncryption as cke
```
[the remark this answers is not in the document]
oxisto: Maybe in the future

policies/rego.go
oxisto: No :D

service/discovery/azure/compute_test.go
```diff
@@ -464,169 +71,137 @@ func Test_azureComputeDiscovery_discoverFunctionsWebApps(t *testing.T) {
...
+					Raw:                 "{\"*armappservice.Site\":[{\"id\":\"/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/res1/providers/Microsoft.Web/sites/WebApp1\",\"kind\":\"app\",\"location\":\"West Europe\",\"name\":\"WebApp1\",\"properties\":{\"httpsOnly\":true,\"publicNetworkAccess\":\"Enabled\",\"resourceGroup\":\"res1\",\"siteConfig\":{},\"virtualNetworkSubnetId\":\"/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/res1/providers/Microsoft.Network/virtualNetworks/vnet1/subnets/subnet1\"},\"tags\":{\"testKey1\":\"testTag1\",\"testKey2\":\"testTag2\"}}],\"armappservice.WebAppsClientGetConfigurationResponse\":[{\"name\":\"WebApp1\",\"properties\":{\"minTlsCipherSuite\":\"TLS_AES_128_GCM_SHA256\",\"minTlsVersion\":\"1.1\"},\"type\":\"Microsoft.Web/sites/config\"}]}",
+					NetworkInterfaceIds: []string{"/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/res1/providers/Microsoft.Network/virtualNetworks/vnet1/subnets/subnet1"},
+					ResourceLogging: &ontology.ResourceLogging{
+						Enabled: true,
+					},
+					/*HttpEndpoint: &ontology.HttpEndpoint{
```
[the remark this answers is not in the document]
oxisto: Yes I would keep it commented out to remind us that we want to have this in the future.

oxisto (approved):

PR #1327, opened 2024-02-14
Workaround to fill the ontology/openapi.yaml

anatheka:


anatheka:
Too many conflicts here, we do it again in a new PR.

PR #1330, opened 2024-02-16
Added quickstart with UI

oxisto:
This PR adds a quickstart option to start the engine with the UI.

This also adds the UI as a submodule..


README.md
```diff
@@ -27,6 +27,18 @@ Key features are:
 - descriptive development of custom rules using [Cloud Compliance Language (CCL)](clouditor-engine-azure/src/main/resources/rules/azure/compute/vm-data-encryption.md) to support individual evaluation scenarios
 - integration of custom security requirements and mapping to rules
 
+## QuickStart with UI
+
+In order to just build and run the Clouditor , without generating the protobuf file, one can use the following script.
```
anatheka, line 32:
```suggestion
In order to just build and run the Clouditor, without generating the protobuf file, one can use the following script.
```

README.md
```diff
@@ -27,6 +27,18 @@ Key features are:
...
+This still requires Go and nodejs to be installed.
```
anatheka, line 33:
```suggestion
This still requires Go and Node.js to be installed.
```

run-engine-with-ui.sh
```diff
@@ -0,0 +1,22 @@
...
+# Make sure the UI submodules are up-to-date
+git submodule update --remote
+
+# Build engine. This requires Go to be installed
+go build ./cmd/engine/engine.go
+
+# Install cl binary. 
+go install ./cmd/cli/cl.go
+
+# Build and run UI (in background). This requires nodejs to be installed
```
anatheka, line 12:
```suggestion
# Build and run UI (in background). This requires Node.js to be installed
```

service/discovery/discovery.go
```diff
@@ -277,15 +277,15 @@ func (svc *Service) Start(ctx context.Context, req *discovery.StartDiscoveryRequ
...
+				continue
 			}
 			discoverer = append(discoverer,
 				aws.NewAwsStorageDiscovery(awsClient, svc.csID),
 				aws.NewAwsComputeDiscovery(awsClient, svc.csID))
 		default:
 			newError := fmt.Errorf("provider %s not known", provider)
 			log.Error(newError)
-			return nil, status.Errorf(codes.InvalidArgument, "%s", newError)
+			continue
```
anatheka, line 288: Really a continue? Shouldn't the program exit here?

oxisto: The problem is that the discovery will fail even if only one of the discoverers fails. Most prominently AWS fails if no .aws config file is present. But I want to enable "all" discoverers in the Quickstart, so that whichever work, work.

oxisto: Reverted back to old state

PR #1333, opened 2024-02-16
Harmonize Azure resource ID to all lowercase

oxisto:
Because Azure is annoying.

To avoid issue https://github.com/MicrosoftDocs/azure-docs/issues/36222 we are harmonising all resource (group) IDs to be lowercase. Microsoft says that they compare case-insensitive.

### DIFF
--- a/clouditor.code-workspace
+++ b/clouditor.code-workspace
@@ -38,6 +38,7 @@
       "clitest",
       "Clouditor",
       "conds",
+      "dataprotection",
       "Debugf",
       "dialector",
       "discoverytest",
@@ -82,6 +83,7 @@
       "testutil",
       "timestamppb",
       "unmarshalling",
+      "virtualnetworks",
       "Warnf"
     ]
   }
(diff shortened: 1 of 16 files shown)


oxisto:
Also related to https://github.com/Azure/azure-rest-api-specs/issues/10357

Rasmus-breidahl:
@oxisto great change but

Found an example where i did not lower case the resourceGroups to resourcegroup

BTW the ID itself also have the same issue with UpperCase in some cases and cause issue all over
Name in Azure for VM =  Azure-test-ftp
TF finds name but = AZURE-TEST-FTP

azurerm_backup_protected_vm

source_vm_id        = "/subscriptions/xx/resourceGroups/xx/providers/Microsoft.Compute/virtualMachines/Azure-test-ftp" -> "/subscriptions/xx/resourceGroups/xx/providers/Microsoft.Compute/virtualMachines/AZURE-TEST-FTP" # forces replacement

PR #1335, opened 2024-02-16
Add Azure VM Scale Set

anatheka:
This PR adds
- Azure VM Scale Set discovery with VMs


service/discovery/azure/compute.go
```diff
@@ -399,6 +479,7 @@ func runtimeInfo(runtime string) (runtimeLanguage string, runtimeVersion string)
 }
 
 // Discover virtual machines
+// TODO(all): Do we want to discover the VMs of a scale set, too?
```
anatheka, line 482:
```suggestion
```

service/discovery/azure/azure.go
```diff
@@ -165,9 +165,11 @@ type clients struct {
 	sitesClient *armappservice.WebAppsClient
 
 	// Compute
-	virtualMachinesClient *armcompute.VirtualMachinesClient
-	blockStorageClient    *armcompute.DisksClient
-	diskEncSetClient      *armcompute.DiskEncryptionSetsClient
+	virtualMachinesClient           *armcompute.VirtualMachinesClient
+	virtualMachineScaleSetClient    *armcompute.VirtualMachineScaleSetsClient
```
lebogg, line 169: just minor thing to stay consistent: Use plural in the last word before Client

service/discovery/azure/compute.go
```diff
@@ -136,6 +144,78 @@ func (d *azureComputeDiscovery) List() (list []voc.IsCloudResource, err error) {
...
+		d.clients.virtualMachineScaleSetClient.NewListPager,
+		func(res armcompute.VirtualMachineScaleSetsClientListAllResponse) []*armcompute.VirtualMachineScaleSet {
+			return res.Value
+		},
+		func(res armcompute.VirtualMachineScaleSetsClientListResponse) []*armcompute.VirtualMachineScaleSet {
+			return res.Value
+		},
+
+		func(scaleSet *armcompute.VirtualMachineScaleSet) error {
+			r, err := d.discoverVMScaleSetVMs(scaleSet)
```
lebogg, line 168: Shouldn't that be `handleXXX` to be consistent?

lebogg: Ok, it seems to be a nested discover-then-handle-function

service/discovery/azure/compute.go
```diff
@@ -436,6 +516,65 @@ func (d *azureComputeDiscovery) discoverVirtualMachines() ([]voc.IsCloudResource
...
+					Region: *vm.Location,
+				},
+				labels(vm.Tags),
+				voc.ResourceID(parent),
+				voc.VirtualMachineType,
+				vm,
+			),
+		},
+
+		AutomaticUpdates: automaticUpdatesScaleSet(vm.Properties.OSProfile),
```
lebogg, line 536: Testing with the testbed on Azure, I get a nil pointer dereference here because `vm.Properties` is nil

lebogg (changes requested):
One nil pointer deference occured

service/discovery/azure/compute.go
anatheka: No, because first the VMs in the scale set must be discovered and then the handleX method is called.

PR #1336, opened 2024-02-16
Key Vault: Add Secrets

lebogg:


lebogg (commented):
should be ok now

voc/certificate.go
```diff
@@ -0,0 +1,40 @@
...
+//  \_______|\__| \______/  \______/  \_______|\__|   \____/  \______/ \__|
+//
+// This file is part of Clouditor Community Edition.
+
+package voc
+
+var CertificateType = []string{"Certificate"}
+
+type Certificate struct {
+	*Resource
```
anatheka, line 31: I think you do not need the Resource here.

voc/secret.go
```diff
@@ -0,0 +1,41 @@
...
+//  \_______|\__| \______/  \______/  \_______|\__|   \____/  \______/ \__|
+//
+// This file is part of Clouditor Community Edition.
+
+package voc
+
+var SecretType = []string{"Secret"}
+
+type Secret struct {
+	*Resource
```
anatheka, line 31: I think you do not need the Resource here.

service/discovery/azure/key_vault.go
```diff
@@ -120,25 +127,51 @@ func (d *azureKeyVaultDiscovery) discoverKeyVaults() (list []voc.IsCloudResource
...
 				list = append(list, k)
 			}
+			log.Infof("Adding secrets '%s'", secretIDs)
+			for _, s := range secrets {
+				list = append(list, s)
+			}
+			log.Infof("Adding certificates '%s'", secretIDs)
+			for _, c := range certificates {
+				list = append(list, c)
+			}
```
anatheka, line 174: I think in line 179 an error message would be good.

service/discovery/azure/key_vault.go
```diff
@@ -158,6 +191,15 @@ func (d *azureKeyVaultDiscovery) initKeysClient() (err error) {
...
 }
+func (d *azureKeyVaultDiscovery) initSecretsClient() (err error) {
+	d.clients.secretsClient, err = initClient(d.clients.secretsClient, d.azureDiscovery, armkeyvault.NewSecretsClient)
+	return
+}
+
+func (d *azureKeyVaultDiscovery) initCertificatesClient(kv *armkeyvault.Vault) (err error) {
+	d.clients.certificationsClient, err = azcertificates.NewClient(util.Deref(kv.Properties.VaultURI), d.cred, nil)
+	return
+}
```
anatheka, line 202: Just an idea for line 216: Shouldn't we discover also the keyVaults that are not used at the moment comparable to the virtual machines. We discover all VMs whether they are stopped or not. We could use an property "enabled" in the keyVault object.

voc/certificate.go
lebogg: We do need it. Certificates/Keys/Secrets are all resources themselves and attached to the corresponding key vault (=parent). As such they have IDs, creation Dates, types, tags. Metrics are/will be evaluated on these. 

Another option is to store all of them directly into the key vault. This was my first idea, but this way here is more clean.

voc/secret.go
lebogg: We do need it. Certificates/Keys/Secrets are all resources themselves and attached to the corresponding key vault (=parent). As such they have IDs, creation Dates, types, tags. Metrics are/will be evaluated on these. 

Another option is to store all of them directly into the key vault. This was my first idea, but this way here is more clean.

service/discovery/azure/key_vault.go
```diff
@@ -280,7 +337,7 @@ func (d *azureKeyVaultDiscovery) getKeys(kv *armkeyvault.Vault) ([]*voc.Key, err
 		}
 		for _, k := range page.Value {
 			// We have to request each single key because this lazy NewListPager doesn't fill out all key information
-			res, err := d.clients.keysClient.Get(context.Background(), util.Deref(d.rg), util.Deref(kv.Name),
+			res, err := c.Get(context.Background(), util.Deref(d.rg), util.Deref(kv.Name),
```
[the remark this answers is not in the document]
lebogg: It is just a check here because it should be initialized elsewhere/one level up I think (like in `discoverKeyVaults`)

service/discovery/azure/key_vault.go
```diff
@@ -309,6 +366,115 @@ func (d *azureKeyVaultDiscovery) getKeys(kv *armkeyvault.Vault) ([]*voc.Key, err
 	return keys, nil
 }
 
+func (d *azureKeyVaultDiscovery) getSecrets(kv *armkeyvault.Vault) ([]*voc.Secret, error) {
+	var (
+		secrets []*voc.Secret
+		c       *armkeyvault.SecretsClient
+	)
+	c = d.clients.secretsClient
```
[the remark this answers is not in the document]
lebogg: It is just a check here because it should be initialized elsewhere/one level up I think (like in `discoverKeyVaults`)

service/discovery/azure/key_vault.go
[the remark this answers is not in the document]
anatheka: The init function also checks first whether it has already been initialized.

anatheka (approved):

PR #1343, opened 2024-02-19
Better and more consistent asserts

oxisto:
This PR adds an internal `assert` package that should be used exclusivly by the tests. Under the hood, the existing testify library is used, but this might change in the future.


anatheka (approved):
LGTM

PR #1353, opened 2024-02-22
Compare resource group names in discovery using `strings.EqualFold`

oxisto:
Resource group names are case-insensitive. This fixes a bug, where the resource group was not discovered.


anatheka (approved):
LGTM

PR #1357, opened 2024-02-23
Extend discovery: Fix web app redundancy

lebogg:
We now finally get the right redundancy mode via App Service Farms

### DIFF
--- a/service/discovery/azure/azure.go
+++ b/service/discovery/azure/azure.go
@@ -162,8 +162,8 @@ type clients struct {
 	networkSecurityGroupsClient *armnetwork.SecurityGroupsClient
 
 	// AppService
-	sitesClient           *armappservice.WebAppsClient
-	appServiceFarmsClient *armappservice.PlansClient
+	sitesClient *armappservice.WebAppsClient
+	plansClient *armappservice.PlansClient
 
 	// Compute
 	virtualMachinesClient *armcompute.VirtualMachinesClient
--- a/service/discovery/azure/azure.go
+++ b/service/discovery/azure/azure.go
@@ -162,7 +162,8 @@ type clients struct {
 	networkSecurityGroupsClient *armnetwork.SecurityGroupsClient
 
 	// AppService
-	sitesClient *armappservice.WebAppsClient
+	sitesClient           *armappservice.WebAppsClient
+	appServiceFarmsClient *armappservice.PlansClient
 
 	// Compute
 	virtualMachinesClient *armcompute.VirtualMachinesClient
--- a/service/discovery/azure/compute.go
+++ b/service/discovery/azure/compute.go
@@ -232,7 +232,7 @@ func (d *azureComputeDiscovery) discoverFunctionsWebApps() ([]voc.IsCloudResourc
 	}
 
 	// initialize farms client
-	if err := d.initAppServiceFarmsClient(); err != nil {
+	if err := d.initAppServicePlansClient(); err != nil {
 		return nil, err
 	}
 
@@ -991,8 +991,8 @@ func (d *azureComputeDiscovery) initWebAppsClient() (err error) {
 	return
 }
 
-// initWebAppsClient creates the client if not already exists
-func (d *azureComputeDiscovery) initAppServiceFarmsClient() (err error) {
+// initAppServicePlansClient creates the client if not already exists
+func (d *azureComputeDiscovery) initAppServicePlansClient() (err error) {
 	d.clients.plansClient, err = initClient(d.clients.plansClient, d.azureDiscovery, armappservice.NewPlansClient)
 	return
 }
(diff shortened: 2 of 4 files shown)


service/discovery/azure/azure.go
```diff
@@ -162,7 +162,8 @@ type clients struct {
 	networkSecurityGroupsClient *armnetwork.SecurityGroupsClient
 
 	// AppService
-	sitesClient *armappservice.WebAppsClient
+	sitesClient           *armappservice.WebAppsClient
+	appServiceFarmsClient *armappservice.PlansClient
```
anatheka, line 166: Why is it called XFarmsClient and not PlansClient?

service/discovery/azure/compute.go
```diff
@@ -344,22 +349,43 @@ func (d *azureComputeDiscovery) handleWebApp(webApp *armappservice.Site, config
...
-	switch util.Deref(app.Properties.RedundancyMode) {
-	case armappservice.RedundancyModeNone:
-		break
-	case armappservice.RedundancyModeActiveActive:
-		r.Zone = true
-	case armappservice.RedundancyModeFailover, armappservice.RedundancyModeGeoRedundant:
-		r.Zone = true
-		r.Geo = true
+// getRedundancy returns for a given web app/function the redundancy in the voc format
+func (d *azureComputeDiscovery) getRedundancy(app *armappservice.Site) (r *voc.Redundancy) {
```
anatheka, line 357: Why do you get the information from the PlansClient and no longer from the site object? And why is the Geo property not longer set?

service/discovery/azure/azure.go
lebogg: For me, it sounded more intuitive. But yes, the new name is `plans`. I will change it

service/discovery/azure/compute.go
lebogg: Because the version before was hacky and not correct. The information about zone redundancy is stored in the respective service plan. And there, only the Zone Property is set, no information about Geo-Redundancy. We also only want Zone Redandancy for now, so that is enough.

service/discovery/azure/azure.go
lebogg: Funny enough, sometimes in the properties they still use "farms", e.g. "sites.Properties.ServerFarmID"

service/discovery/azure/compute.go
```diff
@@ -231,6 +231,11 @@ func (d *azureComputeDiscovery) discoverFunctionsWebApps() ([]voc.IsCloudResourc
 		return nil, err
 	}
 
+	// initialize farms client
+	if err := d.initAppServiceFarmsClient(); err != nil {
```
anatheka, line 235: If you rename the client, I would rename the init function as well.

service/discovery/azure/compute_test.go
```diff
@@ -3278,6 +3278,79 @@ func Test_getSecretURI(t *testing.T) {
...
+				azureDiscovery: NewMockAzureDiscovery(fake.NewPlansServerTransport(&FakeFarmServer),
+					WithResourceGroup("someRG")),
+			},
+			args: args{app: &armappservice.Site{
+				Properties: &armappservice.SiteProperties{
+					ServerFarmID: util.Ref("/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/someRG/providers/Microsoft.Web/serverfarms/FarmWithZoneRedundancy")}}},
+			wantR: &voc.Redundancy{Zone: true},
+		},
+		{
+			name: "Web App without zone redundancy enabled",
```
anatheka, line 3306:
```suggestion
			name: "Happy path - zone redundancy disabled",
```

service/discovery/azure/compute_test.go
```diff
@@ -3278,6 +3278,79 @@ func Test_getSecretURI(t *testing.T) {
...
+	type args struct {
+		app *armappservice.Site
+	}
+	tests := []struct {
+		name   string
+		fields fields
+		args   args
+		wantR  *voc.Redundancy
+	}{
+		{
```
anatheka, line 3294: Maybe add an additional check for the missing property.

service/discovery/azure/compute_test.go
```diff
@@ -3278,6 +3278,79 @@ func Test_getSecretURI(t *testing.T) {
...
+			d := &azureComputeDiscovery{
+				azureDiscovery: tt.fields.azureDiscovery,
+			}
+			assert.NoError(t, d.initAppServiceFarmsClient())
+			assert.Equalf(t, tt.wantR, d.getRedundancy(tt.args.app), "getRedundancy(%v)", tt.args.app)
+		})
+	}
+}
+
+func Test_getAppServiceFarmName(t *testing.T) {
```
anatheka, line 3327:
```suggestion
func Test_getAppServicePlanName(t *testing.T) {
```

anatheka (approved):
:+1:

service/discovery/azure/compute_test.go
lebogg: Good point

PR #1376, opened 2024-03-05
Add Diagnostic Settings information to Storage Resources

anatheka:
This PR adds the Diagnostic Settings from the storage account and blob/table/file service as ActivitityLogging to the following resources
- [x] StorageService (Storage Account)
- [x] ObjectStorage
- [x] DatabaseStorage
- [x] FileStorage

Further changes: 
- [x]  CosmosDB Diagnostic Settings are added as ActivityLogging forStorageService (CosmosDB)
- [x] Change ResourceLogging -> ActivityLogging for WebApp/Function

### DIFF
--- a/service/discovery/azure/storage.go
+++ b/service/discovery/azure/storage.go
@@ -39,6 +39,7 @@ import (
 
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/cosmos/armcosmos"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/dataprotection/armdataprotection"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/monitor/armmonitor"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/security/armsecurity"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/sql/armsql"
 	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/storage/armstorage"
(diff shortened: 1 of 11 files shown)


service/discovery/azure/storage.go
```diff
@@ -150,7 +151,13 @@ func (d *azureStorageDiscovery) discoverCosmosDB() ([]voc.IsCloudResource, error
 			return res.Value
 		},
 		func(dbAccount *armcosmos.DatabaseAccountGetResults) error {
-			cosmos, err := d.handleCosmosDB(dbAccount)
+			// Get ActivityLogging for the CosmosDB
+			activityLogging, err := d.discoverDiagnosticSettings(util.Deref(dbAccount.ID))
```
lebogg, line 155: maybe move that part to the `handleCosmosDB` function?

service/discovery/azure/storage.go
```diff
@@ -418,20 +433,22 @@ func (d *azureStorageDiscovery) discoverStorageAccounts() ([]voc.IsCloudResource
 			return res.Value
 		},
 		func(account *armstorage.Account) error {
+			activityLoggingAccount, activityLoggingBlob, activityLoggingFile, activityLoggingTable := d.getActivityLogging(account)
```
lebogg, line 436: See above

lebogg: Ah ok, here it is more efficient because you put it into multiple functions

service/discovery/azure/storage.go
```diff
@@ -502,7 +635,13 @@ func (d *azureStorageDiscovery) discoverObjectStorages(account *armstorage.Accou
 		}
 
 		for _, value := range pageResponse.Value {
-			objectStorages, err := d.handleObjectStorage(account, value)
+			// Discover diagnostic settings for the given container
+			_, err := d.discoverDiagnosticCategory(util.Deref(account.ID))
```
lebogg, line 639: If we don't use the first return parameter, do we actually need to define it?

service/discovery/azure/storage.go
anatheka: I would prefer to add all calls to Azure to the discoverX() functions. HandleX() functions schould not call Azure endpoints, but only process the information from Azure. I know we haven't done this in the past, but we should.

service/discovery/azure/storage.go
anatheka: This is a leftover from my tests.  :+1:

lebogg (commented):
LGTM, but as you suggested I would also change `resourceLogging` to `activityLogging` and adding diagnostic settings there

PR #1385, opened 2024-03-15
Fix WebApp

anatheka:
This PR fixes the following:
- Fix resource group in Azure calls
- Function/WebApp discovery now checks the Application Settings and Diagnostic Settings, but the following problem still exists: We are not able to discover the status of the Application Settings, we can only check it the Application Settings string is set, but not if it is enabled/disabled.


service/discovery/azure/compute.go
```diff
@@ -1129,23 +1131,32 @@ func (d *azureComputeDiscovery) initBackupInstancesClient() (err error) {
 // getResourceLogging determines if logging is activated for a given web app or function by checking the respective app setting
 // In this case logging means the Application Insights logging. The Application Insights logging is automatically forwarded to Log Analytics which is defined in the Diagnostic Settings.
 // TODO(all): Maybe be should discover also the Diagnostic Settings and split the logging from Application Insights and Diagnostic Settings (Log Analytics).
```
anatheka, line 1133:
```suggestion
// getResourceLogging determines if logging is activated for a given web app or function by checking the respective app setting
// First, it is checked if Application Insights is configured. If this is not configured, it is checked if Diagnostic Settings are configured and the logs are stored in a Log Analytics Workspace.    
// The Application Insights logging is automatically forwarded to Log Analytics.
```

service/discovery/azure/key_vault.go
```diff
@@ -353,6 +353,7 @@ func (d *azureKeyVaultDiscovery) getKeys(kv *armkeyvault.Vault) ([]*voc.Key, err
 	if c == nil {
 		return nil, errors.New("keys client is empty")
 	}
+	// TODO(all): We should use the listPager function, otherwise the call does not work if the Clouditor starts without a specific resource group.
```
lebogg, line 356: Doesn't work here because the list functions are different from the generic ones (here you only have one because a resource group always have to be provided as well). Anyway that is a different issuer and PR...

service/discovery/azure/service_fabric.go
```diff
@@ -80,6 +80,7 @@ func (d *azureServiceFabricDiscovery) discoverClusters() ([]voc.IsCloudResource,
 	}
 
 	var clusters []*armservicefabric.Cluster
+	// TODO(all): I think it would be better to use the listPager function.
```
lebogg, line 83: see comment above

service/discovery/azure/key_vault.go
lebogg: And yes it works if you replace `(util.Deref(d.azureDiscovery.rg` with sth. like `resourceGroup(util.Deref(kv.name))`. but again, different issue+PR

service/discovery/azure/storage.go
```diff
@@ -433,22 +433,22 @@ func (d *azureStorageDiscovery) discoverStorageAccounts() ([]voc.IsCloudResource
 			return res.Value
 		},
 		func(account *armstorage.Account) error {
-			activityLoggingAccount, activityLoggingBlob, activityLoggingFile, activityLoggingTable := d.getActivityLogging(account)
+			activityLoggingAccount, activityLoggingBlob, activityLoggingFile, activityLoggingTable, rawAccount, rawBlob, rawTable, rawFile := d.getActivityLogging(account)
```
lebogg, line 436: Ok for now, but in future/when migrating, I would maybe consider it doing it differently because for the sake of readability. First thought is to use a struct.
Also the rawXXX variables are a little bit confusing since they are not the raw values of the whole thing but only of the activity logging as far as I can tell. This would be also solved with a activity struct and putting everything inside.

lebogg (commented):
LGTM, some minor things. But please please in the future don't mix up/in other things not directly belonging here like the resource groups bug. It is a different issues and therefore should be in a different PR IMO

service/discovery/azure/azure.go
```diff
@@ -687,3 +687,58 @@ func resourceID(id *string) string {
...
+	listPager := d.clients.diagnosticSettingsClient.NewListPager(resourceURI, &armmonitor.DiagnosticSettingsClientListOptions{})
+	for listPager.More() {
+		pageResponse, err := listPager.NextPage(context.TODO())
+		if err != nil {
+			err = fmt.Errorf("%s: %v", ErrGettingNextPage, err)
+			return nil, "", err
+		}
+
+		// store raw information for resource
+		raw, _ = voc.ToStringInterface([]interface{}{pageResponse})
```
anatheka, line 713:
```suggestion
		raw, err = voc.ToStringInterface([]interface{}{pageResponse})
		if err != nil {
		    log.Errorf("%v: %v", voc.ErrConvertingStructToString, err)
		}
```

service/discovery/azure/azure.go
```diff
@@ -687,3 +687,58 @@ func resourceID(id *string) string {
...
+
+		for _, value := range pageResponse.Value {
+			// Check if data is send to a log analytics workspace
```
anatheka, line 716:
```suggestion
			// Check if data is sent to a log analytics workspace
```

service/discovery/azure/compute.go
```diff
@@ -1127,25 +1129,34 @@ func (d *azureComputeDiscovery) initBackupInstancesClient() (err error) {
 }
 
 // getResourceLogging determines if logging is activated for a given web app or function by checking the respective app setting
```
anatheka, line 1131:
```suggestion
// getActivityLogging determines if logging is activated for a given web app or function by checking the respective app setting
```

service/discovery/azure/azure.go
```diff
@@ -687,3 +687,58 @@ func resourceID(id *string) string {
 
 	return strings.ToLower(*id)
 }
+
+// discoverDiagnosticSettings discovers the diagnostic setting for the the given resource URI and returns the information of the needed information of the log properties as voc.ActivityLogging object.
```
anatheka, line 691:
```suggestion
// discoverDiagnosticSettings discovers the diagnostic setting for the given resource URI and returns the information of the needed information of the log properties as voc.ActivityLogging object and the Azure response.
```

anatheka (commented):
Fix review comments

service/discovery/azure/key_vault.go
anatheka: Regarding the first comment: You're right, we would have to use the allPages function. Maybe, we could think about exporting the allPages function and use that in the future.

service/discovery/azure/service_fabric.go
anatheka: I think we can use the listPager function here. If  I see it right, there is a NewListAllPager and NewListPager method. 

I've changed the TODOs.

service/discovery/azure/compute.go
```diff
@@ -321,7 +321,7 @@ func (d *azureComputeDiscovery) discoverSlots(baseSite *armappservice.Site, c ar
 		s    voc.IsCompute
 	)
 
-	pager := d.clients.sitesClient.NewListSlotsPager(util.Deref(d.rg), util.Deref(baseSite.Name), &armappservice.WebAppsClientListSlotsOptions{})
+	pager := d.clients.sitesClient.NewListSlotsPager(resourceGroupName(*baseSite.ID), util.Deref(baseSite.Name), &armappservice.WebAppsClientListSlotsOptions{})
```
lebogg, line 324: util.Deref(baseSite.ID)` would be safer against nil pointer exceptions although the ID shouldn't be empty ever :)

service/discovery/azure/compute.go
```diff
@@ -437,11 +438,11 @@ func (d *azureComputeDiscovery) handleWebApp(webApp *armappservice.Site, config
 		ni = []voc.ResourceID{voc.ResourceID(resourceID(webApp.Properties.VirtualNetworkSubnetID))}
 	}
 
-	activityLogging := d.getActivityLogging(webApp)
+	activityLogging, appSettings := d.getActivityLogging(webApp)
 
 	// Check if secrets are used and if so, add them to the 'secretUsage' dictionary
 	// Using NewGetAppSettingsKeyVaultReferencesPager would be optimal but is bugged, see https://github.com/Azure/azure-sdk-for-go/issues/14509
-	settings, err := d.clients.sitesClient.ListApplicationSettings(context.TODO(), util.Deref(d.rg),
+	settings, err := d.clients.sitesClient.ListApplicationSettings(context.TODO(), resourceGroupName(*webApp.ID),
```
lebogg, line 445: util.Deref

service/discovery/azure/compute.go
```diff
@@ -526,7 +528,7 @@ func (d *azureComputeDiscovery) getRedundancy(app *armappservice.Site) (r *voc.R
 		return
 	}
 	planName := getAppServicePlanName(app.Properties.ServerFarmID)
-	farm, err := d.clients.plansClient.Get(context.TODO(), util.Deref(d.rg), planName, &armappservice.PlansClientGetOptions{})
+	farm, err := d.clients.plansClient.Get(context.TODO(), resourceGroupName(*app.ID), planName, &armappservice.PlansClientGetOptions{})
```
lebogg, line 531: util.Deref

lebogg (approved):
LGTM

service/discovery/azure/compute.go
anatheka, line 324:
```suggestion
	pager := d.clients.sitesClient.NewListSlotsPager(resourceGroupName(util.Deref(baseSite.ID)), util.Deref(baseSite.Name), &armappservice.WebAppsClientListSlotsOptions{})
```

service/discovery/azure/compute.go
anatheka, line 445:
```suggestion
	settings, err := d.clients.sitesClient.ListApplicationSettings(context.TODO(), resourceGroupName(util.Deref(webApp.ID)),
```

service/discovery/azure/compute.go
anatheka, line 531:
```suggestion
	farm, err := d.clients.plansClient.Get(context.TODO(), resourceGroupName(util.Deref(app.ID)), planName, &armappservice.PlansClientGetOptions{})
```

PR #1388, opened 2024-03-18
Extend Discovery: Fix Key Vault

lebogg:
- [x] Add Check for empty Key Vaults
- [x] Fix expiration days for secrets (done via metrics, not here)
- [x] Fix number of usages of secrets


lebogg (commented):
sanity check :+1:

PR #1391, opened 2024-03-19
Extend Discovery: Fix public network access

lebogg:
Fix:

- [x] Discovery slots (they have their own API endpoints)
- [x] Container access (besides "None", we now  check "Blob" as well)
- [x] Improve cosmos DB public network check
- [x] Improve key vault public network check

### DIFF
--- a/service/discovery/azure/compute.go
+++ b/service/discovery/azure/compute.go
@@ -299,7 +299,7 @@ func (d *azureComputeDiscovery) discoverFunctionsWebApps() ([]voc.IsCloudResourc
 
 				// Also add function/web app slots
 				var slots []voc.IsCloudResource
-				slots, err = d.discoverSlots(site, config, isWebApp)
+				slots, err = d.discoverSlots(site, isWebApp)
 				if err != nil {
 					log.Errorf("Could not discover slots: %v", err)
 				}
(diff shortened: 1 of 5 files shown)


service/discovery/azure/compute.go
```diff
@@ -329,16 +329,23 @@ func (d *azureComputeDiscovery) discoverSlots(baseSite *armappservice.Site, c ar
 			return
 		}
 		for _, slot := range page.Value {
```
anatheka, line 331: Isn't that the same armappservice.Site object as in the discoverFunctionsWebApps()? If so, that shouldn't be called twice.

service/discovery/azure/compute.go
```diff
@@ -299,7 +299,7 @@ func (d *azureComputeDiscovery) discoverFunctionsWebApps() ([]voc.IsCloudResourc
 
 				// Also add function/web app slots
 				var slots []voc.IsCloudResource
-				slots, err = d.discoverSlots(site, config, isWebApp)
+				slots, err = d.discoverSlots(site, isWebApp)
```
anatheka, line 302: Shouldn't we discover the slots independently of the previously discovered WebApps/Functions? I think the discoverSlots() should be called outside the if.

service/discovery/azure/storage.go
```diff
@@ -174,12 +174,12 @@ func (d *azureStorageDiscovery) discoverCosmosDB() ([]voc.IsCloudResource, error
...
-		atRestEnc           voc.IsAtRestEncryption
-		err                 error
-		list                []voc.IsCloudResource
-		publicNetworkAccess = false
+		atRestEnc voc.IsAtRestEncryption
+		err       error
+		list      []voc.IsCloudResource
 	)
 
+	// TODO(lebogg): Initialization can be removed here, it is done in discoverCosmosDBs - or vice versa.
```
anatheka, line 182: However, it does not interfere, as the initClient function checks whether the client already exists.

service/discovery/azure/compute.go
lebogg: The `if r != nil` was there before and as far as I understand, we use it because we don't discover all possibilities of `site`s, e.g. `case "app,linux,container"`. So I wanted to make sure to discover slots only if we discover the corresponding base site.

PR #1396, opened 2024-03-28
Extend discovery: Fix byok

lebogg:
We didn't have the possibility to set encryption for entire storage services (e.g. Storage Account or Cosmos DB). Because it doesn't support AtRestEncryption. As a hotfix, I added a `encryptionManagedByUser` field to the voc file and use it in `handleStorageAccount` and `handleCosmosDB`


anatheka (approved):
LGTM

PR #1407, opened 2024-04-12
Change metric interval from int64 to duration

anatheka:


anatheka:
Do we also want change the Evaluation interval to duration instead of int32? I would leave it as int32. 
https://github.com/clouditor/clouditor/blob/e4b0c992fec57f931cc071527744d7c1d0838969/api/evaluation/evaluation.proto#L74

oxisto (approved):

Issue #1417, opened 2024-04-19
Create an appropriate ontology entity for a CSAF trusted provider and a CSAF security advisory


oxisto:
We would need to the following properties in order to support #1418:

Two entities CSAFDocument and CSAFService. One is a service, the other one is "Data"?!

| Requirements                              | Entity                      | Needed Properties                         |
| ----------------------------------------- | --------------------------- | ----------------------------------------- |
| Valid CSAF document                       | CSAFDocument                | JSON schema validation errors             |
| Valid filename                            | CSAFDocument                | filename                                  |
| TLS                                       | CSAFDocument -> CSAFService | TransportEncryption security feature      |
| TLP_WHITE accessible                      | CSAFDocument -> CSAFService | AccessRestriction security feature, label |
| TLP_AMBER, RED access control             | CSAFDocument -> CSAFService | AccessRestriction security feature, label |
| No redirects                              | ?                           | ?                                         |
| provider-metadata.json                    | CSAFService                 | list of CSAFDocuments                     |
| security.txt                              | AddtionalSecurityInfo??     | Entries                                   |
| Well-known URL for provider-metadata.json | CSAFService                 | URL of provider-metadata.json             |

PR #1424, opened 2024-04-23
Update ontology with CSAF security advisory concepts, generate new proto

immqu:
This PR updates the ontology which now includes objects for SecurityAdvisoryDocument, ServiceMetadataDocument, and other objects that are required to discover and assess evidences related to advisory standards like CSAF.


immqu:
addresses #1417

PR #1430, opened 2024-04-24
Fixed linter execution

oxisto:
It looks like it had a missing permission for the token and there was a conflict between Go and linter cache.


anatheka (approved):
:+1:

PR #1431, opened 2024-04-24
Added ontology entities for a document location

oxisto:
A `Document` can now have a `DocumentLocation`, which is either local or remote; depending on this, we either can have an at-rest-encryption or a transport-encryption for this particular document.

This is necessary to check if a CSAF document is served using TLS.


anatheka (approved):
LGTM

Issue #1425, opened 2024-04-23
Make use of csaf-poc/csaf_distribution#536 to discover multiple PMDs


oxisto:
Currently, we can only check the conformance of a single PMD on a provider. When https://github.com/csaf-poc/csaf_distribution/pull/536 is implemented, we can use the new `Enumerate` function in the metadata-loader to retrieve all PMDs and then create individual evidences for them.

oxisto:
Fixed by https://github.com/clouditor/clouditor/pull/1436

PR #1436, opened 2024-04-26
Use enumerate method to discover all PMDs of a CSAF domain

immqu:
This PR addresses #1425 : It uses the new Enumerate method in the csaf distribution to discover all PMDs that are included in a given domain.


service/discovery/extra/csaf/provider_discover.go
```diff
@@ -13,8 +13,24 @@ import (
 
 func (d *csafDiscovery) discoverProviders() (providers []ontology.IsResource, err error) {
 	loader := csaf.NewProviderMetadataLoader(d.client)
-	lpmd := loader.Load(d.domain)
+	lpmd := loader.Enumerate(d.domain)
```
oxisto, line 16: Maybe name it `lpmds` to denote that they are multiple?

oxisto (approved):
Only a small change

PR #1448, opened 2024-04-29
CSAF: Add  metric for Public OpenPGP Key (Requirement 20)

lebogg:
This PR adds a Clouditor metric for CSAF Standard Requirement 20


lebogg:
@oxisto In the requirement they write about the ONE public key that is used to sign the CSAF document.
But in our ontology object for a SecurityAdvisoryService we can have multiple ones. I check now that we have at least one public key and that each public key is available.

policies/bundles/CSAF/SecurityAdvisoryServicePublicOpenGPG/metric.rego
```diff
@@ -0,0 +1,28 @@
...
+compliant if {
+	# Public part of PGP key must be available, i.e. 1st) there is a public key und 2nd) it is (publicly) available
+	# 1) There is a public key
+	key_ids := service.keyIds
+	count(key_ids) > 0
+
+	# 2) They are publicly availabe
+	every key_id in key_ids {
+		key := service.related[key_id]
+		key.internetAccessibleEndpoint == true
```
oxisto, line 26: Do we actually set this or did you mock it for the tests?

lebogg: Only for tests for now. I didn't dive into how we are going to set it, yet.

I just saw that a `csaf.PGPKey` has a URL which we set a ID for the ontology object. Maybe you could argue that if we have a URL it is also publicly available. Maybe we have to do a short http.get() request to check it. I'm not so sure.

oxisto: We already do this check because we actually fetch the key in `fetchKey`, so we could set it.

lebogg: I tried it. But tbh I am not so into the PGP structure and topics regarding CSAF. Therefore I'm not sure if this change makes sense (e.g. we have information about a key from PMD so we assume it exists. But if we cannot fetch it, is it still wise to assume its existence? And in particular how to translate that to our ontology. I guess we want to have resources/evidence in both cases with appropriate properties.) . Feel free to leave a review :)

oxisto: It looks reasonable to me

oxisto:
> @oxisto In the requirement they write about the ONE public key that is used to sign the CSAF document. But in our ontology object for a SecurityAdvisoryService we can have multiple ones. I check now that we have at least one public key and that each public key is available.

I would also see it this way, in almost every CSAF provider I have seen, there were multiple keys; simple because of key rotation every X years.

oxisto (approved):

PR #1442, opened 2024-04-28
Implemented more metrics to check a CSAF provider

oxisto:
This PR adds more metrics to check a CSAF provider.

PR #1449, opened 2024-04-29
CSAF: Extend `metrics.json` and `csaf.json` to include REQs 18-20

lebogg:
Adds the metrics for the following CSAF requirements to metrics.json:

- Requirement 18: Integrity (implemented in #1442 )
- Requirement 19: Signatures (implemented in #1442 )
- Requirement 20: Public OpenPGP (implemented in #1448 )


lebogg:
Blocked by #1448

anatheka (approved):
LGTM

Just an idea: Maybe we could add a comment to the CSAF metric.go files with the requirement number? Then it is easier to find the corresponding requirement than if you have to look again in csaf.json or metrics.json.

PR #1473, opened 2024-05-21
Sorting indirect dependencies to groups in go.mod

oxisto:
We (unfortunatly) have a lot of indirect depdendencies that we pull in from the "big" SDKs. This PR should make it easier to quickly see which indirect dependency is coming from where and maybe try to cut them down piece by piece.


anatheka (approved):
I'm not sure if this is a better and easier solution, but go for it :)

PR #1507, opened 2024-07-18
Fix control evaluation

anatheka:
This PR fixes the control evaluation.

### DIFF
--- a/service/evaluation/evaluation.go
+++ b/service/evaluation/evaluation.go
@@ -650,27 +650,15 @@ func (svc *Service) evaluateControl(ctx context.Context, toe *orchestrator.Targe
 	var nonCompliantAssessmentResults = []string{}
 
 	for _, r := range results {
-		// status is the current evaluation status, r.Status is the current assessment result status
+		// status is the current evaluation status, r.Status is the status of the evaluation result of the subcontrol
 		// Note: Status should only contain the evaluation status without _MANUALLY!
 		switch status {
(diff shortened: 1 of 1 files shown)


service/evaluation/evaluation.go
```diff
@@ -650,16 +650,29 @@ func (svc *Service) evaluateControl(ctx context.Context, toe *orchestrator.Targe
...
-				status != evaluation.EvaluationStatus_EVALUATION_STATUS_NOT_COMPLIANT_MANUALLY) {
-			status = evaluation.EvaluationStatus_EVALUATION_STATUS_COMPLIANT
-		} else if r.Status == evaluation.EvaluationStatus_EVALUATION_STATUS_PENDING &&
-			status == evaluation.EvaluationStatus_EVALUATION_STATUS_PENDING {
-			status = evaluation.EvaluationStatus_EVALUATION_STATUS_PENDING
-		} else {
-			status = evaluation.EvaluationStatus_EVALUATION_STATUS_NOT_COMPLIANT
+		// status is the current evaluation status, r.Status is the current assessment result status
+		// Note: Status should only contain the evaluation status without _MANUALLY!
+		switch status {
```
oxisto, line 655: What is the point of the extra switch? I fail to see the difference to the original code here

anatheka: The difference is the failing else-part in the previous version.

oxisto: As discussed offline, we should split the individual sub-selects into different functions for better readability.

oxisto (approved):

PR #1529, opened 2024-08-20
Added google.api annotations

oxisto:


anatheka (commented):
Shouldn't all `buf.validate.field).required = true` also be replaced by `(google.api.field_behavior) = REQUIRED`? And the same for all other proto files?

oxisto:
> Shouldn't all `buf.validate.field).required = true` also be replaced by `(google.api.field_behavior) = REQUIRED`? And the same for all other proto files?

Yes

oxisto:
> > Shouldn't all `buf.validate.field).required = true` also be replaced by `(google.api.field_behavior) = REQUIRED`? And the same for all other proto files?
> 
> Yes

I added the remaining annotations but I will keep this constrained to the orchestrator API for now. I want to do some more testing if this is really a good approach.

anatheka (approved):

PR #1541, opened 2024-09-25
Update func for the public network access status

anatheka:
This PR modifies the function for retrieving the public network access status. Previously, the function was limited to WebApp resources; now, it can be utilized by all resource types to obtain the status.

### DIFF
--- a/service/discovery/azure/azure_properties.go
+++ b/service/discovery/azure/azure_properties.go
@@ -260,3 +260,18 @@ func location(region *string) *ontology.GeoLocation {
 		Region: util.Deref(region),
 	}
 }
+
+// publicNetworkAccessStatus returns the public access status of the resource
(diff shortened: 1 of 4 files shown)


service/discovery/azure/azure_properties.go
```diff
@@ -260,3 +260,18 @@ func location(region *string) *ontology.GeoLocation {
 		Region: util.Deref(region),
 	}
 }
+
+// publicNetworkAccessStatus returns the public access status of the resource
+func publicNetworkAccessStatus(status *string) bool {
```
oxisto, line 265: I think we should stick with the old name because we had a lot of trouble with what exactly "public network access status" is

anatheka:
Each resource has its own object, so there is no added value if we outsource the function.

PR #1555, opened 2024-10-10
Ontology Update

oxisto:
- Added generated OpenAPI defs for ontology
- Added metadata for required for discovery service


api/discovery/experimental.proto
```diff
@@ -68,13 +69,13 @@ message ListGraphEdgesRequest {
...
+  repeated GraphEdge edges = 1 [(google.api.field_behavior) = REQUIRED];
   string next_page_token = 2;
 }
 
 message GraphEdge {
-  string id = 1;
-  string source = 2;
-  string target = 3;
-  string type = 4;
+  string id = 1 [(google.api.field_behavior) = REQUIRED];
```
anatheka, line 77: What is about `(buf.validate.field).string.min_len = 1,`, `(buf.validate.field).string.uuid = true,`, ...?

api/ontology/service.proto
```diff
@@ -0,0 +1,47 @@
...
+
+package clouditor.ontology.v1;
+
+import "api/ontology/ontology.proto";
+import "google/api/annotations.proto";
+import "google/protobuf/empty.proto";
+
+option go_package = "clouditor.io/clouditor/v2/api/ontology";
+
+service OntologyService {
```
anatheka, line 36: Did you forget to add the implementation?

anatheka (commented):
Only some minor things.

api/ontology/service.proto
oxisto: No, this was intentionally done to provoke openapi in generating all the resources. I remember you had a similar approach with a fake service.

api/discovery/experimental.proto
oxisto: The IDs here are not UUIDs but are the resource IDs (which are probably URLs), but min length makes sense

anatheka (approved):
LGTM

PR #1560, opened 2024-10-14
Add flag for embedded OAuth 2.0 server public URL

anatheka:


server/rest/rest.go
```diff
@@ -192,11 +195,12 @@ func WithEmbeddedOAuth2Server(keyPath string, keyPassword string, saveOnCreate b
 
 // RunServer starts our REST API. The REST API is a reverse proxy using grpc-gateway that
 // exposes certain gRPC calls as RESTful HTTP methods.
-func RunServer(ctx context.Context, grpcPort uint16, port uint16, serverOpts ...ServerConfigOption) (err error) {
+func RunServer(ctx context.Context, grpcPort uint16, port uint16, host string, serverOpts ...ServerConfigOption) (err error) {
```
oxisto, line 198: I wouldn't to do that this way. the "host" that you are putting into the function here is only the public URL of the auth server and this can differ from the "host" where the rest server is running. For example the host would be something like an internal docker IP where as the public URL is the URL of the load balancer. Instead please check for the appropriate flag inside the function `WithEmbeddedOAuth2Server` where it is needed.

oxisto (changes requested):

oxisto (approved):

PR #1564, opened 2024-10-15
Added support for extended non-compliance-details

oxisto:
This PR adds the possibility for metrics to supply additional (non)-compliance details to an assessment result. This is mainly done through tracking the various comparisons done in a metric. An example metric that uses this new system is also supplied.

Potentially the way this result is created inside the metric can be generalized for multiple metrics.


api/assessment/assessment.proto
```diff
@@ -132,11 +133,31 @@ message AssessmentResult {
...
+// An optional structure containing more details how a comparison inside an assessment result was done and if it was succesful.
+message ComparisonResult {
+  // Property is the property that was compared
+  string property = 1;
+
+  // Value is the value in the property
+  string value = 2;
+
+  // Operator is the operator used in the comparison
+  string operator = 3 [(buf.validate.field).string.pattern = "^(|<|>|<=|>=|==|isIn|allIn)$"];
```
lebogg, line 154: Is the first Pipe intentional?

lebogg (approved):

api/assessment/assessment.proto
oxisto: I copy/pasted it from the metric configuration. Probably it should not be there.

PR #1565, opened 2024-10-18
Supply comments for all assessment results

oxisto:
In order to make our assessment results more human friendly, we set a human readable comment in all cases now. The metric can customize this message, if it wants. We also set the comparion result for all compliant and non-compliant results.


api/assessment/assessment.proto
```diff
@@ -98,66 +99,104 @@ message AssessEvidencesResponse {
...
 
   // Reference to the tool which provided the assessment result
-  optional string tool_id = 21 [(buf.validate.field).string.min_len = 1];
+  optional string tool_id = 21 [
+    (buf.validate.field).string.min_len = 1,
+    (google.api.field_behavior) = REQUIRED
+  ];
 }
 
 // An optional structure containing more details how a comparison inside an assessment result was done and if it was succesful.
```
anatheka, line 171:
```suggestion
// An optional structure containing more details how a comparison inside an assessment result was done and if it was successful.
```

api/evaluation/evaluation.proto
```diff
@@ -136,38 +137,55 @@ message EvaluationResult {
   // Evaluation result id
   string id = 1 [
     (buf.validate.field).string = {uuid: true},
-    (buf.validate.field).ignore_empty = true
+    (buf.validate.field).ignore_empty = true,
+    (google.api.field_behavior) = REQUIRED
```
anatheka, line 141: Does that make sense? UUID is required but can be empty? (l. 139-141)

service/evaluation/evaluation.go
```diff
@@ -661,20 +662,23 @@ func (svc *Service) evaluateControl(ctx context.Context, auditScope *orchestrato
 			status, nonCompliantAssessmentResults = handleCompliant(r)
 		case evaluation.EvaluationStatus_EVALUATION_STATUS_NOT_COMPLIANT, evaluation.EvaluationStatus_EVALUATION_STATUS_NOT_COMPLIANT_MANUALLY:
 			// Evaluation status does not change if it is already not_compliant
-			nonCompliantAssessmentResults = append(nonCompliantAssessmentResults, r.GetFailingAssessmentResultIds()...)
+			nonCompliantAssessmentResults = append(nonCompliantAssessmentResults, r.GetAssessmentResultIds()...)
```
anatheka, line 665: Why are the nonCompliantAssessmentResults no longer included in the Evaluation Result? I would prefer to have that information in the result.

service/orchestrator/certification_target.go
```diff
@@ -48,10 +50,7 @@ import (
 )
 
 const (
-	DefaultCertificationTargetId          = "00000000-0000-0000-0000-000000000000"
-	DefaultCertificationTargetName        = "default"
-	DefaultCertificationTargetDescription = "The default certification target"
-	DefaultCertificationTargetType        = orchestrator.CertificationTarget_TARGET_TYPE_CLOUD
+	DefaultCertificationTargetId = "00000000-0000-0000-0000-000000000000"
```
anatheka, line 53: We could also move that to the constants in /internal/config/config.go.

service/orchestrator/audit_scope.go
```diff
@@ -184,6 +185,13 @@ func (svc *Service) RemoveAuditScope(ctx context.Context, req *orchestrator.Remo
 	}
 	go svc.informToeHooks(ctx, &orchestrator.AuditScopeChangeEvent{Type: orchestrator.AuditScopeChangeEvent_TYPE_AUDIT_SCOPE_REMOVED, AuditScope: auditScope}, nil)
 
+	if req.RemoveEvaluationResults {
```
anatheka, line 188: Should it also be possible to delete the associated assessment results and, if applicable, the evidence?

anatheka (commented):
This PR also includes other useful changes, such as removing evaluation results when the audit scope is deleted, in addition to appending comments for the assessment results. Perhaps we should consider renaming the PR?

service/orchestrator/audit_scope.go
oxisto: No. The assessment result are independent from the audit scope.

service/evaluation/evaluation.go
oxisto: All results are now in the evaluation result

anatheka: But the idea was that we know which assessment results are responsible for negative evaluation results. 
Anyway, one can also manually extract the information again. In any case, I would rename the nonCompliantAssessmentResults to assessmentResults or something similar.

service/orchestrator/certification_target.go
oxisto: Is needed from the outside

service/evaluation/evaluation.go
oxisto: We will keep only the "all" property.

api/assessment/assessment.go
```diff
@@ -40,6 +40,12 @@ var (
 	ErrMetricConfigurationTargetValueMissing = errors.New("target value in metric data is missing")
 )
 
+const (
+	DefaultNonCompliantMessage = "The metric has assessed that the resource contains properties that are not compliant to the target value."
+	DefaultCompliantMessage    = "The result of the metric shows that the evidence is compliant to the target value."
+	AdditionalDetailsMessage   = "Additional details can be found in the comparison below."
```
anatheka, line 46: Sure, that the details are always below? Or are they in the field ComparisonResult?

anatheka (approved):
Great!

api/assessment/assessment.go
oxisto: Well they are below in the UI ;)

anatheka: I find that difficult. If someone changes the UI view and doesn't update it accordingly, no one will know where to find anything. Additionally, if it's retrieved using curl, it probably isn't below. 

I think something like "Additional details can be found in the comparison" or "comparison view/part/field"  better.

PR #1566, opened 2024-10-19
Fixed bug in evaluation

oxisto:
This fixes a major bug in the evaluation, where the eval results basically never went to compliant and was always stuck in pending.


anatheka (approved):
LGTM

PR #1593, opened 2024-12-04
Update postgres from v1.5.0 to v1.5.6

anatheka:
This PR updates the postgres library from version v1.5.0 to v1.5.6. 
With version v1.5.0 we get the following error: 
> Error: could not create storage: error during auto-migration: insufficient arguments

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -86,7 +86,7 @@ require (
 	github.com/jinzhu/now v1.1.5 // indirect
 	github.com/mattn/go-isatty v0.0.17 // indirect
 	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
-	gorm.io/driver/postgres v1.5.0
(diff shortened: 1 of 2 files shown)


api/orchestrator/orchestrator.proto
```diff
@@ -749,7 +749,8 @@ message Category {
   repeated Control controls = 4 [
     (tagger.tags) =
       "gorm:\"foreignKey:category_name,category_"
-      "catalog_id;references:name,catalog_id;constraint:OnDelete:CASCADE\"",
+      "catalog_id;references:name,catalog_id;constraint:OnDelete:CASCADE\""
+,
```
anatheka, line 753:
```suggestion
      "catalog_id;references:name,catalog_id;constraint:OnDelete:CASCADE\"",
```

oxisto (approved):

PR #1606, opened 2025-01-08
Switched UI to confirmate UI

oxisto:
This PR switches the UI to the confirmate UI. It is still not 100 % complete, but we want to continue the development there.


anatheka (approved):
Maybe add a comment in the README!?!

PR #1624, opened 2025-02-03
Revert "Bump actions/labeler from 4 to 5"

anatheka:
Reverts clouditor/clouditor#1372

We should avoid updating to version 5,as it appears to be unstable.


anatheka:
I think I found a way to fix that.

PR #1646, opened 2025-02-25
[Discovery] Remove raw field from evidence

anatheka:
This PR deletes the raw field from the evidence because it was present in both the evidence and the resource field and we only need the raw field in one of both.


anatheka (commented):
@anatheka delete raw field from evidence and "add" to the resource.

oxisto (approved):

PR #1647, opened 2025-02-26
Add launcher check for storage creation

anatheka:
This PR adds a check for the used launcher. If the launcher assessment is used, no storage creation is necessary as the assessment does not store anything to the DB.

### DIFF
--- a/launcher/launcher.go
+++ b/launcher/launcher.go
@@ -76,15 +76,16 @@ func NewLauncher(name string, specs ...ServiceSpec) (l *Launcher, err error) {
 	}
 
 	// Set storage
-	err = l.initStorage()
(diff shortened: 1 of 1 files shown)


launcher/spec.go
```diff
@@ -73,9 +73,14 @@ func (s spec[T]) NewService(db persistence.Storage) (svc service.Service, grpcOp
 	return s.newService(db)
 }
 
+func (s spec[T]) HasStorageFunction() bool {
```
oxisto, line 76: I am not 100 % sure I understand why this is necessary but it seems to work :D Can we make this unexported -> `hasStorageFunction`

oxisto (approved):

PR #1649, opened 2025-02-26
[Azure] Add ML workspace discovery

anatheka:
While we could potentially extend the ontology and include additional information in the resource object, this serves as a starting point for further enhancements.

### DIFF
--- a/service/discovery/azure/machine_learning_discovery.go
+++ b/service/discovery/azure/machine_learning_discovery.go
@@ -32,6 +32,7 @@ import (
 
 	"clouditor.io/clouditor/v2/api/ontology"
 	"clouditor.io/clouditor/v2/internal/util"
+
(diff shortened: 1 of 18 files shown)


service/discovery/azure/machine_learning_discovery.go
```diff
@@ -0,0 +1,122 @@
...
+			return list, err
+		}
+
+		for _, value := range pageResponse.Value {
+			compute, err := d.handleMLCompute(value, workspace.ID)
+			if err != nil {
+				return nil, fmt.Errorf("could not handle ML workspace: %w", err)
+			}
+
+			if compute == nil {
```
lebogg, line 110: This can happen?

lebogg (approved):
LGTM

PR #1676, opened 2025-03-27
api,server: using `ontology.Resource` message in `Evidence.resource`

oxisto:
Instead of an untested `protobuf.Any` we are now using the `ontology.Resource` message, which in turn has a one-of field to populate the `Evidence.resource` field.

This has the advantage that now a proper OpenAPI description is now generated for this field.


api/evidence/evidence.go
```diff
@@ -41,40 +42,37 @@ func (req *StoreEvidenceRequest) GetPayload() proto.Message {
...
-	m, err = ev.Resource.UnmarshalNew()
-	if err != nil {
+
+	if ev.Resource == nil || ev.Resource.Type == nil {
 		return nil
 	}
 
-	resource, ok = m.(ontology.IsResource)
+	// A little bit of dark Go magic
+	typ := reflect.ValueOf(ev.Resource.Type).Elem()
```
lebogg, line 54: I saw that this `.Elem()` function could panic. Therefore I took a closer look and added some tests. It's good due to the nil checks above.

api/evidence/evidence.go
```diff
@@ -41,40 +42,37 @@ func (req *StoreEvidenceRequest) GetPayload() proto.Message {
...
-	// Try to extract the resource out of the evidence
-	m, err = ev.Resource.UnmarshalNew()
-	if err != nil {
+
+	if ev.Resource == nil || ev.Resource.Type == nil {
 		return ""
 	}
 
-	resource, ok := m.(ontology.IsResource)
+	typ := reflect.ValueOf(ev.Resource.Type).Elem()
```
lebogg, line 74: I saw that this `.Elem()` function could panic. Therefore I took a closer look and added some tests and also a nil check like you did above

api/ontology/ontology.go
```diff
@@ -2,10 +2,11 @@ package ontology
 
 import (
 	"encoding/json"
-	"google.golang.org/protobuf/reflect/protoreflect"
 	"slices"
 	"strings"
 
+	"google.golang.org/protobuf/reflect/protoreflect"
+
```
lebogg, line 9:
```suggestion
```

lebogg: empty line

api/ontology/ontology.go
```diff
@@ -142,3 +143,30 @@ func ResourceIDs(r []IsResource) []string {
 
 	return a
 }
+
+// ProtoResource converrts a [IsResource] to a [Resource]
```
lebogg, line 147:
```suggestion
// ProtoResource converts a [IsResource] to a [Resource]
```

api/ontology/ontology.go
```diff
@@ -142,3 +143,30 @@ func ResourceIDs(r []IsResource) []string {
...
+func ProtoResource(resource IsResource) *Resource {
+	var (
+		r  Resource
+		m  protoreflect.Message         = r.ProtoReflect()
+		od protoreflect.OneofDescriptor = m.Descriptor().Oneofs().ByName("type")
```
lebogg, line 152:
```suggestion
		m  = r.ProtoReflect()
		od = m.Descriptor().Oneofs().ByName("type")
```

lebogg: just a suggestion

lebogg: And: I'm not sure if I like to do operations in the var block already. But it is valid Go code so just my two cents

api/ontology/ontology.go
```diff
@@ -142,3 +143,30 @@ func ResourceIDs(r []IsResource) []string {
...
+	)
+
+	for i := range od.Fields().Len() {
```
lebogg, line 155: I trust you on this. Maybe add some comment what we do here

api/ontology/ontology_test.go
```diff
@@ -189,3 +188,36 @@ func TestListResourceIDs(t *testing.T) {
...
+	type args struct {
+		resource IsResource
+	}
+	tests := []struct {
+		name string
+		args args
+		want *Resource
+	}{
+		{
+			name: "happy path",
```
lebogg, line 202: Think about corner cases

service/assessment/assessment.go
```diff
@@ -405,23 +404,11 @@ func (svc *Service) AssessEvidences(stream assessment.Assessment_AssessEvidences
...
-	}
-
-	err = api.Validate(m)
-	if err != nil {
-		return nil, err
-	}
-
-	resource, ok := m.(ontology.IsResource)
-	if !ok {
+	resource = ev.GetOntologyResource()
```
lebogg, line 410: much cleaner now :)

lebogg (commented):
LGTM. Some minor things like adding comments and check if tests are covering the different scenarios

api/evidence/evidence.go
```diff
@@ -41,43 +42,20 @@ func (req *StoreEvidenceRequest) GetPayload() proto.Message {
...
 		resource ontology.IsResource
-		err      error
 		ok       bool
 	)
-	// TODO: find a smarter way, because now we are unmarshalling the resource twice
-	// Try to extract the resource out of the evidence
-	m, err = ev.Resource.UnmarshalNew()
-	if err != nil {
+
+	if ev.Resource == nil || ev.Resource.Type == nil {
```
lebogg, line 49: Would it make sense to use `util.IsNil` here as well?

lebogg (approved):
LGTM

api/evidence/evidence.go
oxisto: No, since ev.Resource is not an interface and we only need the utility function to check for nil on interfaces.

PR #1681, opened 2025-04-03
Update Ontology

anatheka:
Update ontology after splitting into multiple files.


openapi/evidence/openapi.yaml
```diff
@@ -109,3000 +109,27 @@ paths:
                                 $ref: '#/components/schemas/Status'
 components:
     schemas:
-        ABAC:
```
oxisto, line 112: This looks fishy. Why did this PR delete the generated ontology structures for OpenAPI?

anatheka: I fixed it by removing the Evidence class from the ontology.owx file.

oxisto (approved):

PR #1709, opened 2025-05-05
assessment: Add fields `updatedAt` and `history` to AssessmentResult

anatheka:


api/assessment/assessment.proto
```diff
@@ -166,6 +166,25 @@ message AssessmentResult {
     (buf.validate.field).string.min_len = 1,
     (google.api.field_behavior) = REQUIRED
   ];
+
+  // The time of the last update of the assessment result history field″
```
anatheka, line 170:
```suggestion
  // The time of the last update of the assessment result history field
```

api/assessment/assessment.proto
```diff
@@ -166,6 +166,25 @@ message AssessmentResult {
...
+  // Stores the history of evidence IDs and timestamps for evidence that have the same content as the evidence used for this assessment result.
+  repeated Record history = 23 [
+    (tagger.tags) = "gorm:\"serializer:json;constraint:OnDelete:CASCADE\"",
+    (buf.validate.field).required = true,
+    (google.api.field_behavior) = REQUIRED
+  ];
+}
+
+message Record {
+  string evidence_id = 1;
```
anatheka, line 186: Add UUID required field.

service/assessment/assessment.go
```diff
@@ -487,6 +487,11 @@ func (svc *Service) handleEvidence(
 			ComplianceComment:    data.Message,
 			ComplianceDetails:    data.ComparisonResult,
 			ToolId:               util.Ref(assessment.AssessmentToolId),
+			HistoryUpdatedAt:     timestamppb.Now(),
+			History: []*assessment.Record{{
+				EvidenceId:         ev.GetId(),
+				EvidenceRecordedAt: timestamppb.Now(),
+			}},
```
lebogg, line 494: That looks like creating a new list with only one item each time. I guess we want something with append, don't we?

anatheka: Yes, that’s correct. The implementation isn’t finished yet. In the next pull request I’ll add logic to decide whether the assessment result is being created for the first time or whether only its history field is being extended.

These are only preliminary preparations so far.

anatheka: #1724

lebogg (approved):

PR #1733, opened 2025-06-03
evidenceStore: Add endpoint 'ListSupportedResourceTypes'

anatheka:
This PR adds an additional endpoint to the EvidenceStore to get the supported Ontology `resource` types.

### DIFF
--- a/api/evidence/evidence_store.proto
+++ b/api/evidence/evidence_store.proto
@@ -114,5 +114,5 @@ message GetEvidenceRequest {
 message GetSupportedResourceTypesRequest {}
 
 message GetSupportedResourceTypesResponse {
-  repeated string resource_types = 1 [(buf.validate.field).repeated.min_items = 1];
+  repeated string resource_type = 1 [(buf.validate.field).repeated.min_items = 1];
 }
--- a/api/evidence/evidence_store.proto
+++ b/api/evidence/evidence_store.proto
@@ -60,7 +60,7 @@ service EvidenceStore {
   }
 
   // Returns the resource types that are supported by the EvidenceStore.
-  rpc GetSupportedResourceTypes(GetSupportedResourceTypesRequest) returns (GetSupportedResourceTypesResponse) {
+  rpc ListSupportedResourceTypes(ListSupportedResourceTypesRequest) returns (ListSupportedResourceTypesResponse) {
     option (google.api.http) = {get: "/v1/evidence_store/supported_resource_types"};
   }
 }
@@ -111,8 +111,8 @@ message GetEvidenceRequest {
   string evidence_id = 1 [(buf.validate.field).string.uuid = true];
 }
 
-message GetSupportedResourceTypesRequest {}
+message ListSupportedResourceTypesRequest {}
 
-message GetSupportedResourceTypesResponse {
+message ListSupportedResourceTypesResponse {
   repeated string resource_type = 1 [(buf.validate.field).repeated.min_items = 1];
 }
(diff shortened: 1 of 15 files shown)


api/ontology/ontology.go
```diff
@@ -88,6 +88,43 @@ func ResourceTypes(r IsResource) []string {
 	return nil
 }
 
+// GetResourceTypes returns a list of resource types that are supported by the ontology.
+func GetResourceTypes() []string {
+	var types []string
+
+	var resource Resource
+
+	// Zugriff auf die Descriptor der Resource-Message
```
oxisto, line 97: Why are you adding comments in German? :D

api/evidence/evidence_store.proto
```diff
@@ -58,6 +58,11 @@ service EvidenceStore {
   rpc GetEvidence(GetEvidenceRequest) returns (Evidence) {
     option (google.api.http) = {get: "/v1/evidence_store/evidences/{evidence_id}"};
   }
+
+  // Returns the resource types that are supported by the EvidenceStore.
+  rpc GetSupportedResourceTypes(GetSupportedResourceTypesRequest) returns (GetSupportedResourceTypesResponse) {
```
lebogg, line 63: Because you don't provide any content in `GetSupportedResourceTypesRequest` and you actually list all resource types, I would rename it to ListXXX

api/ontology/ontology.go
```diff
@@ -88,6 +88,26 @@ func ResourceTypes(r IsResource) []string {
...
+// GetResourceTypes returns a list of resource types that are supported by the ontology.
+func GetResourceTypes() []string {
+	var types []string
+
+	var resource Resource
+
+	// Accessing the descriptor of the resource message.
+	md := resource.ProtoReflect().Descriptor()
+
+	// Durchlaufen der Felder der Resource
```
lebogg, line 100: English please

api/ontology/ontology.go
```diff
@@ -88,6 +88,26 @@ func ResourceTypes(r IsResource) []string {
...
+	for i := 0; i < md.Fields().Len(); i++ {
+		field := md.Fields().Get(i)
+		if field.ContainingOneof() != nil {
```
lebogg, line 103: What is this about?

api/ontology/ontology_test.go
```diff
@@ -228,3 +228,23 @@ func TestProtoResource(t *testing.T) {
...
+
+func TestGetResourceTypes(t *testing.T) {
+	tests := []struct {
+		name string
+		want assert.Want[[]string]
+	}{
+		{
+			name: "Happy path",
+			want: func(t *testing.T, got []string) bool {
+				return assert.NotEmpty(t, got)
```
lebogg, line 240: This test is ok, but maybe not the best (accurate).

Another option would be to hardcode the amount of resources and test this against the length of the list. We then would need to change it if we create new ontology objects. So we have to test each time if it still works properly.

lebogg (changes requested):
LGTM except the german comment :)

api/ontology/ontology.go
anatheka: Our owl2proto tool generates a oneOf field that currently includes every resource. If a bug in our implementation or a mistake in the ontology file leads to resources appearing outside the oneOf definition, those resources shouldn't displayed.

lebogg: I think you have to explain me it in person :D but for now I guess this is fine :)

lebogg (approved):
LGTM

PR #1751, opened 2025-06-30
streams: Fix panic in GetStream()

anatheka:


api/streams_test.go
```diff
@@ -131,6 +132,40 @@ func TestStreamsOf_GetStream(t *testing.T) {
...
+							return c
+						}(),
+						target:    "mock:1234",
+						component: "mock",
+					},
+				},
+			},
+			args: args{
+				target:    "mock:1234",
+				component: "mock component",
```
lebogg, line 159: the component name "mock component" in `args` is not the same as "mock" in `fields` but it doesn't matter here I guess

api/streams.go
```diff
@@ -134,7 +134,7 @@ func (s *StreamsOf[StreamType, MsgType]) GetStream(target string, component stri
 		// We could have a dead stream that we need to restart. in this case, we can recycle a few things, e.g. the channel
 		c, err = s.restartStream(c, init, opts...)
 		if err != nil {
-			return nil, fmt.Errorf("could not restart stream for %s with target '%s': %w", c.component, c.target, err)
+			return nil, fmt.Errorf("could not restart stream for %s with target '%s': %w", component, target, err)
```
lebogg, line 137: :+1: :smile:

lebogg (approved):
LGTM

api/streams_test.go
lebogg: It is still the same but the test passes

PR #1817, opened 2026-04-01
metrics: update security-metrics submodule

anatheka:
This PR updates the security-metrics submodule, which now includes metrics for Governance.


anatheka:
The version is already too old; a more recent version should be integrated.

PR #1846, opened 2026-09-22
update security metrics repo

immqu:
This PR updates the security metrics repo, which is necessary due to inconsistencies in how some of the metrics are named and handled.


anatheka (commented):
You have to fix the other files before merging, because the newest security-metrics repo uses a lot of optional fields. Otherwise the complete code is broken. 

Christian already fixes that in another branch.

